PR #13, opened 2016-09-08
tells flake to ignore files in egg based dirs

lundjordan:
maybe you're not comfortable with my aggressive *s 😄 

I can be more explicit if you are paranoid

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -39,7 +39,7 @@ commands=
 
 [flake8]
 max-line-length = 160
-exclude = .ropeproject,.tox,sandbox,docs,*.egg*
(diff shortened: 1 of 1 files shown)


lundjordan:
@escapewindow

escapewindow:
Yeah, let's do *.egg and *.egg-info and .eggs

lundjordan:
for the record, this fixes the flake errors when you use a stock 'setup.py develop/install' venv that doesn't install tox.

only `setup.py test` will install tox and will do so by installing tox + reqs locally in the build dir which flake8 parses

PR #14, opened 2016-09-23
ports download_artifacts and download_file to scriptworker

lundjordan:
needs tests

### DIFF
--- a/scriptworker/exceptions.py
+++ b/scriptworker/exceptions.py
@@ -55,3 +55,10 @@ class ScriptWorkerTaskException(ScriptWorkerException):
     def __init__(self, *args, exit_code=1, **kwargs):
         self.exit_code = exit_code
         super(ScriptWorkerTaskException, self).__init__(*args, **kwargs)
+
+
+class DownloadError(ScriptWorkerTaskException):
+    def __init__(self, msg):
+        super(DownloadError, self).__init__(
+            msg, exit_code=3
+        )
--- a/scriptworker/exceptions.py
+++ b/scriptworker/exceptions.py
@@ -60,5 +60,5 @@ class ScriptWorkerTaskException(ScriptWorkerException):
 class DownloadError(ScriptWorkerTaskException):
     def __init__(self, msg):
         super(DownloadError, self).__init__(
-            msg, exit_code=3
+            msg, exit_code=4
         )
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -4,25 +4,23 @@
 Attributes:
     log (logging.Logger): the log object for the module
 """
-from copy import deepcopy
-
 import aiohttp.hdrs
 import arrow
 import asyncio
+from asyncio.subprocess import PIPE
+from copy import deepcopy
 import logging
 import mimetypes
 import os
 import signal
 
-from asyncio.subprocess import PIPE
-
 import taskcluster
 import taskcluster.exceptions
 
+from scriptworker.client import validate_artifact_url, REVERSED_STATUSES
 from scriptworker.exceptions import ScriptWorkerRetryException
 from scriptworker.log import get_log_fhs, log_errors, read_stdout
 from scriptworker.utils import filepaths_in_dir, raise_future_exceptions, retry_async, download_file
-from scriptworker.client import validate_artifact_url, REVERSED_STATUSES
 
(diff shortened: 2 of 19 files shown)


scriptworker/exceptions.py
```diff
@@ -55,3 +55,10 @@ class ScriptWorkerTaskException(ScriptWorkerException):
     def __init__(self, *args, exit_code=1, **kwargs):
         self.exit_code = exit_code
         super(ScriptWorkerTaskException, self).__init__(*args, **kwargs)
+
+
+class DownloadError(ScriptWorkerTaskException):
+    def __init__(self, msg):
+        super(DownloadError, self).__init__(
+            msg, exit_code=3
```
escapewindow: I think 'malformed-payload' means the task definition was wrong, while a DownloadError might mean there's an infrastructure problem.  I'm not sure if we should use a 4, 5, or 1; I've been guessing so far.  Luckily we can adjust these pretty easily.

scriptworker/task.py
```diff
@@ -329,3 +304,24 @@ def max_timeout(context, proc, timeout):
         asyncio.ensure_future(kill(-pid)),
         asyncio.ensure_future(kill(pid))
     ]))
+
+
+async def download_artifacts(context, file_urls, download_func=download_file):
```
escapewindow: I was thinking about putting this and download_file in scriptworker.client since those are more shared with client scripts, and that module could become a standalone module at some point.  I'm not sure if that's a valuable convention, yet, so we don't have to go that way.

scriptworker/utils.py
```diff
@@ -15,11 +15,19 @@
 import shutil
 from taskcluster.utils import calculateSleepTime
 from taskcluster.client import createTemporaryCredentials
-from scriptworker.exceptions import ScriptWorkerException, ScriptWorkerRetryException
+from scriptworker.exceptions import ScriptWorkerException, ScriptWorkerRetryException, DownloadError
 
 log = logging.getLogger(__name__)
 
 
+def mkdir(path):
```
escapewindow: I think I already have a makedirs() in here.

scriptworker/utils.py
```diff
@@ -306,3 +314,19 @@ def format_json(data):
         str: the formatted json.
     """
     return json.dumps(data, indent=2, sort_keys=True)
+
+
+async def download_file(context, url, abs_filename, chunk_size=128):
+    log.info("Downloading %s", url)
+    parent_dir = os.path.dirname(abs_filename)
+    async with context.session.get(url) as resp:
```
escapewindow: considering removing `context` from the arg list, and having an optional kwarg `session`, overrideable from `download_artifacts`.  That would help avoid the weird context.session swapping.

scriptworker/task.py
```diff
@@ -3,10 +3,9 @@
 
 Attributes:
     log (logging.Logger): the log object for the module
-    STATUSES (dict): maps taskcluster status (string) to exit code (int).
-    REVERSED_STATUSES (dict): the same as STATUSES, except it maps the exit code
-        (int) to the taskcluster status (string).
 """
+from copy import deepcopy
```
escapewindow: Oh, a weird style thing:  I try to keep the imports all alphabetized by module.

```python
# 3rd party imports
import arrow
import asyncio
from copy import deepcopy
import tempfile
# local imports
from scriptworker.client ...
from scriptworker log ...
from scriptworker.utils ...
```

scriptworker/task.py
```diff
@@ -4,25 +4,23 @@
 Attributes:
     log (logging.Logger): the log object for the module
 """
-from copy import deepcopy
-
 import aiohttp.hdrs
 import arrow
 import asyncio
+from asyncio.subprocess import PIPE
+from copy import deepcopy
```
lundjordan: heh, I blame my IDE for the unorganized imports 😆

scriptworker/utils.py
```diff
@@ -322,7 +314,7 @@ def format_json(data):
     async with context.session.get(url) as resp:
         if resp.status != 200:
             raise DownloadError("{} status {} is not 200!".format(url, resp.status))
-        mkdir(parent_dir)
+        makedirs(parent_dir)
```
lundjordan: ah yeah, maybe I only grepped for 'mkdir' not 'makedir'

PR #16, opened 2016-10-05
gpg homedir creation

escapewindow:
I fixed the git loop!  By `await`ing the `asyncio.sleep` :( :( :(

Some known issues:

* This isn't a clean history of reviewable commits.  If you need that, I'll have to revisit.
* <s> Known issue: the actual creation of the gpg homedirs dies due to 

```python
  File "/builds/scriptworker/scriptworker/scriptworker/gpg.py", line 1389, in create_initial_gpg_homedirs
    build_gpg_homedirs_from_repo(context, basedir=context.config['base_gpg_home_dir'])
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 387, in run_until_complete
    return future.result()
  File "/tools/python35/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/tools/python35/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/builds/scriptworker/scriptworker/scriptworker/gpg.py", line 1291, in build_gpg_homedirs_from_repo
    await verify_function(context)
  File "/builds/scriptworker/scriptworker/scriptworker/gpg.py", line 1254, in verify_signed_git_commit
    GPG(context), output, context.cot_config['git_key_repo_fingerprints']
  File "/builds/scriptworker/scriptworker/scriptworker/gpg.py", line 1165, in verify_signed_git_commit_output
    fingerprint = keyid_to_fingerprint(gpg, keyid)
  File "/builds/scriptworker/scriptworker/scriptworker/gpg.py", line 133, in keyid_to_fingerprint
    keyid, guess_gpg_home(gpg)
scriptworker.exceptions.ScriptWorkerGPGException: Can't find keyid FC829B7FFAA9AC38 for /home/cltsign/.gnupg!
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/tools/python35/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/tools/python35/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/tools/python35/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
```

I need to fix that.</s>

This is now ready for review, although I can definitely split this up into more reviewable chunks.  This is currently live on signing-linux-{1..4}.

Some things I'd love to fix, but aren't urgent:

* I'd love to get rid of `cot_config.json` now that it's no longer governing the valid fingerprint list.
* I'd love to use yaml instead of json for `scriptworker.json`, since the config is now getting large and would benefit from some comments.

I'm probably going to leave those for post- scriptworker 1.0 , since getting all these scriptworkers up and running for end-to-end testing is the first priority.


scriptworker/gpg.py
```diff
@@ -1113,25 +1131,337 @@ def rebuild_gpg_home_signed(context, real_gpg_home, my_pub_key_path,
...
+    revision = await get_git_revision(path)
+    return old_revision != revision
+
+
+async def verify_signed_git_commit(context, path=None,
+                                   exec_function=asyncio.create_subprocess_exec):
+    """Verify `context.config['git_key_repo_dir']` is on a valid signed commit.
+
+    This function calls `verify_signed_git_commit_output` to make sure the
+    latest non-merge-commit commit is signed with a key that lives in
```
JohanLorenzo: What happens if `git pull` fetches more than one commit? Do we check if the other commits are signed too?

scriptworker/gpg.py
```diff
@@ -1113,25 +1131,337 @@ def rebuild_gpg_home_signed(context, real_gpg_home, my_pub_key_path,
...
+                tasks.append(asyncio.ensure_future(
+                    signed_function(
+                        context, real_gpg_home, my_pub_key_path, my_priv_key_path,
+                        trusted_path, untrusted_path=untrusted_path,
+                        ignore_suffixes=worker_config['ignore_suffixes']
+                    )
+                ))
+            await raise_future_exceptions(tasks)
+    finally:
+        # rm lockfile
```
JohanLorenzo: Nit: I don't think this comment adds value to the line below

config_example.json
```diff
@@ -8,15 +8,22 @@
...
     "artifact_upload_timeout": 1200,
     "task_script": ["bash", "-c", "echo foo && sleep 19 && exit 1"],
     "task_max_timeout": 1200,
     "valid_artifact_schemes": ["https"],
     "valid_artifact_netlocs": ["queue.taskcluster.net"],
     "valid_artifact_path_regexes": ["^/v1/task/(?P<taskId>[^/]+)(/runs/\\d+)?/artifacts/(?P<filepath>.*)$"],
+
     "verify_chain_of_trust": false,
     "sign_chain_of_trust": false,
+    "gpg_path": null,
```
escapewindow: That's the idea, to use gpg in $PATH by default.  What do you think should be in the config_example.json, `/usr/local/bin/gpg` ?

cot_config_example.json
```diff
@@ -0,0 +1,21 @@
+{
+    "git_key_repo_url": "https://github.com/mozilla-releng/cot-gpg-keys.git",
+    "git_commit_signing_pubkey_dir": "...",
```
escapewindow: Sure, though `/tmp/gpg/` will get blown away because it's the `base_gpg_home_dir` in `config_example.json`.

scriptworker/config.py
```diff
@@ -102,30 +109,35 @@ def check_config(config, path):
...
     Then validate it and freeze it.
 
     Args:
-        path (str, optional): the path to the config file.  Defaults to "config.json"
+        config_path (str, optional): the path to the config file.  Defaults to
+            "scriptworker.json"
 
     Returns:
-        tuple: (config dict, credentials dict)
+        tuple: (config frozendict, credentials dict)
```
escapewindow: This is behavior I originally added to mozharness.  The idea is, `context` is used a lot, and `context.config` and `context.cot_config` as well.  If they're dicts, then they can be modified at runtime easily.  This may be intentional, if we want to change `context.config['foo'] = bar`.  Sometimes this is unintentional: we pass `context.config` to something by reference, and that function makes an unintentional change, and we end up with a modified `context.config` that causes weird behavior later on.

By changing `context.config` and `context.cot_config` to frozendicts with tuples instead of lists as child values, we force someone who wants to make a runtime change to explicitly cast to a dict before they can alter values; this is discouraged anyway, and we'll be more clear that that's what's happening.  Otherwise we're more certain that the runtime config will not change accidentally.

scriptworker/constants.py
```diff
@@ -36,16 +37,20 @@
     "max_connections": 30,
     "credential_update_interval": 300,
     "reclaim_interval": 300,
+    "poll_git_interval": 300,
```
escapewindow: Sure.  Comments are the main reason why I'm considering yaml instead of json for the config files, but `constants.py` can take comments in the meantime.

cot_config_example.json
```diff
@@ -0,0 +1,21 @@
...
+    "gpg_keyserver": "hkp://gpg.mozilla.org",
+    "pubkey_path": "...",
```
escapewindow: I made these changes, but in `/tmp` rather than `/tmp/gpg`

scriptworker/config.py
```diff
@@ -134,3 +146,65 @@ def create_config(path="config.json"):
...
+        context (scriptworker.context.Context): the scriptworker context.
+
+    Returns:
+        frozendict: the Chain of Trust config.
+
+    Raises:
+        SystemExit: on failure
+    """
+    cot_config_path = cot_config_path or context.config['cot_config_path']
+    if not os.path.exists(cot_config_path):
```
escapewindow: Sure.  Done.

scriptworker/constants.py
```diff
@@ -55,7 +60,7 @@
     "gpg_secret_keyring": '%(gpg_home)s/secring.gpg',
     # Boolean to use the gpg agent
     "gpg_use_agent": False,
-    # Encoding to use.  Defaults to latin-1
+    # Encoding to use.  Defaults to utf-8
```
escapewindow: None, easy enough to change. Done.

scriptworker/gpg.py
```diff
@@ -1113,25 +1131,337 @@ def rebuild_gpg_home_signed(context, real_gpg_home, my_pub_key_path,
...
+    Args:
+        path (str): the path to run `git rev-parse HEAD` in.
+
+    Returns:
+        str: the revision.
+
+    Raises:
+        ScriptWorkerRetryException: on failure.
+    """
+    proc = await exec_function(
```
escapewindow: Hm.  Makes sense, but I don't think it's pressing just to replace a single call.

scriptworker/gpg.py
escapewindow: Nope.  As it stands, we're only checking for the latest non-merge commit being signed.

This means several things:

* signing a commit means you're signing all the previous unsigned commits, which is a little worrisome, but this should be documented.
* however, this allows for creation of pull requests from automation to add pubkeys for taskcluster workers, and then a human only needs to push a final signed commit to the PR to validate it.  I think this will be easier in terms of process, though we could force a rebase here.

scriptworker/gpg.py
escapewindow: +1. Guessing this was a todo item or more complex logic at some point, not sure.  Removed the comment.

config_example.json
JohanLorenzo: Maybe `gpg` alone would be enough? Otherwise `/usr/local/bin/gpg` sounds good

scriptworker/config.py
JohanLorenzo: I see! Thanks for the details

scriptworker/gpg.py
JohanLorenzo: I see. IIUC, this works if we merge 1 PR per pull interval. Is this risk of having these conditions low enough?

* 1 PR ready to be merged, with one of them already signed by someone trusted
* a GH account corrupted, which creates another (corrupted) PR
* both PRs merged at the same time, so that the genuine commit is placed at the HEAD

scriptworker/gpg.py
```diff
@@ -1036,18 +1055,18 @@ def rebuild_gpg_home_flat(context, real_gpg_home, my_pub_key_path, my_sec_key_pa
...
-                            my_sec_key_path, trusted_path, untrusted_path=None,
-                            ignore_suffixes=(),
-                            consume_function=consume_valid_keys):
+async def rebuild_gpg_home_signed(context, real_gpg_home, my_pub_key_path,
+                                  my_priv_key_path, trusted_path,
+                                  untrusted_path=None, ignore_suffixes=(),
+                                  consume_function=consume_valid_keys):
     """Rebuild `real_gpg_home` with new trustdb, pub+secrings, gpg.conf.
 
     In this 'signed' model, import all the pubkeys in `consume_path` and sign
```
JohanLorenzo: Probably a bad copy and paste: `consume_path` doesn't exist and `rebuild_gpg_home_signed()` does make some fingerprints trusted.

config_example.json
escapewindow: I'm relying on the [gnupg defaults](https://pythonhosted.org/python-gnupg/#getting-started).  That already defaults to 'gpg'.  If you want me to fill it out in the config file to be more explicit, I can do that.

scriptworker/gpg.py
JohanLorenzo: Awesome! I didn't notice the check on Travis side

config_example.json
JohanLorenzo: As we're exposing this file to the other instances of scriptworker, I'd make that explicit. I wouldn't bikeshed on that though, 😃

scriptworker/gpg.py
escapewindow: Good catch. Updated the docstring.

escapewindow:
Added a note in the README about gpg 2.0.x.  I also bumped the version to 0.8.0 and updated the CHANGELOG to match.  I'll push 0.8.0 to pypi and puppet, then publish a blog post.

Thanks for the review!  It's good to have a second set of eyes here, plus hopefully this gets you more up to speed on scriptworker, for future reviews or PRs.

PR #17, opened 2016-10-17
0.8.1: explicitly pass taskId and runId from the azure json

escapewindow:
I've been assuming that `message_info['task_info']` will always look like

```python
{
    "taskId": "...",
    "runId": 0
}
```

However, in [this changeset](https://github.com/taskcluster/taskcluster-queue/pull/124/files#diff-2ba94b1f0cbf39f965dbec5a3f38019cR592), @jonasfj added a `hintId`.

This change explicitly passes `taskId` and `runId` rather than using `message_info['task_info']` as kwargs.

After pushing this patch, and a puppet-specific change, to puppet, the signing scriptworkers began claiming new tasks again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+## [0.8.1] - 2016-10-18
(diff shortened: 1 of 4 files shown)


escapewindow:
@JohanLorenzo if you're up for it, I'd love a review :)

PR #18, opened 2016-10-23
0.8.2: fix polling

escapewindow:
* The symptoms involved the polling dying, but not the git pull.  Instead of relying on a single `create_task` and then an `loop.run_forever()` to poll, put a `loop.run_until_complete()` inside a `while True` loop.  The git pulls are happening every 5 minutes as desired, and the taskcluster polling is still going.

* Log the traceback if we hit any exceptions running `async_main`.  I noticed that supervisor was restarting the process a lot while github was down on Friday, and I wasn't sure what was breaking.  This should give us more info.

* Remove the `rebuild_gpg_homedirs_loop()` basedir after we're done copying the contents, instead of at the beginning of the call.  This helps avoid having background processes step on each other.  We may end up having to monitor the tmpdir for freshness if the git pull hangs, but that hasn't been an issue so far.

Once I get review, I'm going to bump to 0.8.2 and merge, then cherry pick the fixes onto the 0.7.x branch for 0.7.2.

@JohanLorenzo , until @lundjordan gets back you're the one with the most scriptworker background; are you up for the review?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
+- logged the stacktrace if the `main` loop hits an exception.  No longer catch and ignore `RuntimeError`, since it wasn't clear why that was put in.
+- updated `check_config` to make sure taskcluster-related configs match taskcluster requirements
+
+### Fixed
+- changed the way the polling loop works: `async_main` is now a single pass, which `main` calls in a `while True` loop.  This should fix the situation where polling was dying silently while the git update loop continued running every 5 minutes.
 
 ## [0.8.1] - 2016-10-18
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+## [0.8.2] - 2016-10-24
 ### Changed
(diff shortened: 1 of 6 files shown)


scriptworker/worker.py
```diff
@@ -107,9 +107,17 @@ def main():
     with aiohttp.ClientSession(connector=conn) as session:
         context.session = session
         context.credentials = credentials
+        tmp_gpg_home, _ = _get_tmp_gpg_home(context)
+        loop.create_task(
```
JohanLorenzo: ayncio noob question: What happens if we don't wait on the task created here? Does it still execute it somehow, but we don't have control on when it happens?

scriptworker/worker.py
```diff
@@ -107,9 +107,17 @@ def main():
...
         while True:
             try:
-                loop.create_task(async_main(context))
-                loop.run_forever()
-            except RuntimeError:
-                pass
+                loop.run_until_complete(async_main(context))
+            except Exception as exc:
+                log.warning(traceback.format_exc())
+                if not isinstance(exc, RuntimeError):
```
JohanLorenzo: If I understand correctly, this condition is here to filter out async errors like: `RuntimeError: Event loop stopped before Future completed`, isn't it? 

If that's the case, I wonder if it wouldn't worth commenting it it the code, for future asyncio-rookies.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Fixed
```
JohanLorenzo: Maybe we could also append the part about logging every stacktrace for every kind of unhandled exceptions?

PR #22, opened 2016-10-29
0.9.0: Pull gpg homedir creation out of the main scriptworker process.

### LINKED ISSUE #21: intermittent async pexpect errors
`11:00 <aki> async pexpect appears to have an intermittent hang, making async gpg sign_keys unreliable. i may have to tear out gpg homedir creation into a separate process =\`

We basically get errors like

```
DEBUG:scriptworker.gpg:gpg --homedir /tmp/tmpxtfienl7 --no-default-keyring --secret-keyring /tmp/tmpxtfienl7/secring.gpg --keyring /tmp/tmpxtfienl7/pubring.gpg -u C4F19E3EBD445E29E680B26A1E47FDB1FD63EC52 --lsign-key 83A4B550BC682F0B060157B04654904BB484B6B2
ERROR:asyncio:Exception in callback BaseSelectorEventLoop.add_reader(11, <bound method...d=11 polling>>)
handle: <Handle BaseSelectorEventLoop.add_reader(11, <bound method...d=11 polling>>)>
  File "/tools/python35/lib/python3.5/asyncio/selector_events.py", line 234, in add_reader
    key = self._selector.get_key(fd)
  File "/tools/python35/lib/python3.5/selectors.py", line 191, in get_key
    raise KeyError("{!r} is not registered".format(fileobj)) from None
KeyError: '11 is not registered'
```

This appears to be https://github.com/pexpect/pexpect/issues/347; a patch is pending in https://github.com/pexpect/pexpect/pull/376 .

We can either try to get the upstream issue fixed, or tear out gpg homedir creation into its own parallel process.

escapewindow:
Per the commit message:

```
Originally, gpg homedir creation happened before scriptworker launched,
and then a background async task polled git and rebuilt the gpg homedirs
if it found a new valid git commit.  In order to sign the gpg keys in
the background, we used async pexpect.

However, as noted in #21, async pexpect isn't reliable currently.  Our
options included fixing the upstream pexpect bug or pulling the gpg
homedir creation into a separate process that we can call via cron.

This patch allows us to use that latter approach.
```

Fixes #21 .

@JohanLorenzo , I'll let you thumb wrestle with @lundjordan with review now that he's back :)  I'm happy to have either of you review.  Thanks to both of you!

I'm going to release scriptworker 0.9.0 off this PR once it's approved.

### DIFF
--- a/scriptworker/gpg.py
+++ b/scriptworker/gpg.py
@@ -28,8 +28,8 @@ import traceback
 from scriptworker.config import get_context_from_cmdln
 from scriptworker.exceptions import ScriptWorkerException, ScriptWorkerGPGException, \
     ScriptWorkerRetryException
-from scriptworker.log import pipe_to_log
-from scriptworker.utils import filepaths_in_dir, makedirs, raise_future_exceptions, retry_async, rm
+from scriptworker.log import pipe_to_log, update_logging_config
+from scriptworker.utils import filepaths_in_dir, makedirs, retry_async, rm
 
 log = logging.getLogger(__name__)
 
@@ -300,18 +300,13 @@ def export_key(gpg, fingerprint, private=False):
     return key
 
 
-@asyncio.coroutine
 def sign_key(context, target_fingerprint, signing_key=None,
              exportable=False, gpg_home=None):
     """Sign the `target_fingerprint` key with the `signing_key` or default key
 
     This signs the target key with the signing key, which adds to the web of trust.
 
-    This function is async, but should not be used in parallel in the same
-    gpg homedir.  GPG keysigning requires a lock on the keyring, so any
-    parallelization in the same homedir will result in failures and unhappiness.
-    This function is meant to be used async to parallelize multiple GPG
-    homedir operations.
+    Due to pexpect async issues, this function is once more synchronous.
 
     Args:
         context (scriptworker.context.Context): the scriptworker context.
@@ -345,7 +340,7 @@ def sign_key(context, target_fingerprint, signing_key=None,
     child = pexpect.spawn(gpg_path, cmd_args, timeout=context.config['sign_key_timeout'])
     try:
         while True:
-            index = yield from child.expect([pexpect.EOF, b".*Really sign\? \(y/N\) ", b".*Really sign all user IDs\? \(y/N\) "], async=True)
+            index = child.expect([pexpect.EOF, b".*Really sign\? \(y/N\) ", b".*Really sign all user IDs\? \(y/N\) "])
             if index == 0:
                 break
             child.sendline(b'y')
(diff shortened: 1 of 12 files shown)


scriptworker/gpg.py
```diff
@@ -1322,8 +1313,48 @@ def verify_signed_git_commit_output(output):
...
+        ScriptWorkerGPGException: on failure
+    """
+    log.info(
+        "writing last_good_git_revision {} to {}...".format(
+            revision, context.config['last_good_git_revision_file']
+        )
+    )
+    try:
+        with open(context.config['last_good_git_revision_file'], "w") as fh:
+            print(revision, file=fh, end='')
```
JohanLorenzo: Out of curiosity, what's the difference with `fh.write(revision)`?

scriptworker/gpg.py
```diff
@@ -1322,8 +1313,48 @@ def verify_signed_git_commit_output(output):
...
+    except OSError as exc:
+        raise ScriptWorkerGPGException(str(exc))
```
JohanLorenzo: Shouldn't we wrap the exception under another type of Error? IIUC, the current usage of `ScriptWorkerGPGException` is limited to actual GPG operations.

scriptworker/gpg.py
```diff
@@ -1343,125 +1374,164 @@ def verify_signed_git_commit_output(output):
...
+        name (str): the name of the calling function
+
+    Returns:
+        bool: True if lockfile exists.
+    """
+    lockfile = context.config['gpg_lockfile']
+    if os.path.exists(lockfile):
+        log.warning("Skipping {}: lockfile {} exists!".format(name, lockfile))
+        return True
+
```
JohanLorenzo: Nit: `False` doesn't seem returned even though the documentation reads a bool is returned

scriptworker/gpg.py
```diff
@@ -1343,125 +1374,164 @@ def verify_signed_git_commit_output(output):
...
-            # We don't want these exceptions to be fatal; this will happen in the background
-            # of scriptworker tasks
-            log.error(message.format(str(exc)))
-            continue
+    return '{}.tmp'.format(context.config['base_gpg_home_dir'])
 
 
-def create_initial_gpg_homedirs():
-    """Create the initial gpg homedirs.
+def check_lockfile(context, name):
```
JohanLorenzo: `check_lockfile()` sounds ambiguous to me. I'm not sure if `True` means whether the file exists or I'm good to proceed. How about `is_lockfile_present()`?

scriptworker/gpg.py
escapewindow: Looks like the same thing :)

scriptworker/gpg.py
escapewindow: fixed.

scriptworker/gpg.py
escapewindow: Now that we're only calling this from external endpoints, I think we don't have to catch the `OSError` at all.  It does the trick and kills the process, while removing the lockfile in the `finally` block.  I removed the try/except block.

escapewindow:
Thanks again!

PR #23, opened 2016-11-01
scriptworker.yaml

escapewindow:
I split out `cot_config` from `config` earlier to avoid an overly-complex json file.  `config_example.json` has been wanting comments for a while, but hasn't been commented due to the restrictions of the format.

Moving to yaml allows us to combine the configs without too much complexity, since we can use comments in the example config.

This is a pretty large patch (another 61k patch) but a lot of it is tests and requirements-*prod.txt
I thought opening up a PR for just this piece might be easier than having this part of the entire chain of trust verification PR.  If that's not true, I can close this PR and create one massive PR later.

@JohanLorenzo @lundjordan do you prefer thumb wrestling or do you want me to just choose one of you? :)


scriptworker.yaml.tmpl
```diff
@@ -0,0 +1,145 @@
+#-----------------------------------------------------------------------------------------------
+# Taskcluster worker settings.
+# For development, keep the test-dummy-* and dummy-worker-* convention.
```
JohanLorenzo: It so nice to be able to put comments in config files 👍 😃

scriptworker.yaml.tmpl
```diff
@@ -0,0 +1,145 @@
...
+#-----------------------------------------------------------------------------------------------
+# credentials:
+#    clientId: ...
+#    accessToken: ...
+#    certificate: ...
+
+#-----------------------------------------------------------------------------------------------
+# Task configs
+#-----------------------------------------------------------------------------------------------
+artifact_expiration_hours: 24
```
JohanLorenzo: Out of curiosity, would it worth to nest the task configuration? For example:
```yaml
task:
  artifact_expiration_hours: 24
  max_timeout: 1200
```
or 
```yaml
gpg:
  email: "scriptworker@example.com"
```

Edit: After going further in the PR, I realized this'll break too many things.

scriptworker.yaml.tmpl
```diff
@@ -0,0 +1,145 @@
...
+sign_chain_of_trust: false
+verify_chain_of_trust: false
+verify_cot_signature: false
+
+
+#-----------------------------------------------------------------------------------------------
+# Scriptworker paths.
+#-----------------------------------------------------------------------------------------------
+# Scriptworker logs go here; this is a long-lived directory.
+log_dir: "/tmp/log"
```
JohanLorenzo: If this is a long-lived dir, maybe we should use a different example than `/tmp`? What do you think of something like: `/var/log/my_scriptworker_instance/`

scriptworker.yaml.tmpl
```diff
@@ -0,0 +1,145 @@
...
+#-----------------------------------------------------------------------------------------------
+# Valid artifact rules.
+# This is a list of dictionaries.  Each dictionary specifies schemes, netlocs, and path_regexes.
+# All valid artifact downloads should match these.  `filepath` must be specified in the
+# path_regexes.
+#
+# If `taskId` is specified in the path_regex, it must be in task.dependencies, the decision task,
+# or an upstream chain of trust task.
+#-----------------------------------------------------------------------------------------------
+# valid_artifact_rules: [{
```
JohanLorenzo: TIL: JSON is actually a subset of YAML. I wonder if making the configuration purely yaml, is a good idea as a matter of readability. What do you think?

eg: 
```yaml
valid_artifact_rules: 
  - 
    netlocs: 
      - queue.taskcluster.net
    path_regexes: 
      - "^/v1/task/(?P<taskId>[^/]+)(/runs/\\d+)?/artifacts/(?P<filepath>.*)$"
    schemes: 
      - https
```

scriptworker.yaml.tmpl
escapewindow: I never intended /tmp to be a real location for anything but testing.
For puppetized instances, we're going with /builds/scriptworker/... and the like.
I'm not sure we gain that much from moving the `log_dir` for dev instances, and it forces someone testing to create /var/log/... and make sure it's writable for the current user.

I'm going to skip this suggestion for now, but if you feel strongly about it we can revisit.

scriptworker.yaml.tmpl
escapewindow: Sure!  Done.

PR #25, opened 2016-11-08
1.0.0b1: chain of trust verification

escapewindow:
This pull request is not ready for merging; I just wanted to get some preliminary feedback/review as I work on tests and deployment.  I'm going to be adding full unittest coverage in the [cot-verify-test branch](https://github.com/escapewindow/scriptworker/tree/cot-verify-test).

To test manually, install scriptworker in your virtualenv via `python setup.py develop` (from the cot-verify branch), and:

    scriptworker/test/data/verify_cot.py SIGNING_TASK_ID

Where the signing task id is from a `date` non-l10n `Ns`, e.g.

    scriptworker/test/data/verify_cot.py OIy42W0cQPqGbYZEE9NduQ

L10n and other scriptworker instance support will come in later 1.x releases.

@jonasfj, does [`scriptworker.cot.verify`](https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/cot/verify.py) look sane?  The main function is [`verify_chain_of_trust`](https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/cot/verify.py#L1109).
The config values I reference are [here](https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/constants.py#L84-L234).

@JohanLorenzo, @lundjordan : this patch is 161k, *before* I add all the unittest coverage for `scriptworker.cot.verify`.  a) sorry!!!, b) if you prefer a different form of code review (meeting?), or if you want to split up the task, or if you have any questions... let me know.  I'd like to make this as smooth as possible.


lundjordan:
@JohanLorenzo I'm not losing again! I'm happy to take the whole thing or split it up. Will give it a look over first thing pacific time tomorrow.

@escapewindow 👍 nice one.

lundjordan:
sorry for delay here, hoping to finish by tomorrow.

scriptworker/utils.py
```diff
@@ -334,3 +389,47 @@ def format_json(data):
...
+        if parts.scheme not in rule['schemes']:
+            continue
+        if parts.netloc not in rule['netlocs']:
+            continue
+        for regex in rule['path_regexes']:
+            m = re.search(regex, path)
+            if m is None:
+                continue
+            result = callback(m)
+            if result is not None:
```
lundjordan: hm, I thought if there was no return statement in a function/method, the default is to return None?

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+        self.name = name
+        self.task_type = guess_task_type(name)
+        self.context = context
+        self.task_id = task_id or get_task_id(context.claim_task)
+        self.task = context.task
+        self.worker_impl = guess_worker_impl(self)  # this should be scriptworker
+        self.decision_task_id = get_decision_task_id(self.task)
+        self.links = []
+
+    def dependent_task_ids(self):
```
escapewindow: It doesn't ever shrink, but it's only the links that allow us to trace our task back to the tree.  So signing will have a much shorter chain than, say, balrog.

If we want to test the entire chain of trust, we can do so after the graph finishes with an external audit task.  We can't test the chain of trust artifact from a link further down the chain while the graph is still running, since those tasks haven't run yet.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+    def all_tasks(self):
+        """The list of task definitions for all `self.links` + `self`
+
+        Returns:
+            list of dicts: the task definitions for all links + self
+        """
+        return [self.task] + [x.task for x in self.links]
+
+    def is_try(self):
+        """Helper method to determine if any task in the chain is a try task.
```
escapewindow: Yup.  For signing, I have:

* release signing is restricted to [these repos](https://github.com/escapewindow/scriptworker/blob/8e9c08381e049fed24a111da1819f68e3f5e1c78/scriptworker/constants.py#L207-L214)
* nightly signing is restricted to [this superset of repos](https://github.com/escapewindow/scriptworker/blob/8e9c08381e049fed24a111da1819f68e3f5e1c78/scriptworker/constants.py#L220-L230)
* all other repos are able to use dep signing, including try.  That allows us to test the nightly graph on try, even with tasks that depend on signed binaries.  I'm going to push for the other *scripts to follow the same scope pattern, e.g. beetmover won't push into the release bucket from non-release repos.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+            raise CoTError("No single Link matches task_id {}!\n{}".format(task_id, self.dependent_task_ids()))
+        return links[0]
+
+
+# LinkOfTrust {{{1
+class LinkOfTrust(object):
+    """Each LinkOfTrust represents a task in the Chain of Trust and its status.
+
+    Attributes:
+        chain (ChainOfTrust): the ChainOfTrust object.
```
escapewindow: I'm not sure which function you're pointing at because github isn't giving me enough context, but I know a number of functions take a (chain, link) args even if they don't use them, because the function is called from the same calling function, and the calling function doesn't know what args the target functions take.  So they all get chain and link.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+    audit_handler.setFormatter(
+        logging.Formatter(fmt='%(asctime)s %(levelname)8s - %(message)s')
+    )
+    log.addHandler(audit_handler)
+    yield
+    log.removeHandler(audit_handler)
+
+
+# guess_worker_impl {{{1
+def guess_worker_impl(link):
```
escapewindow: Yeah, I don't know how else to know this.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+    """
+    # TODO support the rest of the worker impls... generic worker, taskcluster worker
+    return frozendict({
+        'docker-worker': verify_docker_worker_task,
+        'scriptworker': verify_scriptworker_task,
+    })
+
+
+# guess_task_type {{{1
+def guess_task_type(name):
```
escapewindow: Yeah. I debate whether I should pass in info from the chain/link or the entire object, but either way clean arch functions > methods.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+    Raises:
+        CoTError: on inability to determine the worker implementation
+    """
+    worker_impls = []
+    task = link.task
+    name = link.name
+    errors = []
+
+    if task['payload'].get("image"):
+        worker_impls.append("docker-worker")
```
escapewindow: hm, maybe.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+
+    Returns:
+        str: the task_type.
+
+    Raises:
+        CoTError: on invalid task_type.
+    """
+    parts = name.split(':')
+    task_type = parts[-1]
+    if task_type.startswith('build'):
```
escapewindow: I put in 'startswith' because I had some logic before that allowed for build0, build1, build2...
I don't have that anymore so I can take that out.  We may need something like that again.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+        bool: True if it's try
+    """
+    result = False
+    env = task['payload'].get('env', {})
+    if env.get("GECKO_HEAD_REPOSITORY"):
+        result = result or _is_try_url(task['payload']['env']['GECKO_HEAD_REPOSITORY'])
+    if env.get("MH_BRANCH"):
+        result = result or task['payload']['env']['MH_BRANCH'] == 'try'
+    if task['metadata'].get('source'):
+        result = result or _is_try_url(task['metadata']['source'])
```
escapewindow: I don't think you have to have an MH_BRANCH defined in a try task.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+                asyncio.ensure_future(
+                    download_artifacts(
+                        chain.context, [url], parent_dir=parent_dir,
+                        valid_artifact_task_ids=[task_id]
+                    )
+                )
+            )
+            await raise_future_exceptions(async_tasks)
+        except Exception:
+            log.exception("boo")
```
escapewindow: Hm.  I don't remember.  I'll go back to the docstring behavior, good catch.

scriptworker/cot/verify.py
lundjordan: yep sorry, forgot to delete this comment.

scriptworker/cot/verify.py
lundjordan: this is the class I was referring to. didn't seem to be a chain attr.

scriptworker/cot/verify.py
lundjordan: ah, okay. sgtm.

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+        CoTError: on failure.
+    """
+    cot = link.cot
+    task = link.task
+    errors = []
+    image_hash = cot['environment']['imageHash']
+    if isinstance(task['payload'].get('image'), dict):
+        # Using pre-built image from docker-image task
+        docker_image_task_id = task['extra']['chainOfTrust']['inputs']['docker-image']
+        if docker_image_task_id != task['payload']['image']['taskId']:
```
lundjordan: yeah. thanks for the link

scriptworker/cot/verify.py
```diff
@@ -0,0 +1,1141 @@
...
+        if value != runtime_defn[key]:
+            errors.append("{} {} {} differs!\n graph: {}\n task: {}".format(
+                task_link.name, task_link.task_id, key,
+                pprint.pformat(value), pprint.pformat(runtime_defn[key])
+            ))
+    raise_on_errors(errors)
+
+
+# verify_firefox_decision_command {{{1
+def verify_firefox_decision_command(decision_link):
```
escapewindow: Yes.  I agree 100%.

1. Enable chain of trust verification for dep signing jobs, and make them visible in treeherder.  That way chain of trust bustage is viewable per-push rather than waiting til nightlies or release
2. Put in these rules, and then work towards simplifying them.  Catlee and I were talking and we're thinking we can simplify decision task command logic if we put more logic into the docker image.

But as is, we have to have rules around the decision task, otherwise anyone with the scopes can launch any decision task with any arbitrary env and commandline, which can result in any arbitrary graph.

scriptworker/task.py
```diff
@@ -39,6 +39,58 @@ def worst_level(level1, level2):
     return level1 if level1 > level2 else level2
 
 
+# get_task_id {{{1
+def get_task_id(claim_task):
+    """Given a claim_task json dict, return the taskId.
+
+    Args:
+        claim_task (dict): the claim_task dict.
```
escapewindow: Possibly.  `claim_task` has taskId info and is a superset of `task`.  `task` has the task definition.  If we want all the same arg, we can default to chain or link or context because those have `claim_task` and `task` as properties, or in their context property's properties.  I debated this for a minute, then went with the simplest arg and a docstring.

scriptworker/cot/verify.py
escapewindow: https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/cot/verify.py#L503

escapewindow:
Feedback+ for 8e9c083.
I'm going to be landing review comment fixes, and I still have the multisign branch i'd like to merge in, and I still need to add test coverage.  I'm thinking I'm going to be working on other branches, then squash merge into cot-verify as I'm ready to reduce mail spam from this PR, then release 1.0.0 once I merge into master.

scriptworker/cot/verify.py
lundjordan: wow, sorry, I didn't put the link in: https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/cot/verify.py#L103

hm, so I'm confused. It looks like you don't have a chain attr in LinkOfTrust class definition. But you bind the attr to a LinkOfTrust live object. Should we not put `chain = None` after this line: https://github.com/escapewindow/scriptworker/blob/cot-verify/scriptworker/cot/verify.py#L115 ? 

or am I getting confused? :)

escapewindow: We can.  The difference is an `AttributeError` when we reference `link.chain` or an `AttributeError` when we reference `link.chain.$attr`.  I preferred the earlier `AttributeError`, BUT it looks like we never use `link.chain` so I'm going to just tear it out :)

scriptworker/utils.py
escapewindow: This is in a for loop, so a `return` will exit the for loop.
I don't want to exit the for loop until I get a non-None result.

escapewindow: Ah, looks like I can see more context with the `files` link above, just not in the main issue screen.

lundjordan: 😊 yep, PEBKAC

escapewindow:
I just merged in the `multisign` and `cot-verify-test` branches.
- 100% coverage
- multiple signing formats per file via `upstreamArtifacts`

I want to fix the signing blue's on `date` before releasing 1.0.0[b1?]

I also want to add support for other scriptworker instance types and full docs, but it's unclear if that'll be 1.0.0final or a dot release.

escapewindow:
I:
* fixed the signing timeouts.  These were uncaught aiohttp exceptions on upload.  The fix was twofold: reenabling max connections (we were opening 50some uploads simultaneously) and catching/retrying those aiohttp exceptions.
* added more debugging logging for the above.
* moved `audit_log_handler` to `scriptworker.log.contextual_log_handler` in case I wanted to add contextual logging for the timeouts
* fixed the bustage from https://hg.mozilla.org/integration/autoland/rev/2168cbd4b154#l2.57
 * added a new docker image to the allow list
 * allowed for a nonexistent docker-image `payload.command`
 * allowed for new docker-image env vars
* started adding a new docs framework, which i'll have to flesh out.

Merging and shipping 1.0.0b1.  I'll update docs and start pushing towards cot-enabled other scriptworker instance types.

PR #29, opened 2016-11-23
Bug 1314840 monitoring

srfraser:
Intended initially for the signing scriptworkers,
but hopefully generic enough to be useful.

The entrypoint /usr/local/bin/nagios_check_file_ages can
report on files being too old or nonexistant.

/usr/local/bin/nagios_check_pending_tasks can alert if there
are too many pending tasks for that provisionerId and workerType

Nagios has its own directory for plugins, but python installs
these in /usr/local/bin - unsure at present whether this will
prevent nagios from using them, whether symlinks would be enough
if so.

### DIFF
--- a/scriptworker/monitoring/nagios_file_age_check.py
+++ b/scriptworker/monitoring/nagios_file_age_check.py
@@ -143,6 +143,3 @@ def run_file_age_checks():
 
     print("{0}\n{1}\n".format(service_output, "\n".join(sorted(messages))))
     sys.exit(exit_code)
-
-if __name__ == '__main__':
-    run_file_age_checks()
--- a/scriptworker/monitoring/nagios_file_age_check.py
+++ b/scriptworker/monitoring/nagios_file_age_check.py
@@ -4,7 +4,7 @@
 Checks for the existence and age of specific files, and that they are not
 too old.
 
-Allow 'old' to be defined as warning and critical arguments. 
+Allow 'old' to be defined as warning and critical arguments.
 
 You can either:
 - Specify one file to check
@@ -70,7 +70,7 @@ def file_age_check(filename, warning, critical, optional):
         return STATUS_CODE['CRITICAL'], msg
     else:
         msg = "{0} is ok, {1}/{2} seconds old".format(
-            filename, int(age), max_age)
+            filename, int(age), critical)
         return STATUS_CODE['OK'], msg
 
 
--- a/scriptworker/monitoring/nagios_file_age_check.py
+++ b/scriptworker/monitoring/nagios_file_age_check.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python
+
+"""
+Checks for the existence and age of specific files, and that they are not
+too old.
+
+Allow 'old' to be defined as warning and critical arguments. 
+
+"""
+import os
+import sys
+import time
+import argparse
+
+# Nagios plugin exit codes
+STATUS_CODE = {
+    'OK': 0,
+    'WARNING': 1,
+    'CRITICAL': 2,
+    'UNKNOWN': 3,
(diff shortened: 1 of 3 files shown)


scriptworker/monitoring/nagios_file_age_check.py
```diff
@@ -0,0 +1,145 @@
...
+
+Allow 'old' to be defined as warning and critical arguments.
+
+You can either:
+- Specify one file to check
+- Give the path to a file, that contains a newline-separated list
+  of the files to check.
+
+This script will produce a multi-line nagios report in either case.
+
```
escapewindow: The top level docstring should include attributes of the file: these are module-level variables.  For instance, `STATUS_CODE` would go here.
http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

scriptworker/monitoring/nagios_file_age_check.py
```diff
@@ -0,0 +1,145 @@
...
+DEFAULT_CRITICAL = 60
+
+
+def file_age_check(filename, warning, critical, optional):
+    """file_age_check
+
+    Checks the age and existence of a given filename
+
+    Arguments:
+      - filename: a string, containing a full path
```
escapewindow: I'm not sure if the dashes are needed or if they'll throw off RTD.  One way to test is to 
1. create a virtualenv using `requirements-docs.txt`
2. `cd docs; make html`
3. open `_build/html/scriptworker.html` in your browser, and look at the formatting of these docstrings in the docs.

If they look ok, I'm not going to push too hard.  If they look odd or jarringly different from the other docstrings, the formatting from http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html is:

```
        Args:
            param1 (str): Description of `param1`.
            param2 (:obj:`int`, optional): Description of `param2`. Multiple
                lines are supported.
            param3 (:obj:`list` of :obj:`str`): Description of `param3`.
```

setup.py
```diff
@@ -67,6 +67,8 @@ def run_tests(self):
             "scriptworker = scriptworker.worker:main",
             "create_initial_gpg_homedirs = scriptworker.gpg:create_initial_gpg_homedirs",
             "rebuild_gpg_homedirs = scriptworker.gpg:rebuild_gpg_homedirs",
+            "nagios_check_pending_tasks = scriptworker.monitoring.nagios_pending_tasks:query_pending_task_count",
+            "nagios_check_file_ages = scriptworker.monitoring.nagios_file_age_check:run_file_age_checks",
```
escapewindow: This is awesome.
However, we're adding a new subdirectory, and by default we're going to be missing the scriptworker/monitoring directory in the pypi package.

We need to add `scriptworker.monitoring` to the [`packages` list](https://github.com/mozilla-releng/scriptworker/blob/3269a027c20e27a6fd5efffd20310510a1f7101f/setup.py#L62)

To test, `python setup.py sdist` . You'll get a `dist/scriptworker-VERSION.tar.gz` that you can extract or `tar tvf` to view the contents.  Make sure `scriptworker/monitoring/*.py` are included.  You probably need to touch and `git add` a `scriptworker/monitoring/__init__.py` for that to work.

escapewindow: A side effect of this will be a reduction in overall test coverage because there are no tests in this PR.

We can fix this in a couple ways:
1. add tests.  this is extra work, but having tests seems good... let me know if you need a hand with this.
2. these scripts import from scriptworker, but we can still potentially install them via puppet http://hg.mozilla.org/build/puppet/file/tip/modules/scriptworker/files , bypassing the package coverage requirement.

escapewindow:
Simon's planning on moving this to puppet.

PR #32, opened 2016-12-02
address gpg homedir race conditions

escapewindow:
PR notes: we had some bustage in gpg homedir creation.  This PR tries to deal with the symptoms I saw:

1. When the git repo was already up to date, I saw this stackdump:

> 2016-11-30T17:58:45     INFO - Already up-to-date.
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=False>>
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/tools/python35/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/tools/python35/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/tools/python35/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object

This isn't harmful, but it's confusing seeing this error when trying to debug.  I believe closing the event loop outside of the `if` block will remove this.

2. Error about copying a nonexistent `trustdb.db` into `/builds/scriptworker/gpg`

I believe this is because the cron job started a new `rebuild_gpg_homedirs` before scriptworker finished copying the contents of `gpg.tmp/` into `gpg/`.  I modified the lockfile logic to allow for 3 states: unlocked, locked (a process is creating the gpg homedirs in `gpg.tmp/`), and ready (a process has created the gpg homedirs in `gpg.tmp/`, and scriptworker needs to copy those into `gpg/`)

3. I was already worried about re-running `create_initial_gpg_homedirs` via puppet.  It was weird having two entry points that did nearly the same thing (`rebuild_gpg_homedirs` rebuilds the homedirs but puts the worker homedirs in `/builds/scriptworker/gpg.tmp/`; `create_initial_gpg_homedirs` does rebuilds the homedirs and replaces `/builds/scriptworker/gpg/` with the contents).  I played with the idea of having *three* gpg homedir entry points, one to rebuild `~/.gnupg` for git commit signature verification which puppet could run on pubkey change, one for replacing `/builds/scriptworker/gpg/` for initial setup, and one for populating `/builds/scriptworker/gpg.tmp/` via cron.

I finally settled on removing `create_initial_gpg_homedirs`.  If puppet runs `rebuild_gpg_homedirs`, that's one less thing to test, and one less thing that can stomp on things at runtime.  I was able to do this by moving the check for `/builds/scriptworker/gpg.tmp/` at the beginning of the loop, rather than at the end, so scriptworker would populate `/builds/scriptworker/gpg/` at first run if appropriate.

I also decided I wanted to test a little less on production, so I added a `Dockerfile.gnupg` with a readme update.  It's possible to test `rebuild_gpg_homedirs` on that.


escapewindow:
There's a corresponding puppet patch in [bug 1321748](https://bugzilla.mozilla.org/show_bug.cgi?id=1321748)

scriptworker/worker.py
```diff
@@ -82,12 +82,13 @@
...
     tmp_gpg_home = get_tmp_base_gpg_home_dir(context)
-    lockfile = context.config['gpg_lockfile']
-    await run_loop(context)
-    await asyncio.sleep(context.config['poll_interval'])
-    if os.path.exists(tmp_gpg_home) and not os.path.exists(lockfile):
+    state = is_lockfile_present(context, "scriptworker", logging.DEBUG)
+    if os.path.exists(tmp_gpg_home) and state == "ready":
         overwrite_gpg_home(tmp_gpg_home, context.config['base_gpg_home_dir'])
         rm(tmp_gpg_home)
+        rm_lockfile(context)
```
lundjordan: I wonder if the removal of the lockfile here should be protected under a `finally:` block. just in case we run into an exception while copying/rm gpg home.

docker/Dockerfile.gnupg
```diff
@@ -0,0 +1,32 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && apt-get install -y \
+    python3 \
```
escapewindow: Because I based this on @srfraser's Dockerfile which used ubuntu:16.04 :)

... Sure, looks easy enough trying to use a python3 base instead.

scriptworker/gpg.py
```diff
@@ -1441,31 +1443,7 @@ def _update_git_and_rebuild_homedirs(context, basedir=None):
         # Get rid of spurious event_loop errors.  Since this function is only
```
escapewindow: Good catch. Moving them down to above the new `event_loop.close()` call.

PR #36, opened 2016-12-02
stop uploading all the downloaded chain of trust artifacts

### LINKED ISSUE #35: stop uploading artifacts from other tasks
What this means, in practice, is:

a. the audit log should go in the `task_log_dir`; maybe i should rename it to cot_audit.log or something?
b. the downloaded cot artifacts should go in `work_dir`
c. signingscript will need to use the artifacts from the `work_dir` location rather than the `artifact_dir` location.

escapewindow:
We have pointers to the existing artifacts in other tasks, and now the
log is more verbose around what we're doing.  I think we can make the
log more readable, but I think we have a good coverage of information at
least.

Fixes #35.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -35,7 +35,6 @@ class ChainOfTrust(object):
 
     Attributes:
         context (scriptworker.context.Context): the scriptworker context
-        cot_dir (str): the local path containing this link's artifacts
         decision_task_id (str): the task_id of self.task's decision task
         links (list): the list of ``LinkOfTrust``s
         name (str): the name of the task (e.g., signing)
@@ -135,7 +134,7 @@ class LinkOfTrust(object):
         self.context = context
         self.task_id = task_id
         self.cot_dir = os.path.join(
-            context.config['artifact_dir'], 'public', 'cot', self.task_id
+            context.config['work_dir'], 'cot', self.task_id
         )
 
     def _set(self, prop_name, value):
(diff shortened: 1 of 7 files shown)


scriptworker/cot/verify.py
```diff
@@ -540,7 +551,7 @@ def find_task_dependencies(task, name, task_id):
         CoTError: on failure.
     """
     link = chain.get_link(task_id)
-    log.debug("Verifying {} is in {} cot artifacts...".format(path, task_id))
+    log.debug(" Verifying {} is in {} cot artifacts...".format(path, task_id))
```
JohanLorenzo: Bikeshedding: If I understand correctly, a space describes a debug type of log. Is that correct? If so, I wonder if we couldn't write a Formatter that adds it for us.

escapewindow: We're able to add a Formatter per *level* per logger? I was completely unaware :)
This is just me playing with text formatting to try to make these audit logs more readable.  I imagine there's a lot we can do with wording and formatting to improve this, but this was a quick first stab at it.

JohanLorenzo: Yup! We can use something like:
```py
import logging
import sys

class DebugSpacedFormatter(logging.Formatter):
    def format(self, record):
        if record.levelno == logging.DEBUG:
            record.msg = ' {}'.format(record.msg)
        return logging.Formatter.format(self, record)

logging_handler = logging.StreamHandler(sys.stdout)
logging_handler.setFormatter(DebugSpacedFormatter())
logging.root.addHandler(logging_handler)
logging.root.setLevel(logging.DEBUG)


log = logging.getLogger()
log.debug("that's a debug that'll get an extra space")
log.info("that's an info")
```

Issue #34, opened 2016-12-02
upload_artifacts stackdump


escapewindow:
On signing-linux-1, at 2016-12-02T17:31:40 UTC, we appear to have hit a fatal issue in `upload_artifacts`:

```
Traceback (most recent call last):
  File "/builds/scriptworker/lib/python3.5/site-packages/aiohttp/connector.py", line 608, in _create_d
irect_connection
    local_addr=self._local_addr)
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 713, in create_connection
    sock, protocol_factory, ssl, server_hostname)
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 736, in _create_connection_transport
    yield from waiter
  File "/tools/python35/lib/python3.5/asyncio/futures.py", line 361, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/tools/python35/lib/python3.5/asyncio/tasks.py", line 296, in _wakeup
    future.result()
  File "/tools/python35/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
ConnectionResetError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/builds/scriptworker/lib/python3.5/site-packages/aiohttp/connector.py", line 306, in connect
    yield from self._create_connection(req)
  File "/builds/scriptworker/lib/python3.5/site-packages/aiohttp/connector.py", line 585, in _create_c
onnection
    transport, proto = yield from self._create_direct_connection(req)
  File "/builds/scriptworker/lib/python3.5/site-packages/aiohttp/connector.py", line 631, in _create_d
irect_connection
    (req.host, req.port, exc.strerror)) from exc
aiohttp.errors.ClientOSError: [Errno None] Can not connect to taskcluster-public-artifacts.s3-us-west-2.amazonaws.com:443 [None]

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/builds/scriptworker/lib/python3.5/site-packages/scriptworker/worker.py", line 101, in main
    loop.run_until_complete(async_main(context))
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 387, in run_until_complete
    return future.result()
  File "/tools/python35/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/tools/python35/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
[...]

PR #37, opened 2016-12-07
catch aiohttp exceptions during upload_artifacts.

escapewindow:
In #34 we're hitting an uncaught exception during `upload_artifacts`.
This patch catches it, logs the error, and marks the task as
`resource-unavailable`.

@JohanLorenzo @lundjordan sorry for all the review requests! I'm hoping that a lot of smaller reviews is easier than one big one.

### DIFF
--- a/scriptworker/test/test_worker.py
+++ b/scriptworker/test/test_worker.py
@@ -158,6 +158,8 @@ def test_mocker_run_loop_noop(context, successful_queue, event_loop, mocker):
     'upload_artifacts', ScriptWorkerException, ScriptWorkerException.exit_code
 ), (
     'upload_artifacts', aiohttp.errors.DisconnectedError, STATUSES['resource-unavailable']
+), (
+    'upload_artifacts', aiohttp.errors.ClientError, STATUSES['resource-unavailable']
 )))
 def test_mocker_run_loop_exception(context, successful_queue, event_loop,
                                    mocker, func_to_raise, exc, expected):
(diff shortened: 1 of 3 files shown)


scriptworker/worker.py
```diff
@@ -65,6 +66,9 @@
             except ScriptWorkerException as e:
                 status = worst_level(status, e.exit_code)
                 log.error("Hit ScriptWorkerException: {}".format(str(e)))
+            except (aiohttp.errors.DisconnectedError, aiohttp.errors.ClientError) as e:
+                status = worst_level(status, STATUSES['resource-unavailable'])
+                log.error("Hit aiohttp error: {}".format(str(e)))
```
JohanLorenzo: Nit: `str(e)` shouldn't be necessary as the conversion is already done by `format()`

scriptworker/test/test_worker.py
```diff
@@ -150,9 +152,15 @@ def test_mocker_run_loop_noop(context, successful_queue, event_loop, mocker):
...
 
-@pytest.mark.parametrize("func_to_raise", ['run_task', 'upload_artifacts'])
-def test_mocker_run_loop_exception(context, successful_queue,
-                                   event_loop, mocker, func_to_raise):
+@pytest.mark.parametrize("func_to_raise,exc,expected", ((
+    'run_task', ScriptWorkerException, ScriptWorkerException.exit_code
+), (
+    'upload_artifacts', ScriptWorkerException, ScriptWorkerException.exit_code
+), (
+    'upload_artifacts', aiohttp.errors.DisconnectedError, STATUSES['resource-unavailable']
```
JohanLorenzo: Should we add a case for `aiohttp.errors.ClientError`?

PR #40, opened 2016-12-10
Support beetmover and balrog in chain of trust verification

escapewindow:
@lundjordan @MihaiTabara you will need this PR to land before scriptworker can support beetmover and balrog.  I haven't tested everything all together, but hopefully this Just Works :)

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -123,6 +123,8 @@ DEFAULT_CONFIG = frozendict({
 
     # scriptworker identification
     "scriptworker_worker_types": (
+        "balrogworker-v1",
+        "beetmoverworker-v1",
         "signing-linux-v1",
     ),
     "scriptworker_provisioners": (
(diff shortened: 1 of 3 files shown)


escapewindow:
That's a good idea, but:

* the various scriptworkers have to verify other scriptworker instance types. for instance, a balrog task needs to verify upstream, which includes beetmover and signing.  beetmover needs to verify signing. we can require that each scriptworker instance type installs all the upstream scriptworker script modules, but that gets a little unwieldy.  plus, there has to be logic to figure out which implementation to use to verify each instance type.
* similarly, `verify_cot` runs locally against any task in the graph, so you would have to install every scriptworker module locally to be able to run it and verify all upstream tasks of any arbitrary scriptworker task.
* i think it's good to have all the cot verification in one place, where we can inspect it and audit it and spin it out into its own module.

If we're able to make all the scriptworker task definitions to look the same (we might already be there once they're all ported to use scriptworker 1.0.0), maybe we only have to verify it's a scriptworker task, and we don't care what type during cot verification.  that would simplify things.

I'm presenting the other side because I think there are reasons, but I don't want to discourage you from continuing to have opinions and continuing to speak up about them.  At the least it brings up your questions about the implementation, and you've brought up good points before (and will continue to do so) that lead to improvements in the code and process.  Thanks for adding your ideas!

PR #45, opened 2017-01-04
Bug 1326436

escapewindow:
This PR adds the new docker-image sha, per [bug 1326436](https://bugzilla.mozilla.org/show_bug.cgi?id=1326436).

There's also a ride-along doc update (adding .html to the chain_of_trust link).

@lundjordan @JohanLorenzo if you have a moment :)

### DIFF
--- a/docs/new_instance.md
+++ b/docs/new_instance.md
@@ -4,7 +4,7 @@ We don't yet have a scriptworker provisioner, so spinning up new instances of a 
 
 ## signing scriptworker
 ### gpg keypair
-Generate and sign a gpg keypair for cltsign@**fqdn**, per [these docs](chain_of_trust#adding-new-worker-gpg-keys).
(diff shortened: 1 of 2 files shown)


docs/new_instance.md
```diff
@@ -4,7 +4,8 @@ We don't yet have a scriptworker provisioner, so spinning up new instances of a
 
 ## signing scriptworker
 ### gpg keypair
-Generate and sign a gpg keypair for cltsign@**fqdn**, per [these docs](chain_of_trust#adding-new-worker-gpg-keys).
+Generate and sign a gpg keypair for cltsign@**fqdn**, per [these docs](chain_of_trust.html#adding-new-worker-gpg-keys).
+http://scriptworker.readthedocs.io/en/latest/chain_of_trust.html#adding-new-worker-gpg-keys
```
JohanLorenzo: I guess the raw link here is not necessary, and will be generated via the markdown link above. Is that correct?

escapewindow: Good catch! Thank you!

PR #46, opened 2017-01-06
Enhance error message when creds are missing

JohanLorenzo:
If you don't specify credentials (either because you forgot, or because you didn't paste things correctly in `secrets.json`), you end up with: `scriptworker.yaml credentials: type <class 'NoneType'> is not <class 'frozendict.frozendict'>!`

This patch replaces with that one instead:
```
scriptworker.yaml credentials needs to be defined!
```

### DIFF
--- a/scriptworker/config.py
+++ b/scriptworker/config.py
@@ -30,6 +30,7 @@ CREDS_FILES = (
 # Based on
 # https://github.com/taskcluster/taskcluster-queue/blob/ced3b7bd824b445fd33ce0deb5de87a65f02b8b3/src/api.js#L94
 _GENERIC_ID_REGEX = re.compile(r'^[a-zA-Z0-9-_]{1,22}$')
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Time for me to add a new test

JohanLorenzo:
I might have found a small bug while writing the test, but I'd like to check that with you @escapewindow . If a key is not present in `config`, `check_config` doesn't append an error message. From the code of `create_config`, I believe this should not happen by design. Do I get that correctly?

escapewindow:
That's correct.  The quick fix is probably something like

```python
missing_keys = set(DEFAULT_CONFIG.keys()) - set(config.keys())
if missing_keys:
    messages.append("Missing config keys {}!".format(missing_keys))
```

Want me to add or do you want to?

JohanLorenzo:
I know what test to add. I can add this along this patch

PR #50, opened 2017-01-10
Multi tags

escapewindow:
to address @JohanLorenzo's review comments in #48.
We have unit test coverage, but no docker manual test, yet.  I'll do that now.

### DIFF
--- a/scriptworker/gpg.py
+++ b/scriptworker/gpg.py
@@ -21,7 +21,6 @@ import logging
 import os
 import pexpect
 import pprint
-import re
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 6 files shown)


scriptworker/gpg.py
```diff
@@ -1137,130 +1136,60 @@ def rebuild_gpg_home_signed(context, real_gpg_home, my_pub_key_path,
...
-        for k, v in BAD.items():
-            if line.startswith(k):
-                messages.append(v)
-                continue
-    if not status:
-        messages.append("No trusted signature!")
-    if messages:
-        raise ScriptWorkerGPGException("\n".join(messages + ["output:", output]))
+    proc = await exec_function(
+        'git', "log", "-n1", "--format=format:%H", ref, cwd=path,
```
JohanLorenzo: Nit: A simpler command can be `git show-ref --hash $A_TAG`

scriptworker/test/test_gpg.py
```diff
@@ -692,7 +599,15 @@ def test_verify_signed_git_commit_output_exception(output):
 
 
 @pytest.mark.asyncio
-async def test_get_git_revision_exception(mocker):
+async def test_get_latest_tag():
+    parent_dir = os.path.dirname(__file__)
+    expected = subprocess.check_output(["git", "describe", "--abbrev=0"], cwd=parent_dir)
```
JohanLorenzo: Do we have a way to systematically know the tag names and the number of tags in the test suite? Like either create a new repo and add two tags and see which one comes up?

escapewindow: The test suite currently uses the current scriptworker repo as the testing ground, so no.  I did add a TODO above that says we should create a test git repo tarball with deterministic results.

scriptworker/gpg.py
escapewindow: `git show` gives me revision `0934d8c59949b830813cbea81436ba12bc88d68c`
`git show-ref --hash HEAD` gives me `f3d70ad12d1c59c52f331c0cc7ca7cd1f6301e50`

escapewindow: `git show HEAD` gives me revision `0934d8c59949b830813cbea81436ba12bc88d68c`

escapewindow: `git log -n1 --format=format:%H HEAD` gives me `0934d8c59949b830813cbea81436ba12bc88d68c`

JohanLorenzo: I wonder `git show-ref` doesn't pick the most recent HEAD across every branch. The documentation isn't clear about that. Thanks for pointing out the issue!

scriptworker/test/test_gpg.py
escapewindow: I'll merge this PR and file an issue for the test git repo.  It's easy to add; maybe I'll make this a good first bug.

PR #53, opened 2017-01-12
Make failures report on Treeherder

JohanLorenzo:
~[WIP: _Tests are on their way_]~

This requires:
  1. the presence of a public/logs/live_backing.log
  2. the file to be gzip'd
  3. and served as a gzip on S3

Edit: Depends on #59

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -86,7 +86,7 @@ async def upload_artifacts(context):
 def compress_artifact_if_supported(artifact_path):
     """Compress artifacts with GZip if they're known to be supported.
 
-    This replaces the artifact given by a gzip binary. Nothing happens if the artifact is already zip'd
+    This replaces the artifact given by a gzip binary.
 
     Args:
         artifact_path (str): the path to compress
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -98,7 +98,7 @@ def compress_artifact_if_supported(artifact_path):
     log.debug('"{}" is encoded with "{}" and has mime/type "{}"'.format(artifact_path, encoding, content_type))
 
     if encoding is None and content_type in _GZIP_SUPPORTED_CONTENT_TYPE:
-        log.debug('"{}" can be gzip\'d. Compressing...'.format(artifact_path))
+        log.info('"{}" can be gzip\'d. Compressing...'.format(artifact_path))
         with open(artifact_path, 'rb') as f_in:
             text_content = f_in.read()
 
@@ -106,7 +106,7 @@ def compress_artifact_if_supported(artifact_path):
             f_out.write(text_content)
 
         encoding = 'gzip'
-        log.info('Compressed "{}" with gzip'.format(artifact_path))
+        log.info('"{}" compressed'.format(artifact_path))
     else:
         log.debug('"{}" is not supported for compression.'.format(artifact_path))
 
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -1,0 +1,294 @@
+import aiohttp
+import arrow
+import asyncio
+import gzip
+import logging
+import mimetypes
+import os
+
+from urllib.parse import unquote, urljoin
+
+from scriptworker.client import validate_artifact_url
+from scriptworker.exceptions import ScriptWorkerRetryException
+from scriptworker.task import get_task_id, get_run_id, get_decision_task_id
+from scriptworker.utils import download_file, filepaths_in_dir, raise_future_exceptions, retry_async
+
+
+log = logging.getLogger(__name__)
+
+
+GZIP_SUPPORTED_CONTENT_TYPE = ('text/plain', 'application/json', 'text/html', 'application/xml')
+EXTENSIONS_TO_FORCE_TO_PLAIN_TEXT = ('.asc', '.log')
+
+
+# create_artifact {{{1
+async def create_artifact(context, path, target_path, content_type, content_encoding, storage_type='s3', expires=None):
+    """Create an artifact and upload it.
+
+    This should support s3 and azure out of the box; we'll need some tweaking
+    if we want to support redirect/error artifacts.
+
+    Args:
+        context (scriptworker.context.Context): the scriptworker context.
+        path (str): the path of the file to upload.
+        target_path (str):
+        content_type (str): Content type (MIME type) of the artifact. Values can be found via
+            scriptworker.artifacts.guess_content_type_and_encoding()
+        content_encoding (str): Encoding (per mimetypes' library) of the artifact. None is for no encoding. Values can
+            be found via scriptworker.artifacts.guess_content_type_and_encoding()
+        storage_type (str, optional): the taskcluster storage type to use.
+            Defaults to 's3'
+        expires (str, optional): datestring of when the artifact expires.
+            Defaults to None.
+
+    Raises:
+        ScriptWorkerRetryException: on failure.
+    """
+    payload = {
+        "storageType": storage_type,
+        "expires": expires or get_expiration_arrow(context).isoformat(),
+        "contentType": content_type,
+    }
+    args = [get_task_id(context.claim_task), get_run_id(context.claim_task),
+            target_path, payload]
+
+    tc_response = await context.temp_queue.createArtifact(*args)
+    skip_auto_headers = [aiohttp.hdrs.CONTENT_TYPE]
+    log.info("uploading {path} to {url}...".format(path=path, url=tc_response['putUrl']))
+    with open(path, "rb") as fh:
+        with aiohttp.Timeout(context.config['artifact_upload_timeout']):
+            async with context.session.put(
+                tc_response['putUrl'], data=fh, headers=craft_artifact_put_headers(content_type, content_encoding),
+                skip_auto_headers=skip_auto_headers, compress=False
+            ) as resp:
+                log.info(resp.status)
+                response_text = await resp.text()
+                log.info(response_text)
+                if resp.status not in (200, 204):
+                    raise ScriptWorkerRetryException(
+                        "Bad status {}".format(resp.status),
+                    )
+
+
+# retry_create_artifact {{{1
+async def retry_create_artifact(*args, **kwargs):
+    """Retry create_artifact() calls.
+
+    Args:
+        *args: the args to pass on to create_artifact
+        **kwargs: the args to pass on to create_artifact
+    """
+    await retry_async(
+        create_artifact,
+        retry_exceptions=(
+            ScriptWorkerRetryException,
+            aiohttp.errors.DisconnectedError,
+            aiohttp.errors.ClientError
+        ),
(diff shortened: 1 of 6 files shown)


scriptworker/log.py
```diff
@@ -88,7 +88,8 @@ def get_log_filenames(context):
     Returns:
         tuple: log file path, error log file path
     """
-    log_file = os.path.join(context.config['task_log_dir'], 'task_output.log')
+    # XXX Even though our logs aren't live, Treeherder looks for live_backing.log to show errors in failures summary
+    log_file = os.path.join(context.config['task_log_dir'], 'live_backing.log')
```
escapewindow: I think we should use `task_error.log` as the `live_backing.log`.
Currently `task_error.log` contains `STDERR` + `STDOUT`; `task_output.log` has `STDOUT`.  If we use `task_output.log`, we'll drop everything that goes to `STDERR`.

Also, in practice, I haven't found `task_output.log` useful.  We could actually move to one combined task log file.  If you're up for it, let's do it.  If you want to limit scope, we can stay with `live_backing.log` + a mostly useless `task_output.log`.

scriptworker/task.py
```diff
@@ -229,16 +230,15 @@ def guess_content_type(path):
...
-    headers = {
-        aiohttp.hdrs.CONTENT_TYPE: tc_response['contentType'],
-    }
     skip_auto_headers = [aiohttp.hdrs.CONTENT_TYPE]
     log.info("uploading {path} to {url}...".format(path=path, url=tc_response['putUrl']))
     with open(path, "rb") as fh:
         with aiohttp.Timeout(context.config['artifact_upload_timeout']):
             async with context.session.put(
-                tc_response['putUrl'], data=fh, headers=headers,
+                tc_response['putUrl'], data=fh, headers=_craft_artifact_put_hearders(tc_response),
```
escapewindow: typo?  s,hearders,headers, ?

scriptworker/task.py
```diff
@@ -250,6 +250,29 @@ def guess_content_type(path):
...
+            text_content = f_in.read()
+
+        with gzip.open(artifact_path, 'wb') as f_out:
+            f_out.write(text_content)
+        log.info('Compressed "{}" with gzip'.format(artifact_path))
+    else:
+        log.debug('"{}" is NOT a plain text file, nothing to compress.'.format(artifact_path))
+
+
+def _craft_artifact_put_hearders(tc_response):
```
escapewindow: the typo matches here, which means we're internally consistent at least :)

scriptworker/task.py
```diff
@@ -229,16 +230,15 @@ def guess_content_type(path):
     }
     args = [get_task_id(context.claim_task), get_run_id(context.claim_task),
             target_path, payload]
+    _compress_if_text_plain(path, payload['contentType'])
```
escapewindow: I'm not 100% sure we should compress all artifacts if `text/plain`.  If this ends up compressing `chainOfTrust.json.asc`, do we automatically decompress on download or will we have to decompress before reading?

If we're not sure, we might want to do the compression when we copy the logs to the `upload_dir`.

scriptworker/task.py
```diff
@@ -250,6 +250,29 @@ def guess_content_type(path):
...
+        log.info('Compressed "{}" with gzip'.format(artifact_path))
+    else:
+        log.debug('"{}" is NOT a plain text file, nothing to compress.'.format(artifact_path))
+
+
+def _craft_artifact_put_hearders(tc_response):
+    headers = {
+        aiohttp.hdrs.CONTENT_TYPE: tc_response['contentType'],
+    }
+    if tc_response['contentType'] == 'text/plain':
```
escapewindow: same here.  Instead of changing the `contentType` here for all `text/plain`, we could go back to `guess_content_type` where we set the default contentType for all `.log` files to `text/plain`, and change that?  If we decide we want to gzip all `text/plain` artifacts no matter what, then we can keep this.

scriptworker/log.py
JohanLorenzo: I agree. In practice, I haven't noticed a the difference between the 2 files (except task_output.log contains the exit code at the end). I'll get rid task_output in this PR.

scriptworker/task.py
JohanLorenzo: Typo!

scriptworker/task.py
JohanLorenzo: Thank you autocompletion! 😃

scriptworker/task.py
JohanLorenzo: On my machine, /etc/mime.types maps `asc` to `application/pgp-signature`. As a consequence, we currently don't gzip it. For example:
```sh
 curl -I https://public-artifacts.taskcluster.net/SwE-nbJ_RQalWmmEfMbCuA/0/public/chainOfTrust.json.asc
```

I agree this may not be the case on every machine. I see several solutions:

* Either we don't gzip `asc` file. In this case, we have to:
  * document `/etc/mime.types/` is extensively used, and different machines may provide different uploads
  * make the puppet module fill `/etc/mime.types` for us. For the record, our CentOS VMs don't have this file.
* Or we gzip files that are listed (hardcoded?) in scriptworker. At the moment, I'm thinking of files like: logs, json, asc, html/xml. If we go this way, here are the implications:
  * **Browsers** will be able to gunzip *some files*, which means:
    * If you want to use curl or wget, you have to pass extra args or pipe them (respectively `--compressed` and `wget ... | gunzip -c -`). Documenting that behavior would be nice.
    * If you want browser to understand every file, you have to force `Content-Type` to `text/plain`. This doesn't cause any issues once the file is downloaded. 

I checkout out some existing jobs listed on Treeherder. The majority of them just gzip live_backing.log. However, I saw some of them compressing html or json files.

I currently prefer the second solution, for 4 reasons:

1. We save space (and bandwidth) on S3, so the cost of hosting artifacts is lower.
2. Less unexpected behaviors to document. 
3. Every text file is handled the same.
4. Thanks to specifying `text/plain` for artifacts like `asc`, we make them readable from the browser (instead of downloadable, which happens on my computer). Thus, less steps to see the content of these files, faster dev/troubleshooting! 

What do you think @escapewindow

scriptworker/task.py
JohanLorenzo: I agree, `guess_content_type` is a better central place for handling content-type and encoding. While testing [the arguments above](https://github.com/mozilla-releng/scriptworker/pull/53#discussion_r95971790), I realized I had to change one how a given content-type is handled at several places.

scriptworker/log.py
escapewindow: Thanks!
I think signing logging output goes to stderr, and stack traces as well.  A single unified log seems like a win.

scriptworker/task.py
escapewindow: I would love asc files to be text/plain, so we can read the chain of trust artifacts in the browser without downloading first.  If gzipping doesn't affect the artifact when we download from the queue (i.e., we don't have to gunzip before we can read it), then having asc files gzipped sounds like a win.  If we have to gunzip, then we'd have to change the scriptworker.cot.verify.download_cot to detect when an asc file is gzipped and gunzip (docker-worker asc files may not be gzipped, for example).

JohanLorenzo: Good news! No need to modify download_cot(). [aiohttp gunzip files for us](http://aiohttp.readthedocs.io/en/stable/client.html#binary-response-content) when downloaded (behavior tested locally).

escapewindow: \o/

scriptworker/log.py
JohanLorenzo: In order to make the reviews simpler, I made the renaming/deletion in #59

JohanLorenzo:
Hightlights about what's changed since your last review:

* Using one file is done in #59. #59 should be merged before this PR (otherwise Treeherder still won't look the correct location)
* Artifacts management (as in downloading, creating...) is done in a new package `scriptworker.artifacts`. **If we merge this PR, we'll need to upgrade the version to 1.1.0b1**. I don't mind rolling back this change and keep it until 1.0.0 is out.
* Tests related to compression were added. ⚠️ We don't check if the put request has the correct headers and content. This is due to a limitation of mocks in context managers (not yet supported). If you have a workaround, I'll be glad to make the changes.
* No test related to download_file() was added. I initially wanted to check gunzip was done, but the fake required a deeper knowledge of aiohttp (like knowing what attribute is decoded, in what order). I don't believe this is the right way to proceed, and we're testing more the library than our actual code. That's why I decided to leave the tests in `scriptworker.utils` as is.

escapewindow:
> Artifacts management (as in downloading, creating...) is done in a new package scriptworker.artifacts. If we merge this PR, we'll need to upgrade the version to 1.1.0b1. I don't mind rolling back this change and keep it until 1.0.0 is out.

As mentioned in IRC, [pre-1.0.0 is not stable](http://semver.org/#spec-item-5).  It isn't clear if that means 1.0.0 pre-alpha is locked, or it's locked at beta1 or rc1.  I'm fine being more strict or more loose at this point.  Once we hit 1.0.0 (when we're tier1 on central) I imagine we'll increment the major and minor versions fairly quickly.

To be more clear about this specific PR:  If we consider 1.0.0bX fair game for api changes, let's land pre-1.0.0.  If we think we'll need to bump the minor version for this PR, let's wait til 1.0.0 releases.

scriptworker/artifacts.py
```diff
@@ -0,0 +1,303 @@
...
+        artifact_path (str): the path to compress
+
+    Returns:
+        content_type, content_encoding (tuple):  Type and encoding of the file. Encoding equals 'gzip' if compressed.
+    """
+    content_type, encoding = guess_content_type_and_encoding(artifact_path)
+    log.debug('"{}" is encoded with "{}" and has mime/type "{}"'.format(artifact_path, encoding, content_type))
+
+    if encoding is None and content_type in _GZIP_SUPPORTED_CONTENT_TYPE:
+        log.debug('"{}" can be gzip\'d. Compressing...'.format(artifact_path))
```
JohanLorenzo: I'm good with turning this line into info. I'd keep the final log info as well. Otherwise let's say an error occurs after gziping, then I'm afraid people may wonder if files were gzip'd in the end of not. This may lead them to wrong assumptions about what's failing.

JohanLorenzo:
Conflicts happened when I rebased this PR against #59. Tests passing are a good sign I didn't mess up the resolution. I doubled checked the diff to see if there were leftovers of "log" of "filehandle" => we're good. 

Finally I ran a session of manual testing with https://tools.taskcluster.net/task-inspector/#TRZPZW65RO-adh6aMZb-1Q/ => the task failed as it should have (Google Play refused an existing APK). Logs were uploaded as text/plain + gzip and Treeherder correctly reported the error: https://treeherder.mozilla.org/#/jobs?repo=mozilla-aurora&filter-tier=3&revision=5cf9f0f65d201fc4631530c8f00e3a14887c1ad1&filter-searchStr=gps&selectedJob=70404051

PR #57, opened 2017-01-13
Delete cancelled or claimed tasks from azure queue.

### LINKED ISSUE #55: scriptworker isn't deleting cancelled tasks from the azure queue
Symptoms:
* we try to claim a task, but it's been claimed already (in fact, it's been cancelled); we should get a 409 here
* we ignore it, and move on.  we sleep 5
* repeat

there are dozens of beetmover cancelled tasks, so this is seriously slowing down our polling.  I'm going to delete the task from azure if we hit a 409.

escapewindow:
While testing the nightly migration on the `date` branch, we noticed
scriptworker was slogging through a lot of tasks that were "already
claimed by another worker".  These turned out to be cancelled tasks, and
they stayed in the Azure queue because we didn't delete them.

This patch deletes those tasks from the Azure queue on status code 409.
It also adds retries and catches in `find_task`, making it more robust.
This is already running on beetmoverworker-1, and we've cleared out a
backlog of cancelled tasks from the queue because of it.

This patch fixes #55.

### DIFF
--- a/scriptworker/poll.py
+++ b/scriptworker/poll.py
@@ -115,8 +115,11 @@ async def find_task(context, poll_url, delete_url, request_function):
 
     For a given poll_url/delete_url pair, get the xml from the poll_url.
     For each message in the xml, parse and try to claim the task.
-    Delete the message from the Azure queue whether the claim was successful
-    or not (error 409 on claim means the task was cancelled/expired/claimed).
+
+    Make a best effort to delete the message from the Azure queue if the task
+    was claimed, or if `claim_task` returns None (the task was
+    cancelled/expired/claimed).  If retrying this deletion fails, proceed and
+    clean it up the next round of polling.
 
     If the claim was successful, return the task json.
 
@@ -129,9 +132,8 @@ async def find_task(context, poll_url, delete_url, request_function):
         dict: the claimTask json
 
     Raises:
-        ScriptWorkerException: on failure of request_function
+        ScriptWorkerException: on polling failure
     """
-    # this can raise a ScriptWorkerException on failure
     xml = await request_function(context, poll_url)
     for message_info in parse_azure_xml(xml):
         try:
--- a/scriptworker/poll.py
+++ b/scriptworker/poll.py
@@ -14,8 +14,8 @@ import time
 import urllib.parse
 
 import taskcluster.exceptions
-from scriptworker.exceptions import ScriptWorkerException
-from scriptworker.utils import datestring_to_timestamp, load_json
+from scriptworker.exceptions import ScriptWorkerException, ScriptWorkerRetryException
+from scriptworker.utils import datestring_to_timestamp, load_json, retry_async
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


escapewindow:
@JohanLorenzo , @lundjordan if you have time+headspace to look, that would be awesome.

This patch is currently live on beetmoverworker-1, and helped us churn through a lot of cancelled tasks.  The US is out on Monday the 16th, so no need to rush on a review.

scriptworker/poll.py
```diff
@@ -121,15 +127,33 @@ def get_azure_urls(context):
 
     Returns:
         dict: the claimTask json
+
+    Raises:
+        ScriptWorkerException: on failure of request_function
     """
+    # this can raise a ScriptWorkerException on failure
```
JohanLorenzo: Nit: This comment duplicates the line above

scriptworker/poll.py
```diff
@@ -121,15 +127,33 @@ def get_azure_urls(context):
...
+        except ScriptWorkerRetryException:
+            continue
+        log.debug("Deleting from azure...")
+        delete_url = delete_url.replace("{{", "{").replace("}}", "}").format(**message_info)
+        try:
             response = await request_function(context, delete_url, method='delete', good=[200, 204])
             log.debug(response)
+        except ScriptWorkerException as exc:
+            log.error("Unable to delete from Azure: {}".format(exc))
+            pass
```
lundjordan: hm, do we want to raise this? The docs suggest we do.

I'm concerned since we swallow the ScriptWorkerRetryException exception in the retry_async(claim_task) call above and then again with the delete exception here, we may be having instances where we want to possibly sleep at the following line but never do: https://github.com/mozilla-releng/scriptworker/blob/e934a23443a053e1d37ab7c8d6d5a2d14213b465/scriptworker/worker.py#L48

could be parsing wrong though.

scriptworker/poll.py
```diff
@@ -121,15 +127,33 @@ def get_azure_urls(context):
 
     Returns:
         dict: the claimTask json
+
+    Raises:
+        ScriptWorkerException: on failure of request_function
```
lundjordan: hm, not sure if this is true. see below. Also, in the docs above the:

    "Delete the message from the Azure queue whether the claim was successful or not (error 409 on claim means the task was cancelled/expired/claimed)."

seems to be outdated now since we don't try to delete if we get a non 409 exception.

scriptworker/poll.py
escapewindow: We're retrying, and the side effect of not deleting is slowing down the next poll by 5 seconds, so I saw a `pass` as the lesser of evils:  Proceed with the task with a best effort deletion from Azure.

scriptworker/poll.py
escapewindow: We don't catch a `ScriptWorkerException` in the first `request_function` call.  Also, in `find_task`, we return None and then delete from Azure.  I could take a stab at fixing these docstrings; do you have a recommendation on how to word?

scriptworker/poll.py
escapewindow: Yeah, I wanted to point out that that was the one line where I saw a potential uncaught exception.

PR #58, opened 2017-01-15
Bugfix for bug 1330395 recursive freezevalues and unfreezevalues

### LINKED ISSUE #24: bug 1330395 - recursive `freeze_values` and `unfreeze_values`
We currently have `scriptworker.config.freeze_values` that takes a dict, and changes any list values to tuples, and any dict values to `frozendict`s.  This is useful to avoid inadvertent changes to runtime config.

This was sufficient as long as the config wasn't nested.  That isn't true anymore.

In several places in the tests I do `dict(deepcopy(config))` and do a for loop to change all `frozendict`s to dicts for json serialization.  A reverse function, to recursively unfreeze a `frozendict`, might be useful, as long as it's very carefully used.

We can either write our own recursive `freeze_values` or copy/modify mozharness' `ReadOnlyDict`:

### Write our own

* we'll need to freeze from the bottom up, since freezing a top level dict and then modifying its values won't work
* we'll need to unfreeze from the top down, since trying to unfreeze a child with a locked parent won't work

### Copy mozharness' `ReadOnlyDict`
MPL2, so easy to drop in.  We may want to allow for un-freezing, or make sure a `deepcopy` returns a read-write dict.

I don't expect to get to this before 1.0.0.  Filing in case someone else is interested.

escapewindow:
This fixes #24

### DIFF
--- a/scriptworker/config.py
+++ b/scriptworker/config.py
@@ -32,8 +32,54 @@ CREDS_FILES = (
 _GENERIC_ID_REGEX = re.compile(r'^[a-zA-Z0-9-_]{1,22}$')
 _VALUE_UNDEFINED_MESSAGE = "{path} {key} needs to be defined!"
 
+def freeze_values(d):
+    """Convert a dictionary's list values into tuples, and dicts into frozendicts.
+    A recursive function(bottom up conversion)
+    Args: The current value in the dictionary- can be a dictionary or list
+    ( or any other data type)"""
+
+    if isinstance(d,dict):
+        for k,v in d.items():
+            if isinstance(v,list):
+                freeze(v)
+                d[k]=tuple(v)
+            elif isinstance(v,dict):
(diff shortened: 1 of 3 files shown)


escapewindow:
@ray-pankaj , it looks like we need to merge the latest master changes into this branch.

To do this, you can

```
git checkout master
git pull https://github.com/mozilla-releng/scriptworker.git master
git checkout bugfix-for-bug-1330395-recursive-freezevalues-and-unfreezevalues
git merge master
git push
```

You may also have a "merge latest changes from `master` into this branch" button in this page... I don't.

If you need help with this, I can do the above.

ray-pankaj:
@escapewindow I have merged the two branches on my fork.

escapewindow:
Thanks!  Congratulations, you just fixed the bug.

PR #59, opened 2017-01-17
Use only one log file (live_backing.log)

JohanLorenzo:
Spin off that came out of [this discussion in #53](https://github.com/mozilla-releng/scriptworker/pull/53#discussion_r95844412). I split the changes in a new PR in order to make the review simpler.

PR also tested manually with https://tools.taskcluster.net/task-inspector/#ZXKOTGt1Qq6rUmPFDKfDmw/0

### DIFF
--- a/scriptworker/log.py
+++ b/scriptworker/log.py
@@ -79,35 +79,33 @@ async def pipe_to_log(pipe, filehandles=(), level=logging.INFO):
             break
 
 
-def get_log_filenames(context):
+def get_log_filename(context):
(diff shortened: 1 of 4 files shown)


scriptworker/log.py
```diff
@@ -79,35 +79,33 @@ def update_logging_config(context, log_name=None, file_name='worker.log'):
             break
 
 
-def get_log_filenames(context):
+def get_log_filename(context):
```
escapewindow: A function to return one constant string might be overkill, but it works :)

PR #61, opened 2017-01-18
start using `unfreeze_values`

escapewindow:
This is a bigger patch than I was expecting, mainly because I didn't
catch some errors in the initial `unfreeze_values` patch.

@johanlorenzo how does this look?  This doesn't block anything, but might be nice to fix before we roll out 2.0.0b1 :)

### DIFF
--- a/scriptworker/config.py
+++ b/scriptworker/config.py
@@ -65,24 +65,27 @@ def unfreeze_values(dictionary):
     A recursive function(top-down conversion)
 
     Args:
-        dictionary(frozendict/tuple): the frozendict/tuple to modify in-place.
+        dictionary(frozendict/tuple): the frozendict/tuple.
+
+    Returns:
+        dictionary (dict/list): the unfrozen copy.
     """
-    if isinstance(dictionary, dict):
+    if isinstance(dictionary, (frozendict, dict)):
+        dictionary = dict(deepcopy(dictionary))
         for key, value in dictionary.items():
             if isinstance(value, tuple):
-                dictionary[key] = list(value)
-                unfreeze_values(dictionary[key])
+                dictionary[key] = unfreeze_values(dictionary[key])
             elif isinstance(value, frozendict):
-                dictionary[key] = value.__dict__['_dict']
-                unfreeze_values(dictionary[key])
-    elif isinstance(dictionary, list):
+                dictionary[key] = unfreeze_values(dictionary[key])
+    elif isinstance(dictionary, (list, tuple)):
+        dictionary = list(dictionary)
         for idx in range(len(dictionary)):
             if isinstance(dictionary[idx], tuple):
-                dictionary[idx] = list(dictionary[idx])
-                unfreeze_values(dictionary[idx])
+                dictionary[idx] = unfreeze_values(dictionary[idx])
             elif isinstance(dictionary[idx], frozendict):
-                dictionary[idx] = dictionary[idx].__dict__['_dict']
-                unfreeze_values(dictionary[idx])
+                dictionary[idx] = unfreeze_values(dictionary[idx])
+
+    return dictionary
 
 
 # read_worker_creds {{{1
--- a/scriptworker/test/__init__.py
+++ b/scriptworker/test/__init__.py
@@ -6,13 +6,13 @@ import aiohttp
 import arrow
 import asyncio
 from copy import deepcopy
-from frozendict import frozendict
 import json
 import mock
 import os
 import pytest
 import tempfile
 import taskcluster.exceptions
+from scriptworker.config import unfreeze_values
 from scriptworker.constants import DEFAULT_CONFIG
 from scriptworker.context import Context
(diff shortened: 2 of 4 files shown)


scriptworker/config.py
```diff
@@ -65,24 +65,27 @@ def unfreeze_values(dictionary):
...
+        dictionary (dict/list): the unfrozen copy.
     """
-    if isinstance(dictionary, dict):
+    if isinstance(dictionary, (frozendict, dict)):
+        dictionary = dict(deepcopy(dictionary))
         for key, value in dictionary.items():
             if isinstance(value, tuple):
-                dictionary[key] = list(value)
-                unfreeze_values(dictionary[key])
+                dictionary[key] = unfreeze_values(dictionary[key])
```
JohanLorenzo: As we now return an unfrozen copy, we can simplify unfreeze_values with list/dict comprehensions:
```py
def unfreeze_values(values):
    """Recursively convert tuples into lists, and frozendicts into dicts.

    Conversion is processed bottom-up.

    Args:
        values (any type): the immutable data structure. Mutable structures are accepted.

    Returns:
        values (dict/list): the unfrozen copy.
    """
    if isinstance(values, (frozendict, dict)):
        return {key: unfreeze_values(value) for key, value in values.items()}

    elif isinstance(values, (list, tuple)):
        return [unfreeze_values(value) for value in values]

    # Nothing to unfreeze.
    return values
```

scriptworker/test/__init__.py
```diff
@@ -261,7 +261,7 @@ def tmpdir2():
 def rw_context():
     with tempfile.TemporaryDirectory() as tmp:
         context = Context()
-        context.config = dict(deepcopy(DEFAULT_CONFIG))
+        context.config = unfreeze_values(DEFAULT_CONFIG)
```
JohanLorenzo: Bikeshed: `unfreeze_values` sounds like an impure function to me. How about something like: `craft_unfrozen_copy()`?

scriptworker/test/test_config.py
```diff
@@ -56,12 +56,23 @@ def test_freeze_values(input_dict, expected_dict):
     assert input_dict == expected_dict
 
 
-@pytest.mark.parametrize("input_dict,expected_dict", (({1: 2}, {1: 2}), ({1: (1, 2)}, {1: [1, 2]}), ({1: frozendict({2: 3})}, {1: {2: 3}}),
-                                                      ({1: (frozendict({2: 3}), (4, 5, 6))}, {1: [{2: 3}, [4, 5, 6]]}),
-                                                      ({1: (1, (2, 3), frozendict({4: 5}))}, {1: [1, [2, 3], {4: 5}]}), ((1, 2), (1, 2))))
+@pytest.mark.parametrize("input_dict,expected_dict", ((
+    {1: 2}, {1: 2}
```
JohanLorenzo: This shouldn't matter anymore with the list/dict comprehension version, but we can make the data types more varied. For instance:
```py
@pytest.mark.parametrize("input_values, expected", ((
    {1: 2}, {1: 2}
), (
    {'a': (1.0, 2.5)}, {'a': [1.0, 2.5]}
), (
    {1: frozendict({'a': True})}, {1: {'a': True}}
), (
    {1: (frozendict({'a': None}), ('A', 'B', 'C'))}, {1: [{'a': None}, ['A', 'B', 'C']]}
), (
    {1: (1, [2, 3], frozendict({}))}, {1: [1, [2, 3], {}]}
), (
    (1, ()), [1, []]
), (
    1, 1
)))
def test_unfreeze_values(input_values, expected):
    assert config.unfreeze_values(input_values) == expected
```

scriptworker/config.py
```diff
@@ -65,24 +65,27 @@ def unfreeze_values(dictionary):
     A recursive function(top-down conversion)
```
JohanLorenzo: I noticed some small bugs in `freeze_values()`:

* Like you did at [line 73](https://github.com/mozilla-releng/scriptworker/pull/61/files#diff-93b3cf2396766ce44e10261ffca70d55R73), we need to check for instances of tuples and frozendicts as well. Otherwise if a tuple contains a list, that list won't be frozen.
* line 55: `freeze_values(idx)` has to become `freeze_values(dictionary[idx])`, right?
* the top level structure is not frozen.

This makes wonder if we shouldn't make `freeze_values()` pure. The code would be a couple of line long, and the top-level gets frozen as well. The only disadvantage I see: each time we call freeze_values() somewhere, the next line has to be `del my_var_which_just_got_copied_by_freeze_values`

escapewindow: I think making both functions pure is a good goal. I do want to avoid a complete rewrite of both functions atm, if possible, since a contributor just wrote these.

JohanLorenzo: Oh you're right. My apologies! The changes look good enough to me, then.

escapewindow:
Merging. I need to open a new issue with the review feedback above.

PR #64, opened 2017-01-23
`get_frozen_copy` and `get_unfrozen_copy`

### LINKED ISSUE #63: revisit `freeze_values` and `unfreeze_values`
#61 has a number of review comments we should revisit, now that 1.0.0b7 is out.

escapewindow:
These replace `freeze_values` and `unfreeze_values`.  Fixes #63 .

@JohanLorenzo this should address your review comments from #61.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Added
+- `scriptworker.artifacts` is a new submodule that defines artifact behavior
+- we now support `pushapk` scriptworker instance types in `cot.verify`
 
 ### Changed
 - `freeze_values` is now `get_frozen_copy`, and now returns a frozen copy instead of modifying the object in place.
@@ -10,6 +13,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `check_config` now calls `get_frozen_copy` on the `config` before comparing against `DEFAULT_CONFIG`
 - `create_config` calls `get_unfrozen_copy`, resulting in a recursively frozen config
 - `DEFAULT_CONFIG` now uses `frozendict`s and `tuple`s in nested config items.
+- `.asc` files are now forced to `text/plain`
+- all `text/plain` artifacts are now gzipped, including .log, .asc, .json, .html, .xml
+- we no longer have `task_output.log` and `task_error.log`.  Instead, we have `live_backing.log`, for more treeherder-friendliness
 
 ## [1.0.0b7] - 2017-01-18
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,21 @@
...
+- we now support `pushapk` scriptworker instance types in `cot.verify`
+
+### Changed
+- `freeze_values` is now `get_frozen_copy`, and now returns a frozen copy instead of modifying the object in place.
+- `unfreeze_values` is now `get_unfrozen_copy`
+- `check_config` now calls `get_frozen_copy` on the `config` before comparing against `DEFAULT_CONFIG`
+- `create_config` calls `get_unfrozen_copy`, resulting in a recursively frozen config
+- `DEFAULT_CONFIG` now uses `frozendict`s and `tuple`s in nested config items.
+- `.asc` files are now forced to `text/plain`
+- `text/plain` artifacts are now gzipped
```
JohanLorenzo: Bikeshed: Not sure it's worth being called out, but json, html and xml files are also gzipped https://github.com/mozilla-releng/scriptworker/pull/53/files#diff-ecbeaeb0c758367cccb8ff577090c03fR25

scriptworker/config.py
```diff
@@ -134,11 +114,12 @@ def check_config(config, path):
     """
     messages = []
 
-    missing_keys = set(DEFAULT_CONFIG.keys()) - set(config.keys())
+    config_copy = get_frozen_copy(config)
```
JohanLorenzo: Nit: Shouldn't we create the frozen copy in `create_config()`? This would enable to (ask to) delete the mutable copy there.

scriptworker/constants.py
```diff
@@ -87,40 +91,40 @@
     "cot_schema_path": os.path.join(os.path.dirname(__file__), "data", "cot_v1_schema.json"),
 
     # for download url validation.  The regexes need to define a 'filepath'.
-    'valid_artifact_rules': ({
-        "schemes": ["https"],
-        "netlocs": ["queue.taskcluster.net"],
-        "path_regexes": ["^/v1/task/(?P<taskId>[^/]+)(/runs/\\d+)?/artifacts/(?P<filepath>.*)$"]
-    }, ),
+    'valid_artifact_rules': (frozendict({
```
JohanLorenzo: It's nice to see immutable structures everywhere 😃

scriptworker/config.py
escapewindow: We check `config` vs `DEFAULT_CONFIG` before we remove `credentials` from `config`, which won't work if `config` is immutable.  We could freeze it, check it, unfreeze it, remove credentials, and refreeze it, but this made more sense.  We could also leave `credentials` in there; I'm not sure if I had a strong reason to remove it other than making sure I could freeze the credentials when `freeze_values` wasn't recursive, other than being able to do a config dump without revealing secrets.

As designed.  We can revisit if there's some other way that makes more sense... which may be leaving the credentials in the config.

scriptworker/constants.py
escapewindow: +1 :)

CHANGELOG.md
escapewindow: Updated the changelog.

PR #66, opened 2017-01-24
Bug 1333394 - Use the new name of docker-images

JohanLorenzo:
It seems like a [renaming happened 2 weeks ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1320328). This causes [issues described here](https://bugzilla.mozilla.org/show_bug.cgi?id=1333394).

@escapewindow , if I'm correct, we're gonna need v1.0.0b8 quickly. I can prepare a v1 branch upon which I can backport this patch.

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -179,7 +179,8 @@ DEFAULT_CONFIG = frozendict({
 
     # docker-image cot
     "valid_docker_image_worker_types": (
-        "taskcluster-images",
+        "taskcluster-images",   # TODO: Remove this image once docker-images is the only valid worker type
+        "gecko-images",
     ),
 
     "valid_docker_image_env_vars": (


scriptworker/constants.py
```diff
@@ -179,7 +179,8 @@
 
     # docker-image cot
     "valid_docker_image_worker_types": (
-        "taskcluster-images",
+        "taskcluster-images",   # TODO: Remove this image once docker-images is the only valid worker type
+        "docker-images",
```
lundjordan: I'm a little confused. iiuc, this should be gecko-images. I need to dive in more to understand.

JohanLorenzo:
Thanks for the second pair of eyes! My searches in the logs confused me. It's gecko-images, indeed.

PR #69, opened 2017-01-30
 Restrict googleplay scopes.

escapewindow:
Also rename the current `release` and `nightly` branch aliases to
`all-release-branches` and `all-nightly-branches`, because they are a
superset of the repos for those releases.  Restricting `release`,
`beta`, and `nightly` to the actual repos used for those release trains
will let us avoid pushing the wrong apk to the wrong product (though the
product name helps avoid that as well).

Finally, avoid using the broken taskcluster 1.0.0 and 1.0.1 packages.

@JohanLorenzo , what do you think?

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -129,9 +129,9 @@ DEFAULT_CONFIG = frozendict({
 
     # scriptworker identification
     "scriptworker_worker_types": (
-        "pushapk-v1",
         "balrogworker-v1",
         "beetmoverworker-v1",
+        "pushapk-v1",
         "signing-linux-v1",
     ),
     "scriptworker_provisioners": (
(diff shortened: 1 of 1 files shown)


scriptworker/constants.py
```diff
@@ -176,14 +176,16 @@
...
-            'project:releng:signing:cert:nightly-signing': 'nightly',
+            'project:releng:balrog:release': 'all-release-branches',
+            'project:releng:beetmover:release': 'all-release-branches',
+            'project:releng:googleplay:release': 'release',
+            'project:releng:signing:cert:release-signing': 'all-release-branches',
+            'project:releng:googleplay:beta': 'beta',
+            'project:releng:googleplay:aurora': 'aurora',
+            'project:releng:balrog:nightly': 'all-nightly-branches',
+            'project:releng:beetmover:nightly': 'all-nightly-branches',
+            'project:releng:pushapk:nightly': 'all-nightly-branches',
```
JohanLorenzo: Should we keep this scope, now that we only rely on `project:releng:googleplay:*`?

escapewindow: Good catch!

PR #72, opened 2017-02-01
fix taskcluster>=1.0.0 support; improve upload logging

escapewindow:
First, `taskcluster>=1.0.0` changes how we make new artifact urls.  I think this is an improvement, but it breaks `get_artifact_url`.  I added support for 1.0.x without breaking 0.3.x for now; we can tear out 0.3.x support once we're on >=1.0.0 for a while and don't foresee needing to go back.

Second, while debugging async uploads, we see lots of this:

```
2017-01-25 19:29:53,340 - scriptworker.utils - DEBUG - retry_async: <function put at 0x7f99bcc7f2f0>: sleeping before retry
2017-01-25 19:29:53,341 - beetmoverscript.script - INFO - 200
2017-01-25 19:29:53,342 - beetmoverscript.script - INFO - 
2017-01-25 19:29:53,349 - beetmoverscript.script - INFO - 200
2017-01-25 19:29:53,350 - beetmoverscript.script - INFO - 
2017-01-25 19:29:53,355 - beetmoverscript.script - INFO - 200
2017-01-25 19:29:53,357 - beetmoverscript.script - INFO - 
2017-01-25 19:29:53,360 - beetmoverscript.script - INFO - 200
2017-01-25 19:29:53,363 - beetmoverscript.script - INFO - 
2017-01-25 19:29:56,922 - beetmoverscript.script - INFO - 400
```

It's not clear which artifacts are succeeding and which are retrying.  By adding the path before the status, we'll be able to debug a bit easier.  We'll need a similar patch for beetmoverscript.

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -189,7 +189,7 @@ async def create_artifact(context, path, target_path, content_type, content_enco
                 tc_response['putUrl'], data=fh, headers=_craft_artifact_put_headers(content_type, content_encoding),
                 skip_auto_headers=skip_auto_headers, compress=False
             ) as resp:
-                log.info(resp.status)
+                log.info("create_artifact {}: {}".format(path, resp.status))
                 response_text = await resp.text()
                 log.info(response_text)
                 if resp.status not in (200, 204):
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -225,14 +225,20 @@ def get_artifact_url(context, task_id, path):
     Raises:
         TaskClusterFailure: on failure.
     """
-    url = urljoin(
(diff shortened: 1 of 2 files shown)


scriptworker/artifacts.py
```diff
@@ -189,7 +189,7 @@ def guess_content_type_and_encoding(path):
                 tc_response['putUrl'], data=fh, headers=_craft_artifact_put_headers(content_type, content_encoding),
                 skip_auto_headers=skip_auto_headers, compress=False
             ) as resp:
-                log.info(resp.status)
+                log.info("create_artifact {}: {}".format(path, resp.status))
```
JohanLorenzo: Sweet!

PR #74, opened 2017-02-07
Bug 1330396 - test git info using archive

### LINKED ISSUE #51: bug 1330396 - add a test git repo tarball into scriptworker/test/data
Currently [two of our tests](https://github.com/mozilla-releng/scriptworker/blob/4d6fb6f67ef8f94676b41755f3c325e30d87cca9/scriptworker/test/test_gpg.py#L593-L605), `test_get_git_revision` and `test_get_latest_tag`, use the current scriptworker git repo for tests.

This is poor for at least two reasons:
1. because we don't know what each person's git repo status will be when they run the tests, we have to run another git command to find the latest revision or the latest tag.  It isn't deterministic.
2. it isn't possible to run the tests without the .git subdir, e.g. from a pip tarball.  The dockerfile currently pushes the .git subdir to the container so we can run unit tests.

Instead, let's

a. create a test git repo, with a few revisions, and a few tags.
b. tar it up and add it to `scriptworker/test/data/`
c. change those two tests to extract that tarball into a tmpdir, and then running the tests against that tmpdir.

we can then have a deterministic latest tag and latest revision.

fuzzmz:
Run `test_get_git_revision` and `test_get_latest_tag` using a pre-made git repository archived in `test_git_repo.tgz`

Extract the archive at the start of each test and delete it once its finished in order to allow users to run the tests individually instead of one depending on the other.

Fixes #51


scriptworker/test/test_gpg.py
```diff
@@ -590,19 +591,24 @@ def test_rebuild_gpg_home_signed(context, trusted_email, tmpdir):
...
 
 # get_git_revision, get_latest_tag {{{1
-# TODO these two tests should use a tarball with a test git repo inside
 @pytest.mark.asyncio
 async def test_get_git_revision():
-    parent_dir = os.path.dirname(__file__)
+    tar = tarfile.open(os.path.join(os.path.dirname(__file__), "data", "test_git_repo.tgz"))
+    tar.extractall(os.path.join(os.path.dirname(__file__), "data"))
+    parent_dir = os.path.join(os.path.dirname(__file__), "data", "testrepo")
     expected = subprocess.check_output(["git", "rev-parse", "HEAD"], cwd=parent_dir)
```
fuzzmz: Thinking a bit about this, do you want to check against a hardcoded SHA (in this case it would be `3e9ed77c`) instead of getting the commit id from git?

On the other, using git itself to grab the sha ensures that the output of `sgpg.get_git_revision` is always the same format as what git returns.

escapewindow: I think a hardcoded sha would be good.

However, the other comment I have is you should use the [`tmpdir` fixture](https://github.com/mozilla-releng/scriptworker/blob/f4475ca/scriptworker/test/__init__.py#L238-L246).

To do this, you would change the function definition to

```
async def test_get_git_revision(tmpdir):
```

pytest will then automatically call the function with the path of a temporary directory that will get cleaned up at the end of the test.  Then you could `tar.extractall` inside of `tmpdir` instead of in `./data`.  You don't have to worry about running `shutil.rmtree` because pytest will clean up `tmpdir` for you after each test run.

scriptworker/test/test_gpg.py
```diff
@@ -590,19 +591,24 @@ def test_rebuild_gpg_home_signed(context, trusted_email, tmpdir):
...
     assert await sgpg.get_git_revision(parent_dir) == expected.decode('utf-8').rstrip()
+    shutil.rmtree(os.path.join(os.path.dirname(__file__), "data", "testrepo"))
 
 
 @pytest.mark.asyncio
 async def test_get_latest_tag():
-    parent_dir = os.path.dirname(__file__)
+    tar = tarfile.open(os.path.join(os.path.dirname(__file__), "data", "test_git_repo.tgz"))
+    tar.extractall(os.path.join(os.path.dirname(__file__), "data"))
```
escapewindow: We can extract to a `tmpdir` here too.

scriptworker/test/test_gpg.py
```diff
@@ -590,19 +591,24 @@ def test_rebuild_gpg_home_signed(context, trusted_email, tmpdir):
...
+    parent_dir = os.path.join(os.path.dirname(__file__), "data", "testrepo")
     expected = subprocess.check_output(["git", "describe", "--abbrev=0"], cwd=parent_dir)
```
escapewindow: If we hardcode the sha in the previous test, it would make sense to hardcode the tag name 'v2' here.

fuzzmz:
Thanks for the feedback @escapewindow . Will update the PR this evening based on it.

fuzzmz:
Updated.

escapewindow:
Thanks @fuzzmz !

PR #78, opened 2017-02-13
allow for retriggered decision tasks

### LINKED ISSUE #77: cot verification: a decision task shouldn't add its decision task to the chain
In [bug 1338933](https://bugzilla.mozilla.org/show_bug.cgi?id=1338933), I retriggered a decision task.

The [original decision task](https://tools.taskcluster.net/task-group-inspector/#/N-cSl1F8Shux4rgxCP4QHg/N-cSl1F8Shux4rgxCP4QHg?_k=6k2aob) had a matching `taskId` and `taskGroupId`, because it was created through cron.yml.

The [retriggered decision task](https://tools.taskcluster.net/task-inspector/#C5_wx697TsydQOxUWH9vbg/) had mismatched `taskId` and `taskGroupId`, because it was retriggered with the `taskGroupId` from the original decision task.

However, if the retriggered decision task is otherwise well-formed and pointing at a valid tree, we shouldn't need to trace it back to the original decision task.

I'm thinking we should put in an exception to the `taskGroupId` rule: if a task is a decision task, don't trace it back to its original decision task.  This assumes that all decision tasks are traceable back to the tree and `mach taskgraph` currently.  If that changes, then we'll have to revisit this rule.

escapewindow:
Only look for tasks inside the task-graph.json if they are not decision
tasks.  (If we generate decision tasks inside the task graph that can
not themselves be traceable back to the tree standalone, we'll need to
revisit this decision.)

Also, allow for more than 2 decision tasks.

Fixes #77 (references https://bugzilla.mozilla.org/show_bug.cgi?id=1338933 )

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -858,7 +858,12 @@ async def verify_decision_task(chain, link):
         path, is_path=True, exception=CoTError, message="Can't load {}! %(exc)s".format(path)
     )
     for target_link in [chain] + chain.links:
-        if target_link.decision_task_id == link.task_id and target_link.task_id != link.task_id:
+        # Verify the target's task is in the decision task's task graph, unless
+        # it's this task or another decision task.
+        # https://github.com/mozilla-releng/scriptworker/issues/77
+        if target_link.decision_task_id == link.task_id and \
+                target_link.task_id != link.task_id and \
+                target_link.task_type != 'decision':
             verify_link_in_task_graph(chain, link, target_link)
     verify_firefox_decision_command(link)
     raise_on_errors(errors)
@@ -980,14 +985,9 @@ def check_num_tasks(chain, task_count):
     # hardcode for now.  If we need a different set of constraints, either
     # go by cot_product settings or by task_count['docker-image'] + 1
     min_decision_tasks = 1
-    max_decision_tasks = 2
     if task_count['decision'] < min_decision_tasks:
         errors.append("{} decision tasks; we must have at least {}!".format(
             task_count['decision'], min_decision_tasks
-        ))
-    elif task_count['decision'] > max_decision_tasks:
-        errors.append("{} decision tasks; we must have at most {}!".format(
-            task_count['decision'], max_decision_tasks
         ))
     raise_on_errors(errors)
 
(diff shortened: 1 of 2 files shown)


escapewindow:
> If I understand correctly, this patch allows to have more than one decision task, which the case of a retrigger.

You do understand correctly.  We also allow for more than one decision task in any other case, and retriggering is one of those cases.

> Do we also allow a completely different decision task to be injected in the graph?

Sure.  This would might be helpful for something like Fennec release promotion: one decision task creates the releasetasks graph, and at the end of that we want to trigger the `mach taskgraph` nightly fennec graph in another decision task.

> If so, does that allow somebody to plug tasks at the end of the taskgraph? Or to insert tasks in the middle?

Good question!  Yes and no.

- By default, `mach taskgraph` will generate a new graph; the decision task will be the only task in the old taskgraph.  Essentially, this means the new decision task has the old decision task's `taskId` as its `taskGroupId`, but all the tasks that the new decision task spawns will have the new decision task's `taskId` as their `taskGroupId`.  That means we verify those new tasks against the new decision task.

- If the new decision task inserts non-decision tasks at the end of the old taskgraph, and those are tasks where we verify the chain of trust artifact, then "at the end of the old taskgraph" means their `taskGroupId` matches the *old* decision task's `taskId`, and we'll verify these tasks exist in the old decision task's `task-graph.json` artifact.  If these are essentially retriggers of previous tasks in the graph, this will work.  If the task definitions have changed at all in a meaningful way, chain of trust verification will fail.

- There is no real way to insert tasks into the "middle" of the old taskgraph.  This essentially means that tasks in the previous graph will rely on artifacts from the new tasks.  We should be explicit about downloading artifacts from upstream `taskId`s, in which case someone would have to replace the definition of that task or the artifacts from that task to MitM the graph.  In the case of overwriting old indexes with new artifacts, that can be done anyway outside the graph; we shouldn't implicitly trust the indexes as they exist today.

escapewindow:
I should mention: we are assuming the decision task is traceable directly to the tree, and it doesn't matter what triggered the decision task.  We have a bunch of checks to verify this currently.  Once we resolve [bug 1328719](https://bugzilla.mozilla.org/show_bug.cgi?id=1328719), we'll be able to directly compare any decision task's definition against a known template, and any divergence will break the chain of trust.

Since decision tasks are assumed and verified to be traceable back to the tree, we don't have to trace and verify what triggered the decision task.

escapewindow:
Thanks!

PR #79, opened 2017-02-15
add helper scripts

### LINKED ISSUE #65: add helper scripts to the repo
- we use a lot of gpg homedirs. It would be great to have something like [gpg.sh](https://github.com/escapewindow/docker-signing-server/blob/master/gpg.sh) packaged up in the scriptworker package (in the `data/` subdir?) for debugging purposes.

- we also need to create gpg keypairs on a regular basis.  It would be great to have something like [gpg.py](https://bug1298553.bmoattachments.org/attachment.cgi?id=8792961) packaged up in the scriptworker package.

Both scripts would need to be more generic, but those are good templates to start from.

fuzzmz:
This is still a work in progress and shouldn't be merged currently.

Add gpg_helper.sh script which allows running of gpg commands with a specific gpg binary and over a particular gpg homedir, declared either as environment variables or as command line arguments.

The issue with using `getopts` to enable the use of named arguments (`-b` for the binary and `-h` for the home folder) is that it requires to separate the rest of the commands with a double dash (`--`).

This way, for example, invoking the script would look like `./gpg_helper.sh -b /usr/local/bin/gpg2 -h ~/.gnupg -- --list-all`

Another option would be to ditch the use of getopts and just rely on the position of the arguments (the first argument would be the binary, the second the home folder, the rest would be gpg commands), which would make the script invocation be `./gpg_helper.sh /usr/local/bin/gpg2 ~/.gnupg --list-all` but which has the downside of breaking if the users doesn't use the arguments in the correct order.

A third option would be to write a custom argument parser, but for such a small script I'm not sure if it's worth the effort.

Any preferred way to proceed with the shell script?

Fixes #65

### DIFF
--- a/helper_scripts/create_gpg_keys.py
+++ b/helper_scripts/create_gpg_keys.py
@@ -1,16 +1,23 @@
 #!/usr/bin/env python
+"""Create gpg keys for scriptworker instances."""
 
 import argparse
 import gnupg
+import logging
 import os
 import shutil
 import scriptworker.gpg
 
 
+log = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+
+
 class FullPaths(argparse.Action):
-    """Expand user- and relative-paths"""
+    """Expand user- and relative-paths."""
 
     def __call__(self, parser, namespace, values, option_string=None):
+        """Override the Action __call__ method."""
         setattr(namespace, self.dest, os.path.abspath(os.path.expanduser(values)))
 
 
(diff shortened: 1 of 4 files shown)


escapewindow:
Hm, I think getopts, falling back to defaults ; I think adding a -- is ok.  Users can alias this if it gets unwieldy.

Am I reading it correctly that it'll look at `$GPGBIN` in the env if not specified?  I think setting GPGBIN to `gpg` if not specified in the env or commandline might be nice, but not required.

Glad to see progress on this, thank you!

fuzzmz:
Yes, for both the GPG binary and home folder the script will fall back to environment variables if not defined on the command line. Will add a check to default to `gpg` if it's not defined on the command line or in the env.

escapewindow:
@fuzzmz are you still working on this?  No rush if you are.  If you're intending on stopping here, maybe I should review `gpg_helper.sh` and we can leave the keygen script for another PR.  Let me know.

fuzzmz:
Should have this done tomorrow evening in the worst case. Sorry for the
delay, work's been a bit hectic.

-- Serban Constantin

escapewindow:
No worry, and no rush. Thanks for the update!

helper_scripts/create_gpg_keys.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class FullPaths(argparse.Action):
+    """Expand user- and relative-paths"""
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, self.dest, os.path.abspath(os.path.expanduser(values)))
+
+parser = argparse.ArgumentParser(description="GPG keypair creation and export script")
+parser.add_argument('-b', '--gpgbin', nargs='?', help="GPG binary used", action=FullPaths, default="/usr/local/bin/gpg2")
```
fuzzmz: What I wanted to know regarding the defaults for the arguments is if we should use the environment variables set by `gpg_helper.sh` instead of the hardcoded defaults already in place.

fuzzmz:
Not exactly sure why flake8 keeps failing with those errors as they seem really strange to me.

escapewindow:
Thanks!
I wanted to get to this today, but will have to put this off at least another day.

I'm guessing flake8-docstrings is being its usual overly-strict self.  We can probably add `helper_scripts` to the [flake8 exclude](https://github.com/fuzzmz/scriptworker/blob/e96a59c125df6fb9c016e6ee83fd864f98be41a6/tox.ini#L42) and it should green up.

PR #82, opened 2017-02-22
Bug 1338186 - add beetmover/balrog release scopes. r=aki

MihaiTabara:
Since we didn't get to ship any release with current {beetmover,balrog}script workers, I think it's safe to overwrite the release-related ones with the new scopes format. For the nightly scopes, we keep the old scopes and add the new scope format. We're likely to delete the former once testing is done and we're backward compatible and green to go.

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -176,14 +176,17 @@ DEFAULT_CONFIG = frozendict({
     # Map scopes to restricted-level
     'cot_restricted_scopes': frozendict({
         'firefox': frozendict({
-            'project:releng:balrog:server:release': 'all-release-branches',
+            'project:releng:balrog:server:release': 'release',
+            'project:releng:balrog:server:beta': 'beta',
+            'project:releng:balrog:server:aurora': 'aurora',
+            'project:releng:balrog:server:esr': 'esr',
             'project:releng:beetmover:bucket:release': 'all-release-branches',
             'project:releng:googleplay:release': 'release',
             'project:releng:signing:cert:release-signing': 'all-release-branches',
             'project:releng:googleplay:beta': 'beta',
             'project:releng:googleplay:aurora': 'aurora',
             'project:releng:balrog:nightly': 'all-nightly-branches',
-            'project:releng:balrog:server:nightly': 'all-nightly-branches',
+            'project:releng:balrog:server:nightly': 'nightly',
             'project:releng:beetmover:nightly': 'all-nightly-branches',
             'project:releng:beetmover:bucket:nightly': 'all-nightly-branches',
             'project:releng:signing:cert:nightly-signing': 'all-nightly-branches',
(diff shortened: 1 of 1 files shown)


scriptworker/constants.py
```diff
@@ -176,14 +176,16 @@
     # Map scopes to restricted-level
     'cot_restricted_scopes': frozendict({
         'firefox': frozendict({
-            'project:releng:balrog:release': 'all-release-branches',
-            'project:releng:beetmover:release': 'all-release-branches',
+            'project:releng:balrog:server:release': 'all-release-branches',
```
escapewindow: This looks good, but:  Do you want to separate nightly/aurora/beta/release/esr on balrog?

We will probably end up using the same creds / server config per scope, unless @mozbhearsum decides we can split the credentials per channel type.  But we can allowlist the various channels each branch can push to: release/releasetest/cdntest etc for release, esr/esrtest/whatever for esr, etc.  In which case, `project:releng:balrog:server:release` would point at `release` instead of `all-release-branches`.

MihaiTabara: Hm, good point. Let me see if I get this right:

`project:releng:balrog:server:release` => `release`,
`project:releng:balrog:server:esr` => `esr`, (I could add a separate dict containing esr45 + esr52)
`project:releng:balrog:server:beta` => `beta`,
`project:releng:balrog:server:release` => `nightly` => `all-nightly-branches`

?

escapewindow: > project:releng:balrog:server:release => release,
> project:releng:balrog:server:esr => esr, (I could add a separate dict containing esr45 + esr52)
> project:releng:balrog:server:beta => beta,

Yes.  And yes, we'll need an `esr` in `cot_restricted_trees`.

> project:releng:balrog:server:release => nightly => all-nightly-branches

I think you mean `project:releng:balrog:server:nightly`, and I think it should point at a new `nightly`, which should just be m-c.

MihaiTabara: Sounds good. Mea culpa for second statement, I copy-pasted and forgot to update.
👍 will follow-up with patch here.
Thanks!

escapewindow: Ok.  For reference, I added the new balrog/beetmover scopes in [this patch](https://reviewboard.mozilla.org/r/113722/diff/2#index_header).

MihaiTabara: I've pushed the refactored patch. Two observations here:
1. In your in-tree [patch](https://reviewboard.mozilla.org/r/113722/diff/2#index_header) shouldn't you have `esr` entry in `BALROG_SERVER_SCOPES` and `BALROG_SCOPE_ALIAS_TO_PROJECT` ?

2. I suppose `aurora` is in the `release` section because of possible staging-beta out of it, but for some reason in my head something is not right with respect to future aurora nightlies once we land this.

escapewindow: * I haven't added esr yet.  We can when we uplift to esr52, after we've proven this model.

* You mean aurora is in `all-release-branches`? Yeah, we can remove it if we're planning on doing all release branches on jamun or something.

PR #87, opened 2017-03-06
cot.verify: don't look at *:decision:decision

escapewindow:
In [bug 1343649](https://bugzilla.mozilla.org/show_bug.cgi?id=1343649), we're launching a decision task from a non-cot-enabled releasetasks graph.  Without this patch, we follow the decision task's `taskGroupId` to the non-cot-enabled decision task, and die.  With this patch, we ignore the releasetasks graph outside of the decision task, and pass.


escapewindow:
Tested via `verify_cot --task-type signing WuA-l_GyQWqfmHlKQTSuTg`

PR #95, opened 2017-03-22
Add get_upstream_artifacts()

JohanLorenzo:
PushAPK needs this list, and beetmover [already implements](https://github.com/mozilla-releng/beetmoverscript/blob/d6d69ef5ee2789569b076f46557f8f9d36b5ffda/beetmoverscript/task.py#L106) the same bits. 

I'm not too sure sorting by taskId is the best data structure. PushAPK doesn't need it, and beetmover sorts by locale.

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -14,7 +14,7 @@ import os
 from urllib.parse import unquote, urljoin
 
 from scriptworker.client import validate_artifact_url
-from scriptworker.exceptions import ScriptWorkerRetryException
+from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 from scriptworker.task import get_task_id, get_run_id, get_decision_task_id
 from scriptworker.utils import download_file, filepaths_in_dir, raise_future_exceptions, retry_async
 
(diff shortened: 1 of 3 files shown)


JohanLorenzo:
@escapewindow Could we release a new version once this PR is merged? That'll help solving https://github.com/mozilla-releng/pushapkscript/pull/15#issuecomment-288484827

scriptworker/artifacts.py
```diff
@@ -306,3 +306,39 @@ def get_expiration_arrow(context):
...
+    return {
+        task_id: [
+            _get_single_upstream_artifact(context, task_id, path)
+            for path in paths
+        ]
+        for task_id, paths in task_ids_and_relative_paths
+    }
+
+
+def _get_single_upstream_artifact(context, task_id, path):
```
escapewindow: `_get_single_upstream_artifact_full_path` ?

Per our conversation in IRC:

- We also generate these full paths in `scriptworker.cot.verify`.  It would be good to reuse this function there, so we only have one place that defines these paths.
    - that may mean we drop the leading underscore
    - we probably want to avoid raising `ScriptWorkerTaskException` here, since the artifact won't exist on disk before we download.  We can either have a kwarg that says we should run this check, or run that check in `get_upstream_artifacts`.

escapewindow: `link.cot_dir` could even be a `functools.partial` that pre-fills-out `context` and `task_id` for `get_single_upstream_artifact_full_path`.  `link.cot_path('task.json')` would then return `$work_dir/cot/$task_id/task.json`.  (This isn't required; just something I thought of.)

I did notice we use `link.cot_dir` to define the `parent_dir` to download the artifacts into: https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/cot/verify.py#L535  Maybe we allow for `path` to be `None` in `get_single_upstream_artifact_full_path` ?  (Or `get_cot_full_path` or something if that gets unwieldy?)

scriptworker/artifacts.py
```diff
@@ -306,3 +306,39 @@ def get_expiration_arrow(context):
 
     await raise_future_exceptions(tasks)
     return files
+
+
+def get_upstream_artifacts(context):
```
JohanLorenzo: I agree. How about `get_upstream_artifacts_full_paths_per_task_id`?

scriptworker/artifacts.py
JohanLorenzo: I ended up not usgin with functools. Instead I used a regular method, and cot_dir is a property.

PR #100, opened 2017-03-25
fix decision image bustage

escapewindow:


scriptworker/constants.py
```diff
@@ -103,6 +103,7 @@
             "sha256:31035ed23eba3ede02b988be39027668d965b9fc45b74b932b2338a4e7936cf9",
             "sha256:7320c720c770e9f93df26f7da742db72b334b7ded77539fb240fc4a28363de5a",
             "sha256:9db282317340838f0015335d74ed56c4ee0dbad588be33e6999928a181548587",
+            "sha256:a22b90c7e16191a701760ef4f9159e86289ba598bf8ff5b22b7b94867530460d",
```
JohanLorenzo: @escapewindow Just to double check, should we have a different revision here, compared to https://reviewboard.mozilla.org/r/123212/diff/1#index_header and https://bug1350045.bmoattachments.org/attachment.cgi?id=8851262 ?

escapewindow:
Sadly, these are two different hashes, as described by https://bugzilla.mozilla.org/show_bug.cgi?id=1326436#c7 .

Right now (aiui), the in-tree hash is a content-addressable hash, and the hash that docker-worker verifies is a docker image ID, and they differ.  I don't know if or how you can determine one from the other.  Scriptworker is able to verify the latter against a manually maintained allowlist.

In the future, for cot v2, it looks like we'll verify that we're requesting the same content-addressable hash as exists in-tree, and assume that docker has downloaded the right image.  That still bothers me a bit, but that seems to be what https://github.com/docker/docker/issues/18133 is saying.  https://bugzilla.mozilla.org/show_bug.cgi?id=1326436#c10

escapewindow:
Thanks!

PR #104, opened 2017-04-05
Bug 1353642 - adjust nightly scopes in the new format for balrog. r=aki

MihaiTabara:


MihaiTabara:
@escapewindow re: overwriting nightly channel with date bits, at the proper moment, I fully agree. 
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1353949 to track this in the future.

PR #105, opened 2017-04-05
3.1.0

escapewindow:


escapewindow:
This will give us a 3.1.0 release for the new balrog scope perms for date.  @MihaiTabara , note that 3.0.0 bumped aiohttp to >=2.0.0; we need >=2.0.4 for logs to upload properly in text/plain.  I used dephash to find all the new dependencies, and `pip -r <reqfile> download` to get all the new modules to upload to puppet.

scriptworker/version.py
```diff
@@ -50,7 +50,7 @@ def get_version_string(version):
 
 # 1}}}
 # Semantic versioning 2.0.0  http://semver.org/
-__version__ = (3, 0, 0)
+__version__ = (3, 1, 0)
```
lundjordan: based on list of changes and chosen versioning semantics, I suppose to me the change could be either a second or third digit bump: e.g. 3.0.1 potentially. I guess it depends on how much of a change something like "allowlists updates" are

escapewindow: +1.  I flipped a coin and decided to define "bug fixes" as "We released something that behaves some way other than expected, and that behavior breaks something", rather than "It behaves as expected, but we want to change that behavior".

But semver.org says

> Bug fixes not affecting the API increment the patch version, backwards compatible API additions/changes increment the minor version, and backwards incompatible API changes increment the major version.

This doesn't change the api at all, so it looks like I should have gone with 3.0.1.  However, I pushed it to pypi, so... next time =\

PR #106, opened 2017-04-06
use claimWork

### LINKED ISSUE #56: bug 1331098 - use claimWork instead of hitting azure
Currently, we ask Taskcluster what the polling urls are, then poll Azure for the task hints, then `claimTask` the appropriate task(s)..

At some point, [claimWork](https://docs.taskcluster.net/reference/platform/queue/references/api#claimWork) is going to become the standard way of claiming tasks from the Taskcluster queue.  No worker currently uses this, but the TC team will test a busy docker-worker workerType against this soon.

A caveat here: we're considering supporting multiple workerTypes in scriptworker, to help autophone (#44).  If we use the claimWork endpoint, we may have to order the workerTypes (workerType B doesn't claim any work until all workerType A tasks are claimed) or randomize, or something.  If we have to support both workerTypes on an equal footing, we may have to stay on Azure.

escapewindow:
Fixes #56.


escapewindow:
I think this will bump us to 4.0.0 :)

scriptworker/worker.py
```diff
@@ -38,43 +37,38 @@
...
-    )
-    for poll_url, delete_url in get_azure_urls(context):
+    try:
+        tasks = await claim_work(context)
+    except ScriptWorkerException:
+        await asyncio.sleep(context.config['poll_interval'])
+        return
+    if tasks and tasks.get('tasks', []):
+        log.info(tasks)
+        context.claim_task = tasks['tasks'][0]  # scriptworker can only handle 1 concurrent task
```
escapewindow: We really shouldn't be logging the task, which includes temp cred info ... so thank you for reminding me to take my debug logging out :D

escapewindow: Ok.  New patch pushed -- the interdiff will be large, but I think the diff against master will be smaller.

escapewindow:
This could use a review pass on the new changes.

I'm also considering holding off landing until we're sure we want to bump to 4.0.0; it might be nice to make these types of large changes all at once.  I'd like to wait until the async fixes to taskcluster-client.py, at least.

escapewindow:
Holding off on merging until https://github.com/taskcluster/taskcluster-client.py/pull/67 lands and is in a new tc-client.py release.

escapewindow:
Integration tests are still busted, but don't show on Travis.  Let's fix those before merging.

escapewindow:
Integration tests are passing with `taskcluster==1.3.1`

PR #111, opened 2017-04-14
allow for a subset of task dependencies in cot fuzzy matching.

escapewindow:
Currently, SOP for release promotion is to cancel breakpoint tasks when
we're ready to push live, and then the downstream tasks can trigger.

Pushapk is scheduled as part of the relpro graph 1, with a breakpoint
task dependency.  However, we currently have to schedule a graph 2 when
we pass the expiry point; this is problematic, as graph 2 isn't
currently CoT enabled.

Fuzzy matching allows us to retrigger a task, and it allows for changed
timestamps and taskIds.  However, until now, the dependency list
had to stay the same.  This was problematic for any task with an expired
breakpoint task dependency.

This change allows us to retrigger the pushapk (or any other) task, with
updated timestamps and a *subset* of dependencies, allowing us to bypass
the need for a CoT-enabled graph 2.  We still want that, but in its own
timeframe.


scriptworker/cot/verify.py
```diff
@@ -710,7 +710,7 @@ def verify_task_in_task_graph(task_link, graph_defn, level=logging.CRITICAL):
     errors = []
     runtime_defn = task_link.task
     # dependencies
-    bad_deps = set(graph_defn['task']['dependencies']).symmetric_difference(set(runtime_defn['dependencies']))
+    bad_deps = set(runtime_defn['dependencies']) - set(graph_defn['task']['dependencies'])
```
JohanLorenzo: Nit: I'd add comment to explain why we allow some tasks to be missing. This may help our future selves.

PR #112, opened 2017-05-23
Clean up google play scopes and switch Google Play Aurora to nightly.

mozbhearsum:
@JohanLorenzo and I talked about this a bit on IRC and decided we could clean up the other channels, too. I don't think there's any reason we need to have "date" in the scopes for Aurora, since we wouldn't want those builds pushed to the Play Store (the branding is wrong).

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -190,6 +190,11 @@ DEFAULT_CONFIG = frozendict({
             'project:releng:googleplay:release': 'release',
             'project:releng:signing:cert:release-signing': 'all-release-branches',
             'project:releng:googleplay:beta': 'beta',
+            # As part of the Dawn project we decided to use the Aurora Google Play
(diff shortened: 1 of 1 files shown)


scriptworker/constants.py
```diff
@@ -189,8 +189,8 @@
             'project:releng:beetmover:bucket:release': 'all-release-branches',
             'project:releng:googleplay:release': 'release',
             'project:releng:signing:cert:release-signing': 'all-release-branches',
-            'project:releng:googleplay:beta': 'betatest',
-            'project:releng:googleplay:aurora': 'auroratest',
+            'project:releng:googleplay:beta': 'beta',
+            'project:releng:googleplay:aurora': 'nightly',
```
JohanLorenzo: Nit: I'd add a comment to explain this discrepancy. This might confuse our future selves.

mozbhearsum: Good idea, I'll add one.

PR #118, opened 2017-06-13
CoT verify: Support both types of artifacts in task payloads

JohanLorenzo:
~~WIP~~

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -780,7 +780,7 @@ def _take_expires_out_from_artifacts_in_payload(payload):
 
     artifacts_iterable = artifacts.values() if isinstance(artifacts, dict) else artifacts
     for artifact_definition in artifacts_iterable:
-        if isinstance(artifact_definition, dict) and 'expires' in value:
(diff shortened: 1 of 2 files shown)


scriptworker/cot/verify.py
```diff
@@ -754,9 +754,7 @@ def verify_task_in_task_graph(task_link, graph_defn, level=logging.CRITICAL):
     # payload - eliminate the 'expires' key from artifacts because the datestring
     # will change
     for payload in (runtime_defn['payload'], graph_defn['task']['payload']):
-        for key, value in payload.get('artifacts', {}).items():
-            if isinstance(value, dict) and 'expires' in value:
-                del(value['expires'])
+        payload = _take_expires_out_from_artifacts_in_payload(payload)
```
escapewindow: You're returning a deepcopy of `runtime_defn['payload']`, and then we're testing the unchanged `runtime_defn`.
If we want to deepcopy, we should probably do that in line 742:

```python
    runtime_defn = deepcopy(task_link.task)
```

JohanLorenzo: Good point. Done

PR #119, opened 2017-06-14
Updated integration_credentials to credentials in README

BrandonTang:
Noticed this problem while getting started with scriptworker.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ With integration tests, first create a client with the ``assume:project:taskclus
 Then  create a ``./secrets.json`` or ``~/.scriptworker`` that looks like::
 
     {
-        "integration_credentials": {
+        "credentials": {
             "clientId": "...",
             "accessToken": "...",
             "certificate": "..."


escapewindow:
The integration credentials are specifically needed for running integration tests through tox. This section is for tox testing, so I believe this is correct; are you hitting errors running tox?

If you're not referring to tox, it's in `scriptworker.yaml.tmpl` [here](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker.yaml.tmpl#L23-L26). The readme might need clarification about that, though.

escapewindow:
We likely want to point people at [ReadTheDocs](http://scriptworker.readthedocs.io/en/latest/index.html) for anything more in depth, and specifically [Adding new scriptworker instances of an existing type](http://scriptworker.readthedocs.io/en/latest/new_instance.html) ([source](https://github.com/mozilla-releng/scriptworker/blob/master/docs/new_instance_types.md)).  That page can probably use some improvement :)

escapewindow:
@BrandonTang do you have thoughts here? I'm leaning towards closing this PR, but I'm happy to make doc clarifications if the above isn't clear.

BrandonTang:
You can go ahead and close the PR, I realize now that the integration credentials are specifically for running integration tests through tox. What you stated is clear enough so I don't believe doc clarifications are necessary, thanks!

PR #120, opened 2017-06-15
4.1.2

escapewindow:
I'm tagging+releasing 4.1.2 off of cb93e51 , so we should perform a merge commit rather than squash&merge.


escapewindow:
Thanks!

PR #123, opened 2017-06-27
verify cot.taskId matches link.task_id

escapewindow:
We specify the `taskId` in the chain of trust artifact to help us verify
it matches the task we downloaded it from. We seem to be missing a test
to make sure they match. This patch adds that test.


escapewindow:
No... ideally we haven't hit this yet. I'm not quite sure if this is a critical check or not, but it seems good to have. Thanks!

PR #125, opened 2017-06-29
Cleanup after Dawn and EOL for esr45.r=aki

MihaiTabara:


MihaiTabara:
I'll have a look into what's breaking later on today and will follow-up with a fix.

MihaiTabara:
Thanks for fixing this, it slipped my todo list in triaging at some point!

PR #129, opened 2017-07-12
Bug 1378744 - allow dependencies on decision tasks

djmitche:


djmitche:
No, I think this feeds back into a complaint you made about retriggering decision tasks, so that's the only rush :)

escapewindow:
Haha :) Thanks for working on that!

PR #130, opened 2017-07-12
4.1.3

escapewindow:
Mainly to resolve #129, but we have a few changes queued up and it's been nearly a month since 4.1.2.


setup.py
```diff
@@ -1,9 +1,21 @@
+from __future__ import print_function
 import json
 import os
 from setuptools import setup
 from setuptools.command.test import test as TestCommand
 import sys
 
+if {'register', 'upload'}.intersection(set(sys.argv)):
```
JohanLorenzo: 👍

Issue #132, opened 2017-08-11
stop rotating logs in scriptworker


escapewindow:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1389512 , kmoir added log rotation via logrotate.d, which allows us to compress the logfiles.

Let's support that log rotation on all scriptworker instance types in puppet, and strip out log rotation support from scriptworker.

escapewindow:
Fixed in #142 .

PR #133, opened 2017-08-11
stop rotating logs in scriptworker #132

kmoir:


scriptworker/log.py
```diff
@@ -51,13 +51,6 @@ def update_logging_config(context, log_name=None, file_name='worker.log'):
 
     # Rotating log file
     makedirs(context.config['log_dir'])
-    path = os.path.join(context.config['log_dir'], file_name)
-    handler = logging.handlers.RotatingFileHandler(
-        path, maxBytes=context.config['log_max_bytes'],
-        backupCount=context.config['log_num_backups'],
-    )
-    handler.setFormatter(formatter)
-    top_level_logger.addHandler(handler)
```
escapewindow: This change will stop all scriptworker logging to logfiles; we'll only get log output to the console.
I'm pretty sure that's not intended :)

We would want to make sure we had logrotate.d rotation with all scriptworker instances first. We're missing signing, pushapk, and transparency atm.

We might want to switch to [WatchedFileHandler](https://docs.python.org/2/library/logging.handlers.html#watchedfilehandler) since it references `logrotate`.

With that change, we probably want to remove [`log_max_bites` and `log_num_backups`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/constants.py#L39-L40) from `scriptworker/constants.py`. If those were specified in puppet, we'd need to remove them [here](https://hg.mozilla.org/build/puppet/file/tip/modules/scriptworker/templates/scriptworker.yaml.erb) (but they're not, so we're good there).

kmoir:
sorry I was going to fix this before I left for the day but then was investigating another issue, I meant to cancel the PR. Thanks for the suggestions.

escapewindow:
I tried adding watched file handler support, based on your patch: https://github.com/mozilla-releng/scriptworker/compare/master...escapewindow:watchedfilehandler

I'm hoping that WatchedFileHandler will notice when logrotate.d rotates the log file, and we won't have more open filehandle issues.

escapewindow:
@kmoir how would you like to proceed? If you want to keep working on this, I can review your patches and we can land, or I can keep working on this. I'm happy either way.

escapewindow:
We landed these changes, rebased, in #142 . Thanks Kim!

PR #134, opened 2017-08-14
bug 1389978 - close the contextual log file handle

escapewindow:
In [bug 1389978](https://bugzilla.mozilla.org/show_bug.cgi?id=1389978), Nick found we were keeping tons of old log filehandles open, which ate up a ton of disk. Let's not do that.

### DIFF
--- a/scriptworker/log.py
+++ b/scriptworker/log.py
@@ -143,4 +143,5 @@ def contextual_log_handler(context, path, log_obj=None, level=logging.DEBUG,
     contextual_handler.setFormatter(formatter)
     log_obj.addHandler(contextual_handler)
     yield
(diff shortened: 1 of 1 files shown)


nthomas-mozilla:
Is it something to do with the async nature of these functions or yielding that means this doesn't work as is ? I'm confused why the call to verify_chain_of_trust() doesn't complete, destroying the file handle generated by contextual_log_handler(), well before we get to the end of job cleanup.

nthomas-mozilla:
Also might be slightly preferable to remove the handler before closing the file.

escapewindow:
[python does not release filehandles to logfile](https://stackoverflow.com/a/15474586)

PR #136, opened 2017-08-16
bug 1389850 - add new allowed decision/image workerTypes

djmitche:


scriptworker/constants.py
```diff
@@ -149,13 +149,19 @@
 
     # decision task cot
     "valid_decision_worker_types": (
-        "gecko-decision",
+        "gecko-1-decision",
+        "gecko-2-decision",
+        "gecko-3-decision",
+        "gecko-decision", # legacy
```
escapewindow: Linting requires 2 spaces before an inline comment. Also, is there a timeframe or bug to follow for when `gecko-decision` can be removed?

scriptworker/constants.py
```diff
@@ -149,13 +149,19 @@
...
     ),
 
     # docker-image cot
     "valid_docker_image_worker_types": (
         "taskcluster-images",   # TODO: Remove this image once docker-images is the only valid worker type
-        "gecko-images",
+        "gecko-images", # legacy
```
escapewindow: Same here.
Also, for the above `TODO`, do you know if there's a timeframe or bug to follow for when `taskcluster-images` can be removed?

djmitche:
Lint fixed.

I don't know when the others can be removed -- I'd defer to you and the release team for that.  Once all release branches aren't using those workerTypes, they can be removed here and from the list of workerTypes in the provisioner.

escapewindow:
Thank you!

PR #150, opened 2017-08-31
bug 1393277 - support verifying action tasks; support verifying any valid task type in `verify_cot`

escapewindow:
Hey @JohanLorenzo ,

Action tasks are special types of decision tasks. The first commit in this PR adds support for action task `task.payload.command`s, so we can support verifying retriggers, backfills, and possibly special release graph actions.

The second commit adds support in the `verify_cot` commandline tool, so we can now verify any valid task type... previously we could only verify scriptworker task types. So we can

```
# I debated adding an `action` task type; maybe I will, as an alias to `decision` ?
verify_cot --task-type decision --cleanup -- FS8XnGK4STiSzPz6DDhmgw
verify_cot --task-type build --cleanup -- LwmS1TaMQUC_zYiE3lU3JA
verify_cot --task-type signing --cleanup -- OzdLQ7QjTNyqOHFj3BzfSw
```

for the tasks in https://tools.taskcluster.net/groups/FS8XnGK4STiSzPz6DDhmgw . I found this very helpful in trying to test the action task verification.

This might not be my only PR in support of [bug 1393277](https://bugzilla.mozilla.org/show_bug.cgi?id=1393277), but it should unblock a lot of things.


scriptworker/cot/verify.py
```diff
@@ -31,6 +34,18 @@
...
 
 
+DECISION_MACH_COMMANDS = ((
+    './mach', 'taskgraph', 'decision'
+), (
+    './mach', 'taskgraph', 'add-tasks'
+), (
+    './mach', 'taskgraph', 'backfill'
+), (
+    './mach', 'taskgraph', 'action-callback'
```
JohanLorenzo: Question: I see what the other commands are for. What is this one?

escapewindow: I think it's the new way of doing things. https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/actions.html#creating-a-callback-action

PR #152, opened 2017-08-31
Bug 1394883 - add new decision and docker-image hashes

djmitche:


escapewindow:
Puppet needs a patch as well.

PR #153, opened 2017-08-31
Bug 1395680: allow /home/worker or /builds/worker

djmitche:
These paths recently changed as part of
https://bugzilla.mozilla.org/show_bug.cgi?id=1338651#c195 so CoT needs
to support both while the change rides the trains.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -922,7 +922,7 @@ def verify_firefox_decision_command(decision_link):
     allowed_args = ('--', 'bash', '/bin/bash', '-cx')
     # /home/worker is an old path, before https://bugzilla.mozilla.org/show_bug.cgi?id=1338651#c195
     if command[0] not in('/home/worker/bin/run-task', '/builds/worker/bin/run-task'):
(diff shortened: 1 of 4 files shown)


scriptworker/cot/verify.py
```diff
@@ -920,7 +920,8 @@ def verify_firefox_decision_command(decision_link):
     errors = []
     command = decision_link.task['payload']['command']
     allowed_args = ('--', 'bash', '/bin/bash', '-cx')
-    if command[0] != '/home/worker/bin/run-task':
+    # /home/worker is an old path, before https://bugzilla.mozilla.org/show_bug.cgi?id=1338651#c195
+    if command[0] != '/home/worker/bin/run-task' and command[0] != '/builds/worker/bin/run-task':
```
escapewindow: Could be `command[0] not in (...):`, but it's equivalent.

scriptworker/cot/verify.py
```diff
@@ -920,7 +920,8 @@ def verify_firefox_decision_command(decision_link):
...
         errors.append("{} {} command must start with /home/worker/bin/run-task!".format(
```
escapewindow: We probably want to update this error message.

PR #157, opened 2017-09-04
bug 1396517 - don't add a decision task's decision task to cot

escapewindow:
The current assumption is decision tasks should be standalone and
verifiable as-is. This is fallout from
https://github.com/mozilla-releng/scriptworker/commit/0356bbb4b8e73c9af4d8a2ca7eec4e88cc0b97a6#diff-340a1f42e56ecb68f523d2badfa58686L513 .

I was able to run `verify_cot` against the failing fennec signing task
and its decision task; they now pass.

### DIFF
--- a/scriptworker/test/test_cot_verify.py
+++ b/scriptworker/test/test_cot_verify.py
@@ -393,8 +393,10 @@ def test_verify_docker_image_sha_bad_allowlist(chain, build_link, decision_link,
         cotverify.verify_docker_image_sha(chain, decision_link)
 
 
+# find_sorted_task_dependencies{{{1
 @pytest.mark.parametrize("task,expected,task_type", ((
-    {'taskGroupId': 'task_id', 'extra': {}, 'payload': {}},
+    # Make sure we don't follow other_task_id on a decision task
+    {'taskGroupId': 'other_task_id', 'extra': {}, 'payload': {}},
     [('decision:decision', 'task_id')],
     'decision'
 ), (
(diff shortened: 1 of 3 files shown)


escapewindow:
```
verify_cot --task-type signing --cleanup -- ZXdZAznCSh-55ZenYWYIYQ
verify_cot --task-type decision --cleanup -- dkTyxu_OT6auOGjYNGpS_g
```

scriptworker/test/test_cot_verify.py
```diff
@@ -393,8 +393,10 @@ def test_verify_docker_image_sha_bad_allowlist(chain, build_link, decision_link,
         cotverify.verify_docker_image_sha(chain, decision_link)
 
 
+# find_sorted_task_dependencies{{{1
 @pytest.mark.parametrize("task,expected,task_type", ((
-    {'taskGroupId': 'task_id', 'extra': {}, 'payload': {}},
+    # Make sure we don't follow other_task_id on a decision task
+    {'taskGroupId': 'other_task_id', 'extra': {}, 'payload': {}},
```
JohanLorenzo: I'm not sure to see how this data point tests the change. More precisely, I don't understand why it didn't fail prior to this change.

I see the expected payload is "decision_task_id" which was different from just "task_id". How did that pass?

escapewindow: By default, we add the `taskGroupId` to the list of dependencies, to make sure we add this task's decision task. The task name is `decision`, so it's a decision task; that should trigger the new `if` block... we're going to add this task's `taskId` to the chain, rather than its decision task's (so, add `task_id` rather than `other_task_id` in this case).

Previously, the `taskGroupId` and `taskId` of the decision task were both `task_id`, so it wouldn't have caught breakage of the new logic.

Make sense?

JohanLorenzo: Yes, it does!

PR #158, opened 2017-09-23
bug 1393277 - cot-verifiable action/parent tasks.

escapewindow:
Action tasks are now full fledged graph- or subgraph- generators. This task's `task.extra.parent` points at the decision- or action- task that generated this task, so when we verify our task definition, we should look at the parent instead of the decision task. Action tasks should still check their decision task's validity; in the future we'll rebuild their task definitions from their context and the decision task's `actions.json`.

This should fix our retrigger verifications once [bug 1370343](https://bugzilla.mozilla.org/show_bug.cgi?id=1370343)  is fixed.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -538,6 +538,7 @@ def verify_docker_image_sha(chain, link):
         # Using downloaded image from docker hub
         task_type = link.task_type
         image_hash = cot['environment']['imageHash']
+        # Use the decision docker_image_allowlist for action tasks.
         if task_type == 'action':
             task_type = 'decision'
         if image_hash not in link.context.config['docker_image_allowlists'][task_type]:
@@ -1061,8 +1062,8 @@ async def verify_parent_task(chain, link):
     Action task verification is currently in the same verification function as
     decision tasks, because sometimes we'll have an action task masquerading as
     a decision task, e.g. in templatized actions for release graphs. To make
-    sure our guess of decision or action task isn't fatal, call this function
-    first, and then use ``is_action()`` to determine whether to call
+    sure our guess of decision or action task isn't fatal, we call this
+    function; this function uses ``is_action()`` to determine whether to call
     ``verify_decision_task`` or ``verify_action_task``.
 
     Args:
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -2,9 +2,12 @@
 """Chain of Trust artifact verification.
 
 Attributes:
+    ACTION_MACH_COMMANDS (tuple): the allowlisted mach commands for an action task,
+        ignoring options.
     DECISION_MACH_COMMANDS (tuple): the allowlisted mach commands for a decision task,
         ignoring options.
     DECISION_TASK_TYPES (tuple): the decision task types.
+    PARENT_TASK_TYPES (tuple): the parent task types.
     log (logging.Logger): the log object for this module.
 
 """
@@ -27,7 +30,7 @@ from scriptworker.context import Context
 from scriptworker.exceptions import CoTError, DownloadError, ScriptWorkerGPGException
 from scriptworker.gpg import get_body, GPG
 from scriptworker.log import contextual_log_handler
-from scriptworker.task import get_decision_task_id, get_worker_type, get_task_id
+from scriptworker.task import get_decision_task_id, get_parent_task_id, get_worker_type, get_task_id
 from scriptworker.utils import format_json, get_hash, load_json, makedirs, match_url_regex, raise_future_exceptions, rm
 from taskcluster.exceptions import TaskclusterFailure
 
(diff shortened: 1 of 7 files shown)


scriptworker/cot/verify.py
```diff
@@ -36,14 +39,16 @@
 
 DECISION_MACH_COMMANDS = ((
     './mach', 'taskgraph', 'decision'
-), (
+), )
+ACTION_MACH_COMMANDS = ((
```
escapewindow: @imbstack - should I remove all `ACTION_MACH_COMMANDS` except for `action-callback`? I'm assuming we only want to support cot verification if the action is in `actions.json`, but I want to doublecheck.

imbstack: Definitely true in the long-term. We should hold off on removing them for a bit because of older trees though?

escapewindow:
Johan, I'm flagging you because you have the most in-depth cot familiarity. I'm aware you're on PTO; I'm ok with the delay. This works for me locally, using `verify_cot` against retriggered / relpromo-ed actions. Hoping it's clear and makes sense.

escapewindow:
Hm, I think I want to put fuzzy task matching back in for fennec release pushapk.

escapewindow:
Rebased on top of scriptworker 5.1.5.

scriptworker/cot/verify.py
```diff
@@ -965,6 +1014,57 @@ def verify_firefox_decision_command(decision_link):
...
+
+
+# verify_parent_task {{{1
+async def verify_parent_task(chain, link):
+    """Verify the parent task Link.
+
+    Action task verification is currently in the same verification function as
+    decision tasks, because sometimes we'll have an action task masquerading as
+    a decision task, e.g. in templatized actions for release graphs. To make
+    sure our guess of decision or action task isn't fatal, call this function
```
JohanLorenzo: Outdated comment? `verify_parent_task` already calls `is_action()` and the right `verify_*_task()`. 

If not, I'm not sure to understand why calling these function once more adds more security.

scriptworker/task.py
```diff
@@ -70,16 +70,36 @@ def get_run_id(claim_task):
...
+
     Args:
         task (dict): the task dict.
 
     Returns:
-        str: the taskId.
+        str: the taskId of the decision task.
 
     """
     return task['taskGroupId']
```
JohanLorenzo: Should we raise a specific error if a decision/action task doesn't reference a taskGroupId?

scriptworker/constants.py
```diff
@@ -191,15 +191,15 @@
         "schemes": ("https", "ssh", ),
         "netlocs": ("hg.mozilla.org", ),
         "path_regexes": (
-            "^(?P<path>/mozilla-(central|unified))(/|$)",
-            "^(?P<path>/integration/(autoland|fx-team|mozilla-inbound))(/|$)",
-            "^(?P<path>/releases/mozilla-(beta|release|esr52))(/|$)",
+            "^(?P<path>/mozilla-(central|unified))(raw-file)?(/|$)",
```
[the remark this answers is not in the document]
escapewindow: The relpro action has a source of `https://hg.mozilla.org/projects/dateraw-file/00c5a978edeab9d5b01d346e67262a30830f6365/taskgraph/actions/release_promotion.py` ; this portion of the patch was specifically to get the `dateraw-file` to work. However, it looks like this link should be in the form of `https://hg.mozilla.org/projects/maple/raw-file/tip/taskcluster/taskgraph/actions/release_promotion.py`. I'm going to dig to see where the bug is.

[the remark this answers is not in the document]
escapewindow: Definitely a bug :) Great catch!!
https://reviewboard.mozilla.org/r/186114/#review191130

scriptworker/cot/verify.py
```diff
@@ -498,8 +538,8 @@ def verify_docker_image_sha(chain, link):
         # Using downloaded image from docker hub
         task_type = link.task_type
         image_hash = cot['environment']['imageHash']
-        # XXX we will need some way to allow trusted developers to update these
-        # allowlists
+        if task_type == 'action':
+            task_type = 'decision'
```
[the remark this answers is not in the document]

scriptworker/task.py
escapewindow: I don't think there's a valid way to get a task without a `taskGroupId`; am I wrong?

scriptworker/cot/verify.py
escapewindow: This is an explanation of what this does, rather than telling users to do additional steps. I can clarify.

scriptworker/cot/verify.py
```diff
@@ -420,6 +431,35 @@ def is_try(task):
...
+def is_action(task):
+    """Determine if a task is an action task.
+
+    Trusted decision and action tasks are important in that they can generate
+    other valid tasks. The verification of decision and action tasks is slightly
+    different, so we need to be able to tell them apart.
+
+    This checks for the following things::
+
+        * ``task.payload.env.ACTION_CALLBACK`` exists
```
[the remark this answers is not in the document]
escapewindow: Either one can exist and it'll mark it as an action. I can add a test case.

scriptworker/task.py
JohanLorenzo: Oh right, I think that was possible with the Scheduler API, but that's out of scope of CoT.

PR #162, opened 2017-10-05
bug 1395578 - retry_request should retry on server timeout.

escapewindow:


escapewindow:
Thanks! Can you mark this as approved?

PR #163, opened 2017-10-06
don't run docker-worker tests on the chain object

escapewindow:
@MihaiTabara this is another one you can look at :) No rush on any of these.

`verify_cot` breaks when run against a docker-worker task, because we
try to verify chain.cot for docker-worker specific info. Because the
chain object is assumed to be the currently running task, we don't
create a cot object for it.

I ran into this while testing

```
verify_cot --task-type l10n --cleanup -- Tz_iPxtTRqCLBg64rHK8uQ
```

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1280,14 +1280,19 @@ async def verify_docker_worker_task(chain, link):
 
     Args:
         chain (ChainOfTrust): the chain we're operating on
-        obj (ChainOfTrust or LinkOfTrust): the trust object for the signing task.
+        link (ChainOfTrust or LinkOfTrust): the trust object for the signing task.
 
     Raises:
         CoTError: on failure.
 
     """
-    check_interactive_docker_worker(link)
-    verify_docker_image_sha(chain, link)
+    if chain != link:
+        # These two checks will die on `link.cot` if `link` is a ChainOfTrust
+        # object (e.g., the task we're running `verify_cot` against is a
+        # docker-worker task). So only run these tests if they are not the chain
+        # object.
+        check_interactive_docker_worker(link)
+        verify_docker_image_sha(chain, link)
 
 
 # verify_generic_worker_task {{{1
@@ -1296,7 +1301,7 @@ async def verify_generic_worker_task(chain, link):
 
     Args:
         chain (ChainOfTrust): the chain we're operating on
-        obj (ChainOfTrust or LinkOfTrust): the trust object for the signing task.
+        link (ChainOfTrust or LinkOfTrust): the trust object for the signing task.
 
     Raises:
         CoTError: on failure.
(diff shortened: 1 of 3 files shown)


MihaiTabara:
Am hoping to get this done before Aki's PTO. Otherwise, by Monday for sure. If it needs to be completed before that, please feel free to 302 to @JohanLorenzo.

scriptworker/cot/verify.py
```diff
@@ -1286,8 +1286,13 @@ def check_num_tasks(chain, task_count):
         CoTError: on failure.
 
     """
-    check_interactive_docker_worker(link)
-    verify_docker_image_sha(chain, link)
+    if chain != link:
+        # These two checks will die on `link.cot` if `link` is a ChainOfTrust
+        # object (e.g., the task we're running `verify_cot` against is a
+        # docker-worker task. So only run these tests if they are not the chain
```
MihaiTabara: Nit: Missing closing round bracket `)` after "task".

scriptworker/test/test_cot_verify.py
```diff
@@ -1109,9 +1109,17 @@ def test_check_num_tasks(chain, num, raises):
...
 async def test_verify_docker_worker_task(mocker):
-    mocker.patch.object(cotverify, 'check_interactive_docker_worker', new=noop_sync)
-    mocker.patch.object(cotverify, 'verify_docker_image_sha', new=noop_sync)
-    await cotverify.verify_docker_worker_task(mock.MagicMock(), mock.MagicMock())
+    m = mock.MagicMock()
+    m2 = mock.MagicMock()
+    check = mock.MagicMock()
+    mocker.patch.object(cotverify, 'check_interactive_docker_worker', new=check.method1)
+    mocker.patch.object(cotverify, 'verify_docker_image_sha', new=check.method2)
+    await cotverify.verify_docker_worker_task(m, m)
```
MihaiTabara: Nit: not that in matters from a functionality perspective, but I'd rename `s/m/chain` and `s/m2/link`. And at this statement we could call it `cotverify.verify_docker_worker_task(chain, link=chain)`. But I could be wrong.

scriptworker/cot/verify.py
```diff
@@ -1286,8 +1286,13 @@ def check_num_tasks(chain, task_count):
         CoTError: on failure.
 
     """
-    check_interactive_docker_worker(link)
```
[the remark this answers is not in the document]
escapewindow: Good catch! The docstring is out of date or incorrect; amending.

scriptworker/cot/verify.py

scriptworker/test/test_cot_verify.py
escapewindow: Good call on the renames; done. I think it might not break things to use link=chain, but it's an arg not a kwarg, so it's probably less correct. I'm calling it with `cotverify.verify_docker_worker_task(chain, chain)` .

scriptworker/cot/verify.py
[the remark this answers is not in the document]
MihaiTabara: FWIW, same under `verify_generic_worker_task` below.

scriptworker/test/test_cot_verify.py
MihaiTabara: Yeah, makes more sense that way 👍

Issue #164, opened 2017-10-10
bug 1407380 - Add error message on python crash


escapewindow:
At [this line](https://github.com/mozilla-releng/scriptworker/blob/46f913f/scriptworker/task.py#L175-L177), we print `"exit code: {}".format(exitcode)` to the logfile.

In [bug 1398800](https://bugzilla.mozilla.org/show_bug.cgi?id=1398800), we're hitting a lot of `exit code: -11`. This is an issue we need to address elsewhere, but in the meantime, we can add an informative error message that will be more easily spotted.

Let's do something like

```
if exitcode == -11:
    status_line = "Automation Error: python exited with signal {}".format(exitcode)
```

before the `log.info` and `print` at the bottom of that block.

PR #166, opened 2017-10-15
Bug 1407380 - Add error message on python crash

Dyex719:
Add a helpful error message


Dyex719:
I think I understand, thank you for the idea!

scriptworker/test/test_task.py
```diff
@@ -150,6 +150,23 @@ def test_reclaim_task_non_409(context, successful_queue, event_loop):
 
 
 @pytest.mark.asyncio
+async def test_run_task_negative_11(context, mocker):
```
escapewindow: We lost the `@pytest.mark.asyncio` for the next test below with this change. I pushed a fix for this, just moved this up. Thank you!

escapewindow:
Fixes #164.

escapewindow:
Thank you!

scriptworker/test/test_task.py
Dyex719: oops. Sorry for the mix up. You're welcome :)

PR #168, opened 2017-10-20
allow for unsigned cot artifacts when verify_cot_signature is False

escapewindow:
This python-gnupg behavior seems to have changed, or my memory is
faulty. I remember `gpg.decrypt` returning an unsigned body when the
input was signed, and returning the original string when the input
wasn't signed. Now we're dying on `verify_chain_of_trust: True`,
`verify_cot_signature: False`, and an unsigned cot artifact.

Get around this by explicitly returning the original string if
`verify_cot_signature` is `False`, and if `gpg.decrypt` returns an empty
string.


escapewindow:
@MihaiTabara had enough of reviews yet??? ;-D

MihaiTabara:
Sorry about not reviewing this today. I opened it in a tab in the morning and closed accidentally and forgot of it. I really wish Github had a "review queue" or something to warn me for pendings.

escapewindow:
No worries! You've reviewed a lot this week.

PR #169, opened 2017-10-23
Bug 1385401 - Support optional upstreamArtifact

JohanLorenzo:
It's unit-tested, but I'd like to test the integration with pushapkscript before asking a formal review.

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -325,7 +325,9 @@ def get_upstream_artifacts_full_paths_per_task_id(context):
         context (scriptworker.context.Context): the scriptworker context.
 
     Returns:
-        dict: lists of the paths to existing upstream artifacts, sorted by task_id
+        dict, dict: lists of the paths to upstream artifacts, sorted by task_id.
+            First dict represents the existing upstream artifacts. The second one
+            maps the optional artifacts that couldn't be downloaded
 
     Raises:
         scriptworker.exceptions.ScriptWorkerTaskException: when an artifact doesn't exist.
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -17,7 +17,7 @@ from urllib.parse import unquote, urljoin
 from scriptworker.client import validate_artifact_url
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 from scriptworker.task import get_task_id, get_run_id, get_decision_task_id
-from scriptworker.utils import download_file, filepaths_in_dir, raise_future_exceptions, retry_async
+from scriptworker.utils import download_file, filepaths_in_dir, raise_future_exceptions, retry_async, add_enumerable_item_to_dict
 
 
 log = logging.getLogger(__name__)
@@ -331,18 +331,35 @@ def get_upstream_artifacts_full_paths_per_task_id(context):
         scriptworker.exceptions.ScriptWorkerTaskException: when an artifact doesn't exist.
 
     """
+    upstream_artifacts = context.task['payload']['upstreamArtifacts']
     task_ids_and_relative_paths = [
         (artifact_definition['taskId'], artifact_definition['paths'])
-        for artifact_definition in context.task['payload']['upstreamArtifacts']
+        for artifact_definition in upstream_artifacts
     ]
 
-    return {
-        task_id: [
-            get_and_check_single_upstream_artifact_full_path(context, task_id, path)
-            for path in paths
-        ]
-        for task_id, paths in task_ids_and_relative_paths
-    }
+    optional_artifacts_per_task_id = get_optional_artifacts_per_task_id(upstream_artifacts)
+
+    upstream_artifacts_full_paths_per_task_id = {}
+    failed_paths_per_task_id = {}
+    for task_id, paths in task_ids_and_relative_paths:
+        for path in paths:
+            try:
+                path_to_add = get_and_check_single_upstream_artifact_full_path(context, task_id, path)
+                add_enumerable_item_to_dict(
+                    dict_=upstream_artifacts_full_paths_per_task_id,
+                    key=task_id, item=path_to_add
+                )
+            except ScriptWorkerTaskException:
+                if path in optional_artifacts_per_task_id.get(task_id, []):
+                    log.warn('Optional artifact "{}" of task "{}" not found'.format(path, task_id))
+                    add_enumerable_item_to_dict(
+                        dict_=failed_paths_per_task_id,
+                        key=task_id, item=path
+                    )
+                else:
+                    raise
+
+    return upstream_artifacts_full_paths_per_task_id, failed_paths_per_task_id
 
 
 def get_and_check_single_upstream_artifact_full_path(context, task_id, path):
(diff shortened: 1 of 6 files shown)


escapewindow:
Started a review; I'll have to get back to it tomorrow..

scriptworker/cot/verify.py
```diff
@@ -504,6 +506,12 @@ def verify_docker_image_sha(chain, link):
     cot = link.cot
     task = link.task
     errors = []
+
+    if not cot:
+        log.warn('Chain of Trust for {} does not exist. See above log for more details. \
+Skipping docker image sha verification'.format(task['taskId']))
+        return
```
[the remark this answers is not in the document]
JohanLorenzo: We do if we want to support the case where `chainOfTrust.json.asc` wasn't generated by an upstream task.

scriptworker/cot/verify.py
```diff
@@ -714,64 +741,109 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
+
+    mandatory_artifacts_paths = await raise_future_exceptions(mandatory_artifact_tasks)
+    succeeded_optional_artifacts_paths, failed_optional_artifacts = \
+        await get_results_and_future_exceptions(optional_artifact_tasks)
 
+    if failed_optional_artifacts:
+        log.warn('Could not download {} artifacts: {}'.format(len(failed_optional_artifacts), failed_optional_artifacts))
 
-# download_firefox_cot_artifacts {{{1
-async def download_firefox_cot_artifacts(chain):
```
escapewindow: I'm not entirely sure I'm happy about removing this; I'm explicitly allowing for future alternate behavior in different products. However, we haven't had to deal with this yet, so it's possible YAGNI. #170 doesn't seem to need separate function names, so we might be ok.

scriptworker/artifacts.py
```diff
@@ -331,18 +331,35 @@ def get_upstream_artifacts_full_paths_per_task_id(context):
...
+                if path in optional_artifacts_per_task_id.get(task_id, []):
+                    log.warn('Optional artifact "{}" of task "{}" not found'.format(path, task_id))
+                    add_enumerable_item_to_dict(
+                        dict_=failed_paths_per_task_id,
+                        key=task_id, item=path
+                    )
+                else:
+                    raise
+
+    return upstream_artifacts_full_paths_per_task_id, failed_paths_per_task_id
```
escapewindow: You're changing the return type of a non-private function. I think pushapkscript is the only consumer, so we're probably ok here.

1. this means this will be a major version bump
2. Please update the docstring.

scriptworker/cot/verify.py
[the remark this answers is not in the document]
escapewindow: Ok, and we would have failed out already if we had mandatory artifacts based on this task?

scriptworker/cot/verify.py
```diff
@@ -655,23 +663,37 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
-    paths = await raise_future_exceptions(async_tasks)
+
+        if is_task_required_by_any_mandatory_artifact(chain, task_id):
+            mandatory_artifact_tasks.append(coroutine)
+        else:
+            optional_artifact_tasks.append(coroutine)
+
+    mandatory_artifacts_paths = await raise_future_exceptions(mandatory_artifact_tasks)
+    succeeded_optional_artifacts_paths, failed_optional_artifacts = \
+        await get_results_and_future_exceptions(optional_artifact_tasks)
```
escapewindow: We're going to download in two phases, which is probably fine. I'm not sure how to improve that without adding too much complexity.

scriptworker/cot/verify.py
```diff
@@ -714,64 +741,109 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
 
 
 # download_cot_artifacts {{{1
-async def download_cot_artifacts(chain, artifact_dict):
-    """Call ``download_cot_artifact`` in parallel for each key/value in ``artifact_dict``.
+async def download_cot_artifacts(chain):
```
escapewindow: Another reason for a major version bump. scriptworker 6.0.0, here we come!

scriptworker/cot/verify.py
```diff
@@ -714,64 +741,109 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
+        log.warn('Could not download {} artifacts: {}'.format(len(failed_optional_artifacts), failed_optional_artifacts))
 
-# download_firefox_cot_artifacts {{{1
-async def download_firefox_cot_artifacts(chain):
-    """Download artifacts needed for firefox chain of trust verification.
+    return mandatory_artifacts_paths + succeeded_optional_artifacts_paths
 
-    This is only task-graph.json so far.
+
+def is_task_required_by_any_mandatory_artifact(chain, task_id):
```
escapewindow: There's a lot of new artifact functions in `cot.verify`. I'm wondering if they belong here or `artifacts.py`, but I don't have a strong opinion here. If we're going to move functions around, though, we already have a major version bump coming :)

scriptworker/cot/verify.py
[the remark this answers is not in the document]
JohanLorenzo: That's correct. We error out here https://github.com/mozilla-releng/scriptworker/pull/169/files#diff-340a1f42e56ecb68f523d2badfa58686R682, if `chainOfTrust.json.asc` was needed by any mandatory artifact.

scriptworker/cot/verify.py
JohanLorenzo: Right. Maybe we could define a function like `raise_future_exceptions(mandatory_artifact_tasks, optionals=optional_artifact_tasks)`. It would perform the download and sort what completed from what's not. It can be done in a follow up, if you wish.

scriptworker/cot/verify.py
JohanLorenzo: Yeah, I thought of YAGNI too. Worst case scenario, we still have the patch in git, and we can revive it.

scriptworker/cot/verify.py
JohanLorenzo: I'm glad we think alike. I started to put some of these functions in `artifacts.py`. However, they have `chain` as an argument, which makes me think it should somewhere under `cot`. I can spin off a new package like `cot.artifacts` in a quick follow up, if you're interested.

scriptworker/artifacts.py
JohanLorenzo: Ooops, I thought I did. Thanks for spotting it. It's now fixed.

scriptworker/cot/verify.py
escapewindow: Other than the docstring I'm ok leaving the other review comments alone, unless you think we should act on them.

JohanLorenzo: Docstring was fixed in https://github.com/mozilla-releng/scriptworker/pull/169/commits/4d448b080a4942eb7f9c43dc4c4c7ec0ad1059b8 . Let's handle the rest in followups, then

scriptworker/utils.py
```diff
@@ -309,23 +309,54 @@ def create_temp_creds(client_id, access_token, start=None, expires=None,
...
-    return result
+    succeeded_results, _ = await _process_future_exceptions(tasks, raise_at_first_error=True)
+    return succeeded_results
+
+
+async def get_results_and_future_exceptions(tasks):
+    """Given a list of futures, await them, then return results and exceptions.
+
+    This is similar to raise_future_exceptions, except that it doesn't raise any
+    exception. They are returned instead. This allow some tasks to optionally fail.
```
MihaiTabara: tiny nit: s,allow,allows

scriptworker/utils.py
```diff
@@ -490,3 +521,31 @@ def match_url_regex(rules, url, callback):
...
                 return result
+
+
+def add_enumerable_item_to_dict(dict_, key, item):
+    """Add an item to a list contained in a dict.
+
+    For example: If the dict is ``{'some_key': ['an_item']}``, then calling this function
+    will alter the dict to ``{'some_key': ['an_item', 'another_item']}``.
+
+    If the key doesn't exist yet, the function initialize it with a list containing the
```
MihaiTabara: tiny nit: s,initialize,initializes

JohanLorenzo: Fixed

scriptworker/cot/verify.py
```diff
@@ -655,23 +663,37 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
+            mandatory_artifact_tasks.append(coroutine)
+        else:
+            optional_artifact_tasks.append(coroutine)
+
+    mandatory_artifacts_paths = await raise_future_exceptions(mandatory_artifact_tasks)
+    succeeded_optional_artifacts_paths, failed_optional_artifacts = \
+        await get_results_and_future_exceptions(optional_artifact_tasks)
+
+    if failed_optional_artifacts:
+        error_messages = '\n'.join(' * '.format(failed_optional_artifacts))
```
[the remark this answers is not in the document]
JohanLorenzo: Great catch. I fixed it

PR #170, opened 2017-12-04
Handle `cot_product` keying generically.

tomprince:


escapewindow:
Awesome! I need to give this a closer look, and it looks like we need coverage for https://coveralls.io/jobs/31735023/source_files/2146963557#L1401 .

scriptworker/config.py
```diff
@@ -34,6 +35,10 @@
 _VALUE_UNDEFINED_MESSAGE = "{path} {key} needs to be defined!"
 
 
+class UnknownProduct(Exception):
+    """The configured `cot_product` doesn't have a value for a given key."""
```
escapewindow: Hm, does this belong in `scriptworker.exceptions`? Might be a child of `CoTError`.

scriptworker/constants.py
```diff
@@ -204,73 +204,77 @@
     }), ),
```
JohanLorenzo: Nit: We also need to key `valid_vcs_rules`.

scriptworker/config.py
tomprince: Sure, I hadn't noticed it.  I don't think it should be a child of CoTError, since it happens on worker startup, not as part of verifying a task. I've renamed it `ConfigError`.

scriptworker/constants.py
tomprince: Sure, this will allow keying any top-level config entry, but I think adding keying can wait for a separate PR where other keys are actually added.

scriptworker/cot/verify.py
```diff
@@ -1393,14 +1392,13 @@ def get_firefox_source_url(obj):
...
     rules = {}
-    cot_product = chain.context.config['cot_product']
     for my_key, config_key in {
         'scopes': 'cot_restricted_scopes',
         'trees': 'cot_restricted_trees'
     }.items():
-        rules[my_key] = chain.context.config[config_key].get(cot_product)
+        rules[my_key] = chain.context.config[config_key]
         if not isinstance(rules[my_key], (dict, frozendict)):
-            raise_on_errors(["{} invalid for {}: {}!".format(config_key, cot_product, rules[my_key])])
```
tomprince: Thinking more about this check, now that the config type-checking looks through keyed config entries, I don't think this is needed anymore.

tomprince:
Fix the coverage by removing the now obsolete check.

escapewindow:
@tomprince Did you want this merged, or are you holding off for something else?

tomprince:
I think this is ready to merge.

PR #171, opened 2017-12-07
Add support for verifying comm-central based decision tasks.

tomprince:
Depends on #170.

PR #172, opened 2018-01-03
Bump 6.0.0

JohanLorenzo:
I just landed #169. I see #170 is ready too (but not #171). I don't mind if 6.0.0 includes #170. However, the changes there don't require a major version bump, so we can easily release 6.1.0 right after. @escapewindow, @tomprince, what do you guys think?


JohanLorenzo:
For the record, @tomprince agreed with the first comment via IRC.

PR #186, opened 2018-02-20
Add treescriptworker config

Callek:


scriptworker/constants.py
```diff
@@ -228,8 +229,9 @@
                 'project:releng:balrog:server:esr': 'esr',
                 'project:releng:beetmover:bucket:release': 'all-release-branches',
                 'project:releng:googleplay:release': 'release',
-                'project:releng:signing:cert:release-signing': 'all-release-branches',
                 'project:releng:googleplay:beta': 'beta',
+                'project:releng:signing:cert:release-signing': 'all-release-branches',
+                'project:releng:treescript:action:push': 'all-release-branches',
```
escapewindow: Do we really want to prevent version bumping and tagging on, say, `try` staging releases?

Callek: As discussed in person. We likely do want this available for try at some point, but right now there are a bunch of other things blocking try staging releases, and we can let this run without push scopes and still see an `hg out` of what will have changed anyway.

PR #187, opened 2018-02-20
Bug 1433459: Add bouncer support

JohanLorenzo:
Modeled after #181. I reorganized the scopes by alphabetical/tree order.

cc @MihaiTabara

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -413,6 +413,7 @@ def get_valid_task_types():
         'scriptworker': verify_scriptworker_task,
         'balrog': verify_balrog_task,
         'beetmover': verify_beetmover_task,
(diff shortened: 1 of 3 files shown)


scriptworker/cot/verify.py
```diff
@@ -1484,6 +1485,22 @@ def compare_jsone_task_definition(parent_link, compare_definition):
     return await verify_scriptworker_task(chain, obj)
 
 
+async def verify_bouncer_task(chain, obj):
```
escapewindow: This follows the model, and is good. @Callek skipped it and is defining his task type as `scriptworker`, which also works because the `verify_INSTANCE_task` functions only call `verify_scriptworker_task`. I think we might want to move to that model at some point. Either way works for now.

PR #192, opened 2018-02-28
set artifact expiration date to task expiration date

escapewindow:
This is what we do in-tree for docker-worker (and I think generic-worker) tasks; let's do the same in scriptworker.

### DIFF
--- a/scriptworker/test/test_context.py
+++ b/scriptworker/test/test_context.py
@@ -6,6 +6,8 @@ import json
 import os
 import pytest
 import taskcluster
+import scriptworker.context as swcontext
+from copy import deepcopy
 from . import tmpdir
 from . import rw_context as context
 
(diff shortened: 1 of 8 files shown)


scriptworker/test/test_context.py
```diff
@@ -90,3 +92,26 @@ def test_temp_queue(context, mocker):
     assert taskcluster.async.Queue.called_once_with({
         'credentials': context.temp_credentials,
     }, session=context.session)
+
+
+@pytest.mark.asyncio
+async def test_projects(context, mocker):
```
JohanLorenzo: I'm not sure to follow what this test does in regard to expiration date.

escapewindow: I noticed we were missing some coverage without integration tests, possibly after I made some artifact test changes.

escapewindow:
Thanks!

PR #194, opened 2018-03-03
Private artifacts

escapewindow:
We potentially need private artifact support for partner repacks; we may need it for other purposes in the future. Per [the docs](https://docs.taskcluster.net/manual/tasks/artifacts#public-and-private-artifacts), any public artifact path starts with `public/`. All other artifacts are private.

To access these artifacts, we need a client with the scope `queue:get-artifact:<path>`; I'm thinking we'll use `partner/<name>/...`. We can add these to the appropriate scriptworker clientIds when needed.

This is tested! and successful: https://bugzilla.mozilla.org/show_bug.cgi?id=1442793#c2
https://tools.taskcluster.net/groups/fyuV8WLnTmqNQHTEodo2sQ/tasks/Fi181XtXQuyKjzZk38lNiQ/runs/0/artifacts


scriptworker/test/test_utils.py
```diff
@@ -409,6 +409,16 @@ def test_load_json_or_yaml(string, is_path, exception, raises, result):
     ) == {"credentials": ["blah"]}
 
 
+# get_loggable_url
+@pytest.mark.parametrize("url,expected",((
+    "https://foo/bar", "https://foo/bar"
+), (
+    "https://foo/bar?bewit=foobar", "https://foo/bar?<snip>"
```
JohanLorenzo: Nit: `AWSAccessKeyId` is missing from the test

scriptworker/artifacts.py
```diff
@@ -232,18 +238,13 @@ def get_artifact_url(context, task_id, path):
...
-            unquote(context.queue.makeRoute('getLatestArtifact', replDict={
-                'taskId': task_id,
-                'name': path
-            }))
+    if path.startswith("public/"):
+        url = context.queue.buildUrl('getLatestArtifact', task_id, path)
+    else:
+        url = context.queue.buildSignedUrl(
+            'getLatestArtifact', task_id, path,
+            # XXX Can set expiration kwarg in (int) seconds from now;
```
JohanLorenzo: Minor question: Should this comment be a `TODO`?

scriptworker/cot/verify.py
```diff
@@ -686,8 +686,9 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
 
     if path not in link.cot['artifacts']:
         raise CoTError("path {} not in {} {} chain of trust artifacts!".format(path, link.name, link.task_id))
+    # Don't log the url, which may be signed/private
+    log.info("Downloading Chain of Trust artifact:\n{} {}".format(task_id, path))
```
JohanLorenzo: Curiosity: Out of curiosity, why don't we out `loggable_url` in this context?

scriptworker/artifacts.py
escapewindow: I think 15min works for now, since cot verification takes <1min these days; I was just leaving a note in case we want to modify it in the future.

scriptworker/cot/verify.py
escapewindow: Sure. I think it's because I made this change before I wrote `get_loggable_url` :)

PR #196, opened 2018-03-13
Factorize functions in *script part 2

JohanLorenzo:
Follow up #195. Mainly, from discussion in https://github.com/mozilla-releng/signingscript/pull/47#pullrequestreview-103124279

### DIFF
--- a/scriptworker/client.py
+++ b/scriptworker/client.py
@@ -161,6 +161,7 @@ def _init_context(config_path=None):
         config_path = sys.argv[1]
 
     context.config = load_json_or_yaml(config_path, is_path=True)
+    context.task = get_task(context.config)
     return context
 
 
--- a/scriptworker/client.py
+++ b/scriptworker/client.py
@@ -134,7 +134,7 @@ def validate_artifact_url(valid_artifact_rules, valid_artifact_task_ids, url):
     return unquote(filepath).lstrip('/')
 
 
-def sync_main(async_main, config_path=None, should_validate_task=True):
+def sync_main(async_main, config_path=None, default_config=None, should_validate_task=True):
     """Entry point for scripts using scriptworker.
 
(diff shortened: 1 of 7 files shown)


scriptworker/client.py
```diff
@@ -148,19 +148,28 @@ def sync_main(async_main, config_path=None):
...
         config_path = sys.argv[1]
 
-    context.config = load_json_or_yaml(config_path, is_path=True)
+    context.config = {} if default_config is None else default_config
+    context.config.update(load_json_or_yaml(config_path, is_path=True))
+
+    context.task = get_task(context.config)
+
+    # This prevents *script from pasting the whole context.
+    context.write_json = noop_sync
```
escapewindow: To be safer, let's put this at the beginning. It looks like we've removed all `write_json` calls except in `claim_task`, but we may add them in the future.

scriptworker/test/__init__.py
```diff
@@ -15,7 +15,7 @@
 from scriptworker.config import get_unfrozen_copy, apply_product_config
 from scriptworker.constants import DEFAULT_CONFIG
 from scriptworker.context import Context
-from scriptworker.utils import makedirs
+from scriptworker.utils import makedirs, noop_sync
```
escapewindow: We're importing here so other test modules can import from here?

JohanLorenzo: That's right. Do you think tests should directly import from `scriptworker.utils`?

scriptworker/client.py
JohanLorenzo: OK. Done!

scriptworker/test/__init__.py
escapewindow: Eh, both work. Flake8 would probably complain if I hadn't turned it off for `scriptworker/test`; since we're using the imported function, this works.

PR #198, opened 2018-03-26
Bug 1447263 - Add push-snap support

JohanLorenzo:
Modeled after #187

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -245,6 +245,10 @@ DEFAULT_CONFIG = frozendict({
                 'project:releng:signing:cert:nightly-signing': 'all-nightly-branches',
                 'project:releng:signing:cert:release-signing': 'all-release-branches',
 
+                'project:releng:snapcraft:firefox:beta': 'beta',
+                'project:releng:snapcraft:firefox:candidate': 'release',
+                'project:releng:snapcraft:firefox:mock': 'all-staging-branches',
+
                 'project:releng:ship-it:production': 'all-production-branches',
                 'project:releng:ship-it:staging': 'all-staging-branches',
 
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -422,6 +422,7 @@ def get_valid_task_types():
         'decision': verify_parent_task,
         'docker-image': verify_docker_image_task,
         'pushapk': verify_pushapk_task,
+        'pushsnap': verify_pushsnap_task,
         'shipit': verify_shipit_task,
         'signing': verify_signing_task,
         'partials': verify_partials_task,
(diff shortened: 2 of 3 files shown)


JohanLorenzo:
Tested at: https://tools.taskcluster.net/groups/HREGN7z0SiWJTRJ3GbvrXA/tasks/bTpaugywRBm0u66-M87dHw/runs/11/logs/public%2Flogs%2Fchain_of_trust.log

scriptworker/constants.py
```diff
@@ -245,6 +245,10 @@
                 'project:releng:signing:cert:nightly-signing': 'all-nightly-branches',
                 'project:releng:signing:cert:release-signing': 'all-release-branches',
 
+                'project:releng:snapcraft:firefox:beta': 'beta',
+                'project:releng:snapcraft:firefox:candidate': 'release',
+                'project:releng:snapcraft:firefox:mock': 'all-staging-branches',
```
escapewindow: The naming here doesn't necessarily match the others. I'd prefer to start standardizing soon, and so far release/beta/staging seems to be the terminology. Is there a strong reason to avoid that?

JohanLorenzo: I see the confusion. I used this naming because it matches what the [Snap store exposes](https://docs.snapcraft.io/reference/channels#risk-levels-meaning). `release` could be used instead of `candidate`, but like balrog channels, we want to push to candidate before going live.
Regarding the `mock` value, I created it. I didn't use `staging` because we don't have a staging instance of the snap store. Instead, I let pushsnapscript to not call any instance when [this value is passed](https://github.com/JohanLorenzo/pushsnapscript/commit/3a8eb4c3de92403b0b5e202a550efe69c79f8d01#diff-62d14b5f427f0c76b5db0e636464a08cR21).

escapewindow: I think both sets of names make sense, depending on your viewpoint.

If you're diving deep into snap, then using snap-specific terms like `candidate` or `mock` make more sense than `release` or `staging`, because they are more descriptive of what happens.

If you're trying to maintain the set of all scriptworkers (or the set of all scopes), then having to remember that we need `dep-signing` for signing and `mock` for snap is a lot trickier. From that perspective, standardizing on naming is much more useful.

Right now I'm leaning towards the latter, which means we'll need to decide on a standard, revisit all non-standard scopes, hostnames, workerTypes, etc. etc., and standardize them. We can start here with `release` and `staging` and have less cleanup to do to standardize these names, or we can go with `candidate` and `mock` here, and add snap to the list of things we'll need to clean up if and when we want to standardize.

JohanLorenzo: I see your point of view. I understand the nomenclature in `constants.py` would benefit from standard names. We can always map these standard names to the real ones in *script. I don't have much preference at the moment. Let's make a call with other folks. I filed [bug 1449508](https://bugzilla.mozilla.org/show_bug.cgi?id=1449508) for it.

JohanLorenzo: Per bug 1449508, we're okay to keep descriptive names. I'll merge and release a new version, then.

PR #200, opened 2018-04-05
Aiohttp3 support

escapewindow:
This patch gives us aiohttp 3 support. We'll need to also check that the scripts don't use any `with aiohttp.ClientSession() ...` constructs; we should also be able to fix any session unclosed or session-created-outside-of-coroutine warnings.

In parallel, we need python >=3.5.3. I'm working on a puppet patchset to create python 3.6.5 and unhardcode 3.5 from all the paths and names.

I'd like to land this PR on master so it doesn't bitrot, but we won't be able to update to aiohttp3 until the other requirements are ready, so maybe we should leave this unmerged until we're closer to being ready.


scriptworker/test/test_integration.py
```diff
@@ -5,7 +5,8 @@
 import aiohttp
 import arrow
 import asyncio
-from contextlib import contextmanager
+from asyncio_extras.contextmanager import async_contextmanager
+from async_generator import yield_
```
Callek: oo TIL (might be worth a filed issue about removing this when min is >= py3.6 due to it "just working" there)

setup.py
```diff
@@ -17,7 +17,7 @@
     sys.exit(1)
 
 reqs = [
-    "aiohttp>=2.0.0,<3",
+    "aiohttp",
```
Callek: should update requirements* with dephash

tox.ini
```diff
@@ -14,6 +14,7 @@ passenv =
     TRAVIS_BRANCH
 
 deps =
+    asyncio_extras
```
Callek: requirements* update for this.

Callek: Probably also want async_generator listed

Callek:
@escapewindow taskcluster client has async_timeout<3.0 so we'll need to fix that...

PR #201, opened 2018-04-10
Bug 1421062: Add restricted scope definitions for thunderbird.

tomprince:


scriptworker/constants.py
```diff
@@ -255,6 +255,18 @@
...
             }),
             'thunderbird': frozendict({
+                'project:comm:thunderbird:releng:balrog:server:nightly': 'all-nightly-branches',
+                'project:comm:thunderbird:releng:balrog:server:beta': 'beta',
+                'project:comm:thunderbird:releng:balrog:server:esr': 'esr',
+
+                'project:comm:thunderbird:releng:beetmover:bucket:nightly': 'all-nightly-branches',
+                'project:comm:thunderbird:releng:beetmover:bucket:release': 'all-release-branches',
+
+                'project:comm:thunderbird:releng:bouncer:server:production': 'all-production-branches',
```
escapewindow: I'm not sure if `all-production-branches` is clear - we run a mixed production/staging env on maple for Firefox, for instance. Release signing with staging balrog/beetmover/bouncer/etc. Right now `all-production-branches` appears to match `all-release-branches`.

scriptworker/constants.py
```diff
@@ -255,6 +255,18 @@
...
+                'project:comm:thunderbird:releng:bouncer:server:staging': 'all-staging-branches',
```
escapewindow: I'm not sure you need to list staging as a restricted scope. Unless I'm mistaken, atm bouncer:server:staging is locked down to 0 branches (no branch is allowed to use it), which is more restricted than production.

I suppose I could see some benefit to avoid using staging on a release branch, but for the most part we've been saying the staging scopes are unrestricted and can be used anywhere.

PR #203, opened 2018-04-10
Distribute wheels of scriptworker.

tomprince:
This gets rid of uploading gpg signatures to pypi as well (see [here](https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/) for reasoning).

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -100,27 +100,25 @@ Someone with access to the scriptworker package on `pypi.python.org` needs to do
     # cleartext auth!
     # Using a python with `twine` in the virtualenv:
     VERSION=4.1.2
-    # create the source tarball
-    python setup.py sdist
-    # sign the source tarball
-    gpg --detach-sign -a dist/scriptworker-${VERSION}.tar.gz
+    # create the source tarball and wheel
+    python setup.py sdist bdist_wheel
     # upload the source tarball + signature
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -101,26 +101,24 @@ Someone with access to the scriptworker package on `pypi.python.org` needs to do
     # Using a python with `twine` in the virtualenv:
     VERSION=4.1.2
     # create the source tarball
-    python setup.py sdist
-    # sign the source tarball
-    gpg --detach-sign -a dist/scriptworker-${VERSION}.tar.gz
+    python setup.py sdist bdist_wheel
     # upload the source tarball + signature
-    twine upload dist/scriptworker-${VERSION}.tar.gz{,.asc}
+    twine upload dist/scriptworker-${VERSION}.tar.gz dist/scriptworker-${VERSION}-py2.py3-none-any.whl
```
escapewindow: Any reason you removed the gpg signature? We could gpg sign both, right?

tomprince: We could, but they are going to go away in the medium term anyway.

https://caremad.io/posts/2013/07/packaging-signing-not-holy-grail/ has a discussion of why they aren't generally useful from the primary maintainer of pypi.

tomprince: To expand on this, eventually pypi is going to use https://theupdateframework.github.io/ instead of gpg signatures.

PR #208, opened 2018-04-24
cron@noreply.mozilla.org owns cron tasks

escapewindow:
e.g. https://tools.taskcluster.net/groups/bWCCoBqpQEiCSwI2gJOHXw/tasks/bWCCoBqpQEiCSwI2gJOHXw/details

Without this change, `verify_cot --task-type decision bWCCoBqpQEiCSwI2gJOHXw` passes but has to fall back to cotv1 (`DEPRECATED_DECISION_TASK`).

With this change, `verify_cot --task-type decision bWCCoBqpQEiCSwI2gJOHXw` passes without `DEPRECATED_DECISION_TASK`.


escapewindow:
Ok. In the future, when cotv1 is disabled, the above change would have broken cot for nightlies; please be aware when making those changes.

PR #209, opened 2018-04-25
Support mozilla-mobile chains

JohanLorenzo:
~~**Do not merge yet**~~ Ready to land!

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -181,6 +181,9 @@ DEFAULT_CONFIG = frozendict({
         "gecko-2-decision",
         "gecko-3-decision",
         "gecko-decision",  # legacy
+        # gecko-focus is for mozilla-mobile releases. It's named this way because it's a worker type
+        # but using the gecko ChainOfTrust keys. See bug 1455290 for more details.
+        "gecko-focus",
     ),
 
     # docker-image cot
@@ -215,6 +218,13 @@ DEFAULT_CONFIG = frozendict({
                     "^(?P<path>/comm-central)(/|$)",
                     "^(?P<path>/releases/comm-(beta|esr\d+))(/|$)",
                     "^(?P<path>/(try-comm-central))(/|$)",
+                ),
+            }),),
+            'mobile': (frozendict({
+                "schemes": ("https", "ssh", ),
+                "netlocs": ("github.com", ),
+                "path_regexes": (
+                    "^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
                 ),
             }),),
         }),
@@ -268,6 +278,10 @@ DEFAULT_CONFIG = frozendict({
 
                 'project:comm:thunderbird:releng:signing:cert:nightly-signing': 'all-nightly-branches',
                 'project:comm:thunderbird:releng:signing:cert:release-signing': 'all-release-branches',
+            }),
+            'mobile': frozendict({
+                'project:mobile:focus:googleplay:product:focus': 'focus-repo',
+                'project:mobile:focus:releng:signing:cert:release-signing': 'focus-repo',
             }),
         }),
     },
@@ -356,18 +370,25 @@ DEFAULT_CONFIG = frozendict({
                     "/comm-central",
                 ),
             }),
+            'mobile': frozendict({
+                'focus-repo': (
+                    '/mozilla-mobile/focus-android',
+                ),
+            }),
         }),
     },
     'source_env_prefix': {
         'by-cot-product': frozendict({
             'firefox': 'GECKO',
             'thunderbird': 'COMM',
+            'mobile': '',   # mozilla-mobile projects are built on GitHub instead of hg.m.o
         })
     },
     'extra_run_task_arguments': {
         'by-cot-product': frozendict({
             'firefox': (),
             'thunderbird': ('--comm-checkout=',),
+            'mobile': (),
         })
     },
 })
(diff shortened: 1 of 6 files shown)


scriptworker/constants.py
```diff
@@ -179,6 +179,7 @@
         "gecko-2-decision",
         "gecko-3-decision",
         "gecko-decision",  # legacy
+        "gecko-focus",  # mozilla-mobile releases
```
escapewindow: This seems like a misnomer - Focus is not Gecko, correct?
It doesn't block anything, but we may want to spin up a 2nd workerType at some point with a more accurate name, and then retire gecko-focus.

scriptworker/cot/verify.py
```diff
@@ -803,13 +803,13 @@ def get_all_artifacts_per_task_id(chain, upstream_artifacts):
...
             )
         # Download actions.json for decision+action task cot verification
-        if link.task_type in DECISION_TASK_TYPES:
-            add_enumerable_item_to_dict(
-                dict_=all_artifacts_per_task_id, key=link.task_id, item='public/actions.json'
-            )
-            add_enumerable_item_to_dict(
-                dict_=all_artifacts_per_task_id, key=link.task_id, item='public/parameters.yml'
-            )
+        # if link.task_type in DECISION_TASK_TYPES:
```
escapewindow: For now, we could either create a size 0 file in the Focus decision task, or have a map of `cot_product` to `decision_task_artifacts`. When taskgraph is a generic module, we should get these files for free.

scriptworker/cot/verify.py
```diff
@@ -1395,7 +1395,7 @@ def compare_jsone_task_definition(parent_link, compare_definition):
             mach_commands = ACTION_MACH_COMMANDS
         else:
             mach_commands = DECISION_MACH_COMMANDS
-        verify_decision_command(link, mach_commands=mach_commands)
+        # verify_decision_command(link, mach_commands=mach_commands)
```
escapewindow: This check should go away when we kill cotv1.

scriptworker/cot/verify.py
```diff
@@ -1637,7 +1637,7 @@ def check_num_tasks(chain, task_count):
         # docker-worker task). So only run these tests if they are not the chain
         # object.
         check_interactive_docker_worker(link)
-        verify_docker_image_sha(chain, link)
+        # verify_docker_image_sha(chain, link)
```
escapewindow: This check should go away when we kill cotv1.

scriptworker/cot/verify.py
JohanLorenzo: Great. Thanks for the idea!

scriptworker/cot/verify.py
JohanLorenzo: Good idea. I'll do it this way.

scriptworker/cot/verify.py
```diff
@@ -1281,6 +1281,12 @@ def verify_decision_command(decision_link, mach_commands=DECISION_MACH_COMMANDS)
         CoTError: on failure.
 
     """
+    if 'mozilla-mobile/focus-android' in get_source_url(parent_link):
+        log.warn(
+            'Repo "mozilla-mobile/focus-android" does not support JSON-e yet. Skipping parent task verifications'
```
escapewindow: looks like we're not fully verifying by either cotv1 or cotv2?

JohanLorenzo: That's right. I think it's okay to leave it until https://github.com/taskcluster/taskcluster-github/pull/239 gets implemented. No APK can be published to users if the decision task gets altered.

Are you okay with it?

escapewindow: We're not checking the decision task command, and we're not checking the decision task definition - how do we know if it's altered?

JohanLorenzo: > No APK can be published to users if the decision task gets altered.

My apologies, this wasn't very clear. I meant: if an attackers gets the right scopes and forges a decision task, then this person can't get the APK live (unless the attacker has access to Google Play too). There is no alteration verification per se, like you said.

JohanLorenzo:
I'll add tests about this by-pass, if that's okay to you @escapewindow

scriptworker/cot/verify.py
escapewindow: Ok. As long as the Focus team is aware of this hole and don't publish an apk without double checking that it's one they want to publish, this is probably ok.

scriptworker/constants.py
```diff
@@ -267,6 +277,10 @@
                 'project:comm:thunderbird:releng:signing:cert:nightly-signing': 'all-nightly-branches',
                 'project:comm:thunderbird:releng:signing:cert:release-signing': 'all-release-branches',
             }),
+            'mobile': frozendict({
```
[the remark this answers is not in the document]
JohanLorenzo: Good point. I think `focus-repo` is better. At the moment, having a single `cot_product` is easier to implement for mobile projects.

scriptworker/cot/verify.py
JohanLorenzo: Actually, having a condition on the `cot_product` instead of the repo is safer. I'll use. Thanks for the hint!

JohanLorenzo:
Rebased on top of master. Still tests to add.

escapewindow:
Ok. Be aware #221 removes a bunch of code --  lmk if you want this to land first.

JohanLorenzo:
I'd like to land it before #211. The PR is now ready, IMO.

escapewindow:
Good work! Do you need any help with shpiping this?

JohanLorenzo:
It's all good. I just deployed it. 

Thanks for the huge help on making focus automatically signed!

PR #212, opened 2018-04-27
allow for try syntax in comment (only first line)

escapewindow:
Evidently `./mach try` now appends a "Pushed via `mach try syntax`" line to the try commit message. The net result is that the task's GECKO_COMMIT_MSG only matches the first line of the push's commit message.

Without this change, `verify_cot --cleanup --task-type decision UTeFpv0GRZ-Z0P5Fv1oBeg` falls back to cotv1. With this change, it verifies cleanly in cotv2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [10.6.1] - 2018-04-30
+### Fixed
+- restrict compariston to  the first line of hg push comments for try
+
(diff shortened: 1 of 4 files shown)


tomprince:
Looking at the code that generates `GECKO_COMMIT_MSG` ([here](https://github.com/taskcluster/mozilla-taskcluster/blob/master/src/jobs/taskcluster_graph.js#L46-L58)). I think we should either always just look at the first line or look for try syntax explicitly.

I think we shouldn't look at the entire commit message.

escapewindow:
I thought about that too. I was thinking about leaving the commit message in case github repos behaved differently, but it's easy to remove.

PR #213, opened 2018-05-01
Fix try comments part 2

escapewindow:
We don't just use the first line of a comment in a try push; we trim everything before `try:`.

Without this change, `verify_cot --task-type decision --cleanup IfGymcgYRnCwGP3uR9wA9w` fails cotv2 and fails back to cotv1. With this change, `verify_cot --task-type decision --cleanup IfGymcgYRnCwGP3uR9wA9w` works.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1139,11 +1139,9 @@ async def _get_additional_hgpush_jsone_context(parent_link, decision_link):
     decision_comment = get_commit_message(decision_link.task)
     push_comment = pushlog_info['pushes'][pushlog_id]['changesets'][0]['desc']
     allowed_comments = [' ']
-    # try syntax uses the first line of the commit.
-    first_line = push_comment.split('\n')[0]
-    if 'try:' in first_line:
-        parts = first_line.split('try:')
-        allowed_comments.append('try:{}'.format(parts[-1]))
+    if 'try:' in push_comment:
+        try_idx = push_comment.index('try:')
+        allowed_comments.append(push_comment[try_idx:].split('\n')[0])
     if decision_comment not in allowed_comments:
         raise CoTError(
             "Decision task {} comment doesn't match the push comment!\n"
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1138,9 +1138,13 @@ async def _get_additional_hgpush_jsone_context(parent_link, decision_link):
     pushlog_id = list(pushlog_info['pushes'].keys())[0]
     decision_comment = get_commit_message(decision_link.task)
     push_comment = pushlog_info['pushes'][pushlog_id]['changesets'][0]['desc']
+    allowed_comments = [' ']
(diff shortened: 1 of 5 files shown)


escapewindow:
Most recent push to use `parts[-1]` rather than `parts[1]`, in case of multiple `try:`s in a comment =\

scriptworker/cot/verify.py
```diff
@@ -1138,9 +1138,13 @@ def verify_decision_command(decision_link, mach_commands=DECISION_MACH_COMMANDS)
     pushlog_id = list(pushlog_info['pushes'].keys())[0]
     decision_comment = get_commit_message(decision_link.task)
     push_comment = pushlog_info['pushes'][pushlog_id]['changesets'][0]['desc']
+    allowed_comments = [' ']
     # try syntax uses the first line of the commit.
     first_line = push_comment.split('\n')[0]
-    if decision_comment not in (' ', first_line):
+    if 'try:' in first_line:
+        parts = first_line.split('try:')
```
tomprince: Maybe something like:
```python
_, try_syntax = first_line.split('try:', 1)
```
?

It [looks](https://github.com/taskcluster/mozilla-taskcluster/blob/cf5ce67127c694e01a9fa3f0011e9aae6ac5169e/src/jobs/taskcluster_graph.js#L43-L58) like taskcluster splits at the first occurrence of `try:` in a line. Actually, it looks like it will find that syntax anywhere in the commit message even, though I'm not sure we want to care.

scriptworker/cot/verify.py
```diff
@@ -1139,11 +1139,9 @@ def verify_decision_command(decision_link, mach_commands=DECISION_MACH_COMMANDS)
...
     push_comment = pushlog_info['pushes'][pushlog_id]['changesets'][0]['desc']
     allowed_comments = [' ']
-    # try syntax uses the first line of the commit.
-    first_line = push_comment.split('\n')[0]
-    if 'try:' in first_line:
-        parts = first_line.split('try:')
-        allowed_comments.append('try:{}'.format(parts[-1]))
+    if 'try:' in push_comment:
+        try_idx = push_comment.index('try:')
+        allowed_comments.append(push_comment[try_idx:].split('\n')[0])
```
tomprince: I think it would be worthwhile to link to either or both of where this code comes from, or the logic this is mirroring comes from.

scriptworker/test/test_cot_verify.py
```diff
@@ -1282,6 +1282,59 @@ def die(*args):
...
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize("push_comment,task_comment,raises", ((
+    'foo bar baz', ' ', False
+), (
+    'try: a b c', 'try: a b c', False
+), (
+    'blah blah blah blah try: a b c', 'try: a b c', False
+), (
+    'asdfsadfsad', '', True
```
tomprince: Maybe add a couple of multi-line comments tests here?

PR #221, opened 2018-05-14
Cotv3: support action hooks; drop cotv1 support

escapewindow:
There are two hardcodes we should deal with before landing: `repo_scope` and the action-hook `taskGroupId` weirdness.

If we do drop cotv1, we also need to:
- update puppet templates to stop populating docker sha allowlists
- update in-tree comments to remove references to docker sha allowlists


escapewindow:
Rebased against #209.

escapewindow:
We shouldn't roll this out until https://github.com/taskcluster/taskcluster-hooks/pull/100 is rolled out.

escapewindow:
@JohanLorenzo I'd be interested in your opinion here. I'd like to make sure I'm not breaking `pushapk` or `focus` assumptions here.

@tomprince any feedback here? I don't want to break Thunderbird either.

If either or both of you is up for a review, that would be wonderful. This is a pretty big patchset, so I'd be happy to jump on a video call and explain or walk through if that helps the process.

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+- added `verify_repo_matches_url` to stop using `.startswith()` to compare urls
+- added `REPO_SCOPE_REGEX` to allow us to find the `repo_scope` in a task's scopes.
+- added `get_repo_scope` to return the `repo_scope` in a task's scopes (or `None`)
+- added a `test/data/cotv3` dir for action hook test data.
+
+### Changed
+- set `cot_version` to 3.
+- set `min_cot_version` to 2.
+- we now require cot artifacts in `verify_docker_image_sha`.
+- we no longer check docker image shas against an allowlist; they either match chain of trust artifact shas, or they're a task type that allows prebuilt docker images. If these are defined in-tree, we trace the request to the tree, so these should be as trustable as the tree in question.
```
tomprince: I'm not sure I understand the motivation for dropping this. While we trust trust the tree, it seems a bunch of the logic in scriptworker is providing an additional level of protection.

Maybe it would be good to have an explicit threat model for what protection scriptworker is providing.

scriptworker/cot/verify.py
```diff
@@ -887,11 +852,7 @@ def verify_task_in_task_graph(task_link, graph_defn, level=logging.CRITICAL):
     errors = []
     runtime_defn = deepcopy(task_link.task)
     # dependencies
-    # Allow for a subset of dependencies in a retriggered task.  The current use case
-    # is release promotion: we may hit the expiration deadline for a task (e.g. pushapk),
-    # and a breakpoint task in the graph may also hit its expiration.  To kick off
-    # the pushapk task, we can clone the task, update timestamps, and remove the
-    # breakpoint dependency.
```
tomprince: If I'm reading the code correctly, we still allow the actual task to have less dependencies than the one from the decision task?

scriptworker/cot/verify.py
```diff
@@ -960,111 +916,9 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
-        "cd /builds/worker/checkouts/gecko &&
-        ln -s /builds/worker/artifacts artifacts &&
-        ./mach --log-no-times taskgraph decision --pushlog-id='0' --pushdate='0' --project='date'
-        --message='try: -b o -p foo -u none -t none' --owner='amiyaguchi@mozilla.com' --level='2'
-        --base-repository=$GECKO_BASE_REPOSITORY --head-repository=$GECKO_HEAD_REPOSITORY
-        --head-ref=$GECKO_HEAD_REF --head-rev=$GECKO_HEAD_REV --revision-hash=$GECKO_HEAD_REV
-        --triggered-by='nightly' --target-tasks-method='nightly_linux'\n"
-
-    This is very firefox-centric and potentially fragile, but decision tasks are
-    important enough to need to monitor.  The ideal fix would maybe be to simplify
```
tomprince: What has changed that we don't care about this anymore?

escapewindow: This was one of the fragile and weakest parts of CoT v1, looking at the decision task and trying to spot check it to make sure that it looked vaguely like a decision task. It's complex, fragile, and doesn't actually protect us from all decision task changes. CoT v2 added json-e rebuilding of the decision task, so we verify the entirety of the decision task comes from `.taskcluster.yml`, not just the command. [bug 1395413](https://bugzilla.mozilla.org/show_bug.cgi?id=1395413) also describes some reasons why this command check is not ideal.

This was another part of CoT v1 that we were very interested in sunsetting when everything supported CoT v2.

scriptworker/cot/verify.py
```diff
@@ -960,111 +916,9 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
         verify_task_in_task_graph(task_link, graph_defn)
         log.info("Found {} in the graph; it's a match".format(task_link.task_id))
         return
-    # Fall back to fuzzy matching to support retriggers: the taskId and
```
JohanLorenzo: I'm not sure to follow how retriggers work in this model. Does the retrigger action task creates a new task-graph.json and CoT only relies on it?

scriptworker/cot/verify.py
```diff
@@ -1228,41 +1082,101 @@ def verify_decision_command(decision_link, mach_commands=DECISION_MACH_COMMANDS)
...
         decision_link (LinkOfTrust): the parent link's decision task link.
         tasks_for (str): the reason the parent link was created (cron,
             hg-push, action)
 
     Returns:
-        dict: the json-e template.
+        (dict, dict): the json-e context and template.
 
     """
     if tasks_for == 'action':
```
JohanLorenzo: I'm wondering, where do we safeguard against `tasks_for` not being in `('action', 'decision')`?

scriptworker/cot/verify.py
```diff
@@ -1643,12 +1542,6 @@ def check_num_tasks(chain, task_count):
         CoTError: on failure.
 
     """
-    if chain.context.config['cot_product'] == 'mobile':
```
JohanLorenzo: This removal looks good to me! :+1:

scriptworker/task.py
```diff
@@ -240,6 +243,25 @@ def get_and_check_tasks_for(task, msg_prefix=''):
...
+        task (dict): the task definition.
+
+    Returns:
+        str: the ``repo_scope``
+        None: if no ``repo_scope`` is found
+
+    """
+    for scope in task['scopes']:
+        if REPO_SCOPE_REGEX.match(scope):
+            return scope
```
JohanLorenzo: I'm afraid this might cover an issue. Let's imaging we have these 2 scopes:
```
assume:repo:some-repo:action:action1
assume:repo:some-repo:action:action2 
```

The first repo scope will always be returned, even though we might want the `action2` scope. Am I getting this right?

scriptworker/cot/verify.py
escapewindow: [`get_and_check_tasks_for`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L216-L240) throws if `tasks_for` isn't in [`KNOWN_TASKS_FOR`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L28) .

scriptworker/cot/verify.py
escapewindow: That's the theory, and it works for relpro tasks, which create `task-graph.json` with an accurate `taskId` and task definition.

Retriggers seem to create a `task-graph-0.json`, which isn't supported by CoT verification... the reasoning seems to be that retriggers allow for multiple graphs (e.g. retrigger graph 0-5). This means CoT verification will fail.

However, fuzzy task matching only worked for a single task on current retriggers. Any subsequent tasks failed, e.g. a retriggered signing task might pass CoT verification, but a child beetmover task would fail. Fuzzy task matching adds time, log spam, and doesn't fully solve the problem, so we're leaning towards saying "retriggers aren't compatible with CoT", or "retriggers should be for tests only".

scriptworker/task.py
escapewindow: That's correct. If the correct `repo_scope` is actually `assume:repo:some-repo:action:action2`, then the json-e render will create a mismatched task definition, and CoT verification will fail.

At that point we'll hopefully realize that having two `assume:repo:some-repo:action:*` scopes in the same task is a bad idea, and remove one.

escapewindow: Would you prefer we die on multiple `repo_scope`s ?

We're getting rid of `repo_scope` at some point (I think when all actions are hooks), so I don't want to require `repo_scope` not be `None`.

JohanLorenzo: Yeah, dying on multiple scopes sounds safer too me. That would immediately shows something broke.

escapewindow: Ok, added that check.

escapewindow:
https://github.com/lemurheavy/coveralls-public/issues/1128
I think I'm done poking at this PR for a while...

scriptworker/cot/verify.py
escapewindow: Sure. I could change this to `^` instead of `-`, or update the comment. I don't see how someone can a) keep the same taskId, b) remove some dependencies so the task has a subset of graph dependencies, and c) cause us harm. (a) and (b) seem difficult or impossible to do via the taskcluster APIs. At worst I can think of causing downstream tasks to run too early for (c).

PR #224, opened 2018-05-29
changed traceback.print_exc() calls to log.exception()

lisajguo:
so that errors don't print to stdout


escapewindow:
Filed #225 for the test bustage...

escapewindow:
Aha! Flake8 is complaining.

```
./scriptworker/gpg.py:28:1: F401 'traceback' imported but unused
import traceback
^
```

Could you fix that @lisajguo ?

PR #227, opened 2018-05-30
fix mobile prebuilt_docker_image_task_types

escapewindow:
We want to allow any mobile prebuilt_docker_image_task_types, so I added
a special case for `None`. However, we also have a check in
`scriptworker.config` to explicitly disallow `None` values. To fix, I
special-cased the string `any` instead.


escapewindow:
Hm, getting `/builds/scriptworker/scriptworker.yaml prebuilt_docker_image_task_types: type <class 'str'> is not <class 'tuple'>!`

escapewindow:
mobil-* are currently pinned to my environment til we release this.

Issue #128, opened 2017-07-11
exceeding `task_max_timeout` appears to crash scriptworker


escapewindow:
```
2017-07-11T18:56:52    DEBUG - waiting 300 seconds before reclaiming...
2017-07-11T19:00:06  WARNING - Exceeded timeout of 1800 seconds: 20167
Exception in callback max_timeout(<scriptworker...x7fdbd97aa5f8>, <Process 20167>, 1800) at /builds/scriptworker/lib/python3.5/site-packages/scriptworker/task.py:234
handle: <TimerHandle when=4875.318810819001 max_timeout(<scriptworker...x7fdbd97aa5f8>, <Process 20167>, 1800) at /builds/scriptworker/lib/python3.5/site-packages/scriptworker/task.py:234>
Traceback (most recent call last):
  File "/tools/python35/lib/python3.5/asyncio/events.py", line 125, in _run
    self._callback(*self._args)
  File "/builds/scriptworker/lib/python3.5/site-packages/scriptworker/task.py", line 256, in max_timeout
    asyncio.ensure_future(kill(pid))
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 375, in run_until_complete
    self.run_forever()
  File "/tools/python35/lib/python3.5/asyncio/base_events.py", line 340, in run_forever
    raise RuntimeError('Event loop is running.')
RuntimeError: Event loop is running.
2017-07-11T19:00:22     INFO - exit code: -15
2017-07-11T19:00:37    DEBUG - "/builds/scriptworker/artifacts/public/chainOfTrust.json.asc" is encoded with "None" and has mime/type "text/plain"
2017-07-11T19:00:37     INFO - "/builds/scriptworker/artifacts/public/chainOfTrust.json.asc" can be gzip'd. Compressing...
2017-07-11T19:00:38     INFO - "/builds/scriptworker/artifacts/public/chainOfTrust.json.asc" compressed
2017-07-11T19:00:38    DEBUG - "/builds/scriptworker/artifacts/public/logs/chain_of_trust.log" is encoded with "None" and has mime/type "text/plain"
2017-07-11T19:00:38     INFO - "/builds/scriptworker/artifacts/public/logs/chain_of_trust.log" can be gzip'd. Compressing...
2017-07-11T19:00:38     INFO - "/builds/scriptworker/artifacts/public/logs/chain_of_trust.log" compressed
[...]

PR #230, opened 2018-05-31
address event loop bustage in concurrent tests

### LINKED ISSUE #225: concurrent test intermittent bustage
We get intermittent bustage with `RuntimeError: There is no current event loop in thread 'Dummy-1'.`

I'm guessing this is due to something killing or closing the event loop, breaking other concurrent tests that are running simultaneously.

We need to track this down and fix.

escapewindow:
attempt to fix #225. This was fine when we were running tests synchronously, but mocking `get_event_loop` when we're running a number of tests concurrently seems potentially dangerous.

### DIFF
--- a/scriptworker/test/test_client.py
+++ b/scriptworker/test/test_client.py
@@ -211,11 +211,21 @@ def test_bad_artifact_url(valid_artifact_rules, valid_artifact_task_ids, url):
 def test_sync_main_runs_fully(config, event_loop, should_validate_task):
     copyfile(BASIC_TASK, os.path.join(config['work_dir'], 'task.json'))
     generator = (n for n in range(0, 2))
+    count = []
 
     async def async_main(_):
         next(generator)
 
-    kwargs = {}
+    def counter(arg1):
+        count.append(arg1)
+
+    fake_loop = MagicMock()
+    fake_loop.run_until_complete = counter
+
+    def loop_function():
+        return fake_loop
+
+    kwargs = {'loop_function': loop_function}
 
     if should_validate_task:
         schema_path = os.path.join(config['work_dir'], 'schema.json')
@@ -232,7 +242,7 @@ def test_sync_main_runs_fully(config, event_loop, should_validate_task):
         kwargs['config_path'] = f.name
         client.sync_main(async_main, **kwargs)
 
-    assert next(generator) == 1 # async_main was called once
+    assert len(count) == 1  # run_until_complete was called once
 
 
 @pytest.mark.parametrize('does_use_argv, default_config', (
--- a/scriptworker/test/test_client.py
+++ b/scriptworker/test/test_client.py
@@ -210,17 +210,17 @@ def test_bad_artifact_url(valid_artifact_rules, valid_artifact_task_ids, url):
 @pytest.mark.parametrize('should_validate_task', (True, False))
 def test_sync_main_runs_fully(config, event_loop, should_validate_task):
     copyfile(BASIC_TASK, os.path.join(config['work_dir'], 'task.json'))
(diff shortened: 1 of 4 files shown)


escapewindow:
Doh, the 4th set of tests failed. Something's still broken with the event loop & concurrent tests.

escapewindow:
I've run tests locally ~5 times and in Travis ~4 times, so I'm reasonably confident this set of changes improves the intermittent concurrent test bustage.

I'm getting ambitious and trying to track down the unclosed session warnings.

escapewindow:
Ok, those attempts ran into `task_max_timeout` integration intermittent issues, which I think we can address via https://github.com/mozilla-releng/scriptworker/issues/128#issuecomment-393739337 .

Rather than scope creep this PR, let's get this reviewed. @JohanLorenzo are you up for that?

scriptworker/test/test_artifacts.py
```diff
@@ -253,7 +253,8 @@ def test_get_upstream_artifacts_full_paths_per_task_id(context):
     }
 
 
-def test_fail_get_upstream_artifacts_full_paths_per_task_id(context):
+@pytest.mark.asyncio
+async def test_fail_get_upstream_artifacts_full_paths_per_task_id(context):
```
JohanLorenzo: I'm a little bit confused on why this function must be declared async, but not `test_get_and_check_single_upstream_artifact_full_path()`. They are both testing the same function and they both rely on the context. In the function under test, the context doesn't require to be async.

What's the reason for flipping this particular function?

scriptworker/test/test_client.py
```diff
@@ -232,7 +243,10 @@ def test_sync_main_runs_fully(config, event_loop, should_validate_task):
         kwargs['config_path'] = f.name
         client.sync_main(async_main, **kwargs)
 
-    assert next(generator) == 1 # async_main was called once
+    for i in count:
+        await i  # suppress coroutine not awaited warning
+    assert len(count) == 1  # run_until_complete was called once
+    assert next(generator) == 1  # async_main was called once
```
JohanLorenzo: Nit: `count` and `generator` seem to implement the same intent (counting the number of times a function was called) but:
* they do it on 2 different functions
* they're implemented differently.

In order to make the reading easy, I'd suggest to append the mocked function name on the variable name (eg: `count_on_run_until_complete`).
As a matter of consistency, let's make these 2 counters implemented the same way. I have no preference whether to use a generator or a list. 

What do you think?

scriptworker/test/test_artifacts.py
escapewindow: I wasn't sure what exactly was breaking - The error was that the event loop wasn't available, and this was one of the tests that showed the error. This may be because the loop wasn't available in this test, or the warning may be output from another test, but appear next to this test since the tests are run concurrently. I agree that it doesn't make sense that this test would require asyncio but other similar tests don't.

scriptworker/test/test_client.py
escapewindow: Sure. I'm going to use a list because I need to await the saved coroutines.

scriptworker/test/test_artifacts.py
escapewindow: Going to try removing these `@pytest.mark.asyncio`'s and see if that changes anything.

escapewindow:
Testing locally after dropping https://github.com/mozilla-releng/scriptworker/pull/230/commits/f0c9ac2c01888d181b5b5e63d2b35a0bbd581b8d (marking certain tests as asyncio).

escapewindow:
Going to land this to fix #225. I'm going to file a new issue to fix the unclosed session warnings.

PR #232, opened 2018-06-01
bug 1400258 - write cot verification logs to live_backing.log

### LINKED ISSUE #179: cot verification should write to `live_backing.log`
Right now, we write our cot log to `public/logs/chain_of_trust.log`. If this passes, then we start writing `public/logs/live_backing.log` using the task output. This works in the standard case, where cot verification passes.

When cot fails, treeherder doesn't show the cot failure, because it doesn't know about `chain_of_trust.log`. We could solve this by teaching treeherder about this log, or we could solve by:

- writing the chain of trust log to `public/logs/live_backing.log`, and EITHER

- moving the chain of trust log to `public/logs/chain_of_trust.log` on success, OR
- appending the task log to the chain of trust log at `public/logs/live_backing.log`, depending on whether we prefer 2 different log files or not.

escapewindow:
Fixes #179.

This writes the chain of trust verification log to `live_backing.log`, then appends the task log. I could change this to write cot verification to `live_backing.log`, then move it to `chain_of_trust.log` on success, if we prefer not to see this output in the full log.

I'm thinking we should clean up/clarify the cot log output anyway, so maybe more visibility is better.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1795,7 +1795,7 @@ async def verify_chain_of_trust(chain):
         CoTError: on failure
 
     """
-    log_path = os.path.join(chain.context.config["task_log_dir"], "chain_of_trust.log")
+    log_path = os.path.join(chain.context.config["task_log_dir"], "live_backing.log")
     scriptworker_log = logging.getLogger('scriptworker')
     with contextual_log_handler(
         chain.context, path=log_path, log_obj=scriptworker_log,
--- a/scriptworker/log.py
+++ b/scriptworker/log.py
@@ -112,7 +112,7 @@ def get_log_filehandle(context):
     """
     log_file_name = get_log_filename(context)
     makedirs(context.config['task_log_dir'])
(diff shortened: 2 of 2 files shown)


escapewindow:
This PR is on hold; @tomprince is working on bug 1400258.

tomprince:
The solution I had thought of is outlined [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1400258), but isn't as straightforward as I had hoped. I'm not actively working on this, but the first option I outlined there would be easy to implement, if that is the thought to be the best solution.

escapewindow:
On June 1, I noticed that the chain_of_trust.log not showing up in treeherder's logs was causing the sheriffs some headaches. I wrote this patch since this solution seemed the fastest to implement; other than that I don't have strong opinions about which logfile the cot verification goes into.

I'm not sure if the sheriffs are still being negatively affected by the cot log not showing up in treeherder. (It's probably only really visible when cot verification fails.) We could land this and be done, or just leave it as is with until one of us solves it another way.

Issue #197, opened 2018-03-14
randomize test order


escapewindow:
Using http://pythonhosted.org/pytest-random-order/ ... I think just adding this to tox.ini deps will do the trick.

Issue #167, opened 2017-10-17
per-scriptworker exit code config


escapewindow:
In [this commit](https://github.com/mozilla-releng/scriptworker/commit/513fe573feb5eb4b6f9c4780b44586a057d85ebc), I allowed for an exit code of -11 to cause the task to retry, via the `intermittent-task` taskcluster status. This is global across all scriptworker instance types, even if we spin up a scriptworker instance that doesn't use python.

We should add a config dict that we use to update the `REVERSED_STATUSES`, so we can set the various exit code mapping per scriptworker instance type. That probably means `REVERSED_STATUSES` (and possibly `STATUSES`) will move into config.

(`STATUSES` could actually have a tuple of exit codes as values instead of an int... not decided about this.)

We could also ad a `default` value. Right now if we hit an unknown exit code, we specify failure. If we had a per-scriptworker-instance-type default value, we could tweak behavior on unknown exit codes.

Issue #131, opened 2017-08-11
scriptworker should report `worker-shutdown` on shutdown


escapewindow:
if it has claimed a task.

https://docs.taskcluster.net/reference/platform/taskcluster-queue/references/api#report-task-exception

This will send the task back to retry sooner, and won't be confusing like `claim-expired`, which could mean anything.

Related: if there's a way to graceful-shutdown a worker, we should use it to restart scriptworker rather than killing/restarting in puppet.

escapewindow:
@mitchhentges fixed this.

PR #236, opened 2018-06-26
Add user-cancel support

### LINKED ISSUE #42: add user cancel support to scriptworker
Currently, we poll to reclaimTask every 5 min.
I believe when we do that, we'll be notified if the status of the task changes.
If the status is user-cancel, we should kill the task.
If we also change the polling interval to something smaller (30 seconds or less?) we should cancel the script in a relatively responsive period of time.

### LINKED ISSUE #128: exceeding `task_max_timeout` appears to crash scriptworker

escapewindow:
Fixes #42. Also fixes #128, #167, #197.

I cleaned up the patchset and split out the patches into logical chunks; hopefully this makes things easier to review.

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -5,12 +5,22 @@ Attributes:
     DEFAULT_CONFIG (frozendict): the default config for scriptworker.  Running configs
         are validated against this.
     STATUSES (dict): maps taskcluster status (string) to exit code (int).
-    REVERSED_STATUSES (dict): the same as STATUSES, except it maps the exit code
-        (int) to the taskcluster status (string).
 
 """
 from frozendict import frozendict
 import os
+
+# STATUSES {{{1
+STATUSES = {
+    'success': 0,
+    'failure': 1,
+    'worker-shutdown': 2,
+    'malformed-payload': 3,
+    'resource-unavailable': 4,
+    'internal-error': 5,
+    'superseded': 6,
+    'intermittent-task': 7,
+}
 
 # DEFAULT_CONFIG {{{1
 # When making changes to DEFAULT_CONFIG that may be of interest to scriptworker
@@ -43,6 +53,16 @@ DEFAULT_CONFIG = frozendict({
     "reclaim_interval": 300,
     "poll_interval": 10,
     "sign_key_timeout": 60 * 2,
+
+    "reversed_statuses": frozendict({
+        -11: STATUSES['intermittent-task'],
+        -15: STATUSES['intermittent-task'],
+    }),
+    # Report this status on max_timeout. `intermittent-task` will rerun the
+    # task automatically. `internal-error` or other will require manual
+    # intervention.
+    "task_max_timeout_status": STATUSES['intermittent-task'],
+    "invalid_reclaim_status": STATUSES['intermittent-task'],
 
     "task_script": ("bash", "-c", "echo foo && sleep 19 && exit 1"),
 
@@ -362,18 +382,18 @@ DEFAULT_CONFIG = frozendict({
     },
 })
 
-# STATUSES and REVERSED_STATUSES {{{1
-STATUSES = {
-    'success': 0,
-    'failure': 1,
-    'worker-shutdown': 2,
-    'malformed-payload': 3,
-    'resource-unavailable': 4,
-    'internal-error': 5,
-    'superseded': 6,
-    'intermittent-task': 7,
-}
-# Retry on python segfault
-_rev = {v: k for k, v in STATUSES.items()}
-_rev[-11] = 'intermittent-task'
-REVERSED_STATUSES = _rev
+
+# get_reversed_statuses {{{1
+def get_reversed_statuses(context):
+    """Return a mapping of exit codes to status strings.
+
+    Args:
+        context (scriptworker.context.Context): the scriptworker context
+
+    Returns:
+        dict: the mapping of exit codes to status strings.
+
+    """
+    _rev = {v: k for k, v in STATUSES.items()}
+    _rev.update(dict(context.config['reversed_statuses']))
+    return _rev
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -21,7 +21,7 @@ from urllib.parse import unquote, urlparse
 import taskcluster
 import taskcluster.exceptions
 
-from scriptworker.constants import REVERSED_STATUSES
+from scriptworker.constants import get_reversed_statuses
 from scriptworker.log import get_log_filehandle, pipe_to_log
 from scriptworker.utils import match_url_path_callback, match_url_regex
 
@@ -470,12 +470,13 @@ async def complete_task(context, result):
 
     """
     args = [get_task_id(context.claim_task), get_run_id(context.claim_task)]
+    reversed_statuses = get_reversed_statuses(context)
     try:
         if result == 0:
             log.info("Reporting task complete...")
             response = await context.temp_queue.reportCompleted(*args)
-        elif result != 1 and result in REVERSED_STATUSES:
-            reason = REVERSED_STATUSES[result]
+        elif result != 1 and result in reversed_statuses:
+            reason = reversed_statuses[result]
             log.info("Reporting task exception {}...".format(reason))
             payload = {"reason": reason}
             response = await context.temp_queue.reportException(*args, payload)
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -417,7 +417,7 @@ async def run_task(context):
         await asyncio.wait(tasks)
         exitcode = await context.proc.wait()
         status_line = "exit code: {}".format(exitcode)
-        if exitcode == -11:
+        if exitcode < 0:
             status_line = "Automation Error: python exited with signal {}".format(exitcode)
         log.info(status_line)
         print(status_line, file=log_filehandle)
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -22,8 +22,13 @@ import taskcluster
 import taskcluster.exceptions
 
 from scriptworker.constants import get_reversed_statuses
+from scriptworker.exceptions import ScriptWorkerTaskException
 from scriptworker.log import get_log_filehandle, pipe_to_log
-from scriptworker.utils import match_url_path_callback, match_url_regex
+from scriptworker.utils import (
+    get_future_exception,
+    match_url_path_callback,
+    match_url_regex,
+)
 
 log = logging.getLogger(__name__)
 
@@ -387,7 +392,6 @@ async def run_task(context):
         int: exit code
 
     """
-    loop = asyncio.get_event_loop()
     kwargs = {  # pragma: no branch
         'stdout': PIPE,
         'stderr': PIPE,
@@ -396,12 +400,20 @@ async def run_task(context):
         'preexec_fn': lambda: os.setsid(),
     }
     context.proc = await asyncio.create_subprocess_exec(*context.config['task_script'], **kwargs)
-    loop.call_later(context.config['task_max_timeout'], max_timeout, context, context.proc, context.config['task_max_timeout'])
 
     tasks = []
+    timeout_exc = None
     with get_log_filehandle(context) as log_filehandle:
-        tasks.append(pipe_to_log(context.proc.stderr, filehandles=[log_filehandle]))
-        tasks.append(pipe_to_log(context.proc.stdout, filehandles=[log_filehandle]))
+        tasks.append(asyncio.ensure_future(
+            pipe_to_log(context.proc.stderr, filehandles=[log_filehandle])
+        ))
+        tasks.append(asyncio.ensure_future(
+            pipe_to_log(context.proc.stdout, filehandles=[log_filehandle])
+        ))
+        timeout_task = asyncio.ensure_future(
+            max_timeout(context, context.proc, context.config['task_max_timeout'])
+        )
+        tasks.append(timeout_task)
         await asyncio.wait(tasks)
         exitcode = await context.proc.wait()
         status_line = "exit code: {}".format(exitcode)
@@ -409,8 +421,13 @@ async def run_task(context):
             status_line = "Automation Error: python exited with signal {}".format(exitcode)
         log.info(status_line)
         print(status_line, file=log_filehandle)
+        timeout_exc = get_future_exception(timeout_task)
+        for task in tasks:
+            task.cancel()
 
     context.proc = None
+    if timeout_exc:
+        raise timeout_exc
     return exitcode
 
 
@@ -448,6 +465,12 @@ async def reclaim_task(context, task):
         except taskcluster.exceptions.TaskclusterRestFailure as exc:
             if exc.status_code == 409:
                 log.debug("409: not reclaiming task.")
+                if context.proc and task == context.task:
+                    await kill_task(
+                        context,
+                        "Killing task after receiving 409 status in reclaim_task",
+                        context.config['invalid_reclaim_status']
+                    )
                 break
             else:
                 raise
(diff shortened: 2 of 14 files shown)


escapewindow:
@JohanLorenzo , @tomprince do you have headspace for this review?

scriptworker/utils.py
```diff
@@ -355,19 +354,40 @@ def create_temp_creds(client_id, access_token, start=None, expires=None,
     if tasks:
         await asyncio.wait(tasks)
         for task in tasks:
-            exc = task.exception()
-            if exc is None:
-                succeeded_results.append(task.result())
-            else:
+            exc = get_future_exception(task)
+            if exc:
```
tomprince: Given that the new function returns both `False` and `None` with different meanings, I think we should handle both cases explicitly here. From the comment below, it sounds like you don't expect to get `False` in this context, it might be reasonable to just log in that case, although maybe bailing out is better if we hit an impossible situation?

scriptworker/task.py
```diff
@@ -396,21 +400,34 @@ def prepare_to_run_task(context, claim_task):
...
+            pipe_to_log(context.proc.stdout, filehandles=[log_filehandle])
+        ))
+        timeout_task = asyncio.ensure_future(
+            max_timeout(context, context.proc, context.config['task_max_timeout'])
+        )
+        tasks.append(timeout_task)
         await asyncio.wait(tasks)
         exitcode = await context.proc.wait()
         status_line = "exit code: {}".format(exitcode)
         if exitcode == -11:
```
tomprince: This probably should be `exitcode < 0`.

scriptworker/task.py
```diff
@@ -491,8 +514,35 @@ def prepare_to_run_task(context, claim_task):
...
+
+    Returns:
+        None, if the task finishes before the timeout.
+
+    """
+    await asyncio.sleep(timeout)
+    if proc is not context.proc:
+        return
+    await kill_task(
+        context,
```
tomprince: Would it make more sense to pass `proc` here, so that we know that we are killing the process that triggered the timeout? I'm not entirely familiar with the control flow here, but the above check suggests this coroutine could outlive a task, in which case there is going to be a race condition of where the check happens and then `kill_task` starts running, and the `context.proc` gets changed under it.

Thinking about it more, I think this could also use a comment about why we aren't just always killing `proc`, even if it is no longer the current process.

scriptworker/utils.py
escapewindow: We've already waited on all the tasks above, which I believe means they've all completed or thrown exceptions. Since `False` means the future hasn't finished yet, I don't expect to hit that, and adding another if block is another thing I need to cover in tests -- I suppose we could reach that block by mocking enough things, but at the time of writing I thought it would make more sense to leave it off.

Given the above, do you still need to check for `False` ?

scriptworker/task.py
escapewindow: Sure, we could pass `proc`, esp since we don't use `context` anywhere else in `kill_task`.
And a comment makes sense.

scriptworker/worker.py
```diff
@@ -57,29 +124,16 @@
     for task_defn in tasks.get('tasks', []):
         status = 0
         prepare_to_run_task(context, task_defn)
-        loop.create_task(reclaim_task(context, context.task))
```
JohanLorenzo: Nice cleanup! :+1:

scriptworker/constants.py
```diff
@@ -44,6 +54,16 @@
     "poll_interval": 10,
     "sign_key_timeout": 60 * 2,
 
+    "reversed_statuses": frozendict({
+        -11: STATUSES['intermittent-task'],
+        -15: STATUSES['intermittent-task'],
```
JohanLorenzo: I'm confused about these values. How different are -11 and -15?

tox.ini
```diff
@@ -24,6 +24,7 @@ deps =
     pytest-asyncio
     pytest-cov
     pytest-mock
+    pytest-random-order
```
JohanLorenzo: :+1:

scriptworker/task.py
tomprince: Well, all negative exit codes represent the process exiting with a signal, so we should log all of them like that. We may also want to do interesting stuff with things in `reverse_statues`.

scriptworker/utils.py
tomprince: I have two thoughts on this (which are really two different ways of looking at the same thing:
1. As a reader, when I look at this code, I see the `get_future_exception`, see that it has three possible return values and then am wondering why no result result is being treated as success and wondering if it was mistake because two of the return values happen to be false-y (looking at it through a rust lens, I'd expect there two be three case in handling the return value). Given that the you expect the `False` case never to happen, I'd be fine with that branch being marked `pragma: nocover`.
2. Looking at `get_future_exception`, it looks like a fairly trivial wrapper around `future.exception()`. It looks like it exists to turn a `InvalidStateError` into a `False` return value. As a reader, I expect that there is going to be somewhere where this case is not exception and you want to handle all three case in parallel ways. But when I go look at the only caller, it appears that you don't actually handle the `False` case because it is shouldn't ever happen. That suggests that just using `future.exception()` would be better here, and let the impossible exception bubble up, rather than be silently treated as a success.

scriptworker/test/test_task.py
```diff
@@ -342,39 +365,73 @@ def test_reclaim_task_non_409(context, successful_queue, event_loop):
         )
 
 
+@pytest.mark.parametrize("proc", (None, 1))
 @pytest.mark.asyncio
-async def test_reclaim_task_mock(context, mocker, event_loop):
+async def test_reclaim_task_mock(context, mocker, proc):
+
```
tomprince: It'd be somewhat easier to review the tests if the had an explanation of what they were trying to test. (I've often pointed people [here](https://jml.io/pages/test-docstrings.html)).

scriptworker/test/test_task.py
```diff
@@ -342,39 +365,73 @@ def test_reclaim_task_non_409(context, successful_queue, event_loop):
...
+    mocker.patch.object(swtask, 'kill_task', new=fake_kill)
+    await swtask.max_timeout(context, "invalid_proc", 0)
+    assert called == []
+
+
+@pytest.mark.asyncio
+async def test_max_timeout_mock(context, mocker):
+
+
```
tomprince: It looks like some spurious blank lines got added to the beginning of a few of these methods.

scriptworker/test/test_worker.py
```diff
@@ -163,15 +164,22 @@ def test_mocker_run_tasks_noop(context, successful_queue, event_loop, mocker):
...
+    'upload_artifacts', OSError, None, True
 )))
-def test_mocker_run_tasks_exception(context, successful_queue, event_loop,
-                                   mocker, func_to_raise, exc, expected):
+@pytest.mark.asyncio
+async def test_mocker_run_tasks_exception(context, successful_queue, mocker,
+                                          func_to_raise, exc, expected, raises):
     """Raise an exception within the run_tasks try/excepts and make sure the
     status is changed
     """
```
tomprince: Looking at this, I'm not sure whether the code or tests is correct. In the unexpected exception case, it appears that we don't change the status.
- If the code is correct, we need to adjust the documentation to reflect the new behavior.
- However, it appears that an unexpected exception will cause the process to exit (good!) but also leave the task claimed (meaning that it will need to timeout the claim). That isn't unreasonable behavior, but we should probably call it out somewhere.
This probably wants to be adjusted to account for unexpected exceptions

scriptworker/test/test_integration.py
```diff
@@ -172,30 +171,71 @@ def get_temp_creds(context):
...
     }
-    pre = arrow.utcnow()
-    try:
-        event_loop.run_until_complete(maxtimeout_helper(context_function, partial_config, task_id))
-    except RuntimeError:
-        pass
-    post = arrow.utcnow()
-    # This may be flaky because claimWork can take up to 20 seconds..?
-    assert post.timestamp - pre.timestamp <= 8
+    # No temp_creds context because we won't have cancel scopes
```
tomprince: Could you expand on this comment? I'm not really sure what it means.

scriptworker/constants.py
escapewindow: We were getting -11 for segfaults, and we get -15 when we kill the task.

scriptworker/test/test_integration.py
```diff
@@ -172,30 +171,71 @@ def get_temp_creds(context):
...
+        assert status['status']['state'] == 'exception'
+        assert status['status']['runs'][0]['reasonResolved'] == 'canceled'
+        log_url = context.queue.buildUrl(
+            'getLatestArtifact', task_id, 'public/logs/live_backing.log'
+        )
+        log_path = os.path.join(context.config['work_dir'], 'log')
+        await utils.download_file(context, log_url, log_path)
+        with open(log_path) as fh:
+            contents = fh.read()
+        assert contents.rstrip() == "bar\nfoo\nexit code: -15"
```
tomprince: This is fairly dense code. I think it looks correct, but it could do with some comments, or at the least, breaking up into sections (setup, run, verify at a minimum).

scriptworker/task.py
```diff
@@ -448,6 +465,12 @@ def prepare_to_run_task(context, claim_task):
         except taskcluster.exceptions.TaskclusterRestFailure as exc:
             if exc.status_code == 409:
                 log.debug("409: not reclaiming task.")
+                if context.proc and task == context.task:
+                    await kill_task(
+                        context,
+                        "Killing task after receiving 409 status in reclaim_task",
```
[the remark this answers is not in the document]
escapewindow: 409 means we can't reclaim the task because it's no longer ours to reclaim. Aiui, that includes

- superseded (taken by another worker)
- cancelled
- deadline-exceeded
- marked as completed/exception somehow (task finished in another async coroutine?)

scriptworker/worker.py
escapewindow: Thanks!

scriptworker/utils.py
escapewindow: We also use `get_future_exception` in `run_task`: `timeout_exc = get_future_exception(timeout_task)`. I can certainly use `future.exception()` here, and only use `get_future_exception` in the one place.

scriptworker/task.py
escapewindow: Sure. This was in reaction to a specific issue where we were hitting -11 and wanting that in the logs, but I don't see any harm with including all negative statuses.

scriptworker/task.py
```diff
@@ -490,8 +514,38 @@ def prepare_to_run_task(context, claim_task):
...
+
+    Returns:
+        None, if the task finishes before the timeout.
+
+    """
+    await asyncio.sleep(timeout)
+    if proc is not context.proc:
+        # Since this coroutine is run in parallel to the task, it's possible
+        # the running task is no longer the task we were assigned to monitor
+        # for timeout. If `context.proc` doesn't match `proc`, bail.
```
tomprince: Thanks. This is much clearer. Now that we won't hit the next task, is there a reason we need to bail? I can see an argument about wanting to avoid hitting something due to PID reuse. I'd be happy for this to be captured in an issue and postponed.

escapewindow: Now that I'm canceling the timeout task at the end of the loop, I think we might be ok not bailing. However,

- yes, we might kill a reused pid,
- we're executing code that shouldn't be needed, slowing things unnecessarily
- the timeout cancel is new and untested, so it's possible it's fallible

It might make sense, though.

escapewindow: Oh, and we might kill the existing task if we raise an exception mid-run. That might not happen if we don't retrieve the future's exception, just another piece in the not-sure column.

scriptworker/test/test_worker.py
escapewindow: Yeah, I was thinking about fixing that this PR, and decided to leave it for #131. I do log a special string for the unexpected exception, so we can search for them in papertrail. Does that work?

scriptworker/test/test_integration.py
tomprince: *Why* don't they have cancel scopes?

escapewindow: That's a question for the tc team. The temp creds are returned during the first `claimTask` call, and contain just enough scopes to run the task. That doesn't include cancel scopes. I believe that question is outside the scope of this PR.

escapewindow: `claimTask` scopes docs at https://docs.taskcluster.net/docs/reference/platform/taskcluster-queue/references/api#claimTask

escapewindow: I suppose we could add the cancel task scope in the `task.scopes`... not sure this comment is that noteworthy, but if you have suggestions, I can fix.

scriptworker/test/test_worker.py
tomprince: I'm happy to let this be fixed in a separate PR, but then I think the new behavior we are codifying in a test here we think is wrong and are going to correct in a latter PR.

I'm not sure what the solution here is, but I don't think there is value having tests that ossify incorrect behavior.

scriptworker/test/test_integration.py
tomprince: I think maybe something like
> Don't use task-specific temporary credentials, since they don't allow us to cancel the created task.

or
> Don't use temporary credentials from `claimTask`, since they don't allow us to cancel the created task.

is clearer, since it gives some indication of where the temporary credentials are coming from, that we don't have control over them.  I was confused because I was assuming that we were somehow creating temporary credentials in the test.

scriptworker/test/test_worker.py
escapewindow: I think the behavior is as designed, with plans to change the behavior later.
I updated the docstring, since I think that's what you're commenting on.

escapewindow:
Filed #238 for the travis intermittent errors.

scriptworker/test/test_task.py
tomprince: Something like

> When `queue.reclaim_task` raises an error with status 409, `reclaim_task` returns. If there is a running process, `reclaim_task` tries to kill it before returning.

Reading that docstring, suggests to me that `fake_reclaim` should actually be the function that is raising the exception, rather than monkey-patching `pformat` to do that.

escapewindow:
Ok, I think https://github.com/mozilla-releng/scriptworker/pull/236/commits/ed7849662adef02f132722bbbd65e26af0572e10 should address the issues we discussed on vidyo.

PR #237, opened 2018-06-26
scl3 -> mdc1

escapewindow:


MihaiTabara:
Not sure if this is realted to version changes. I see `json-e` went from 2.5.0 to 2.6.0, `ptyprocess` from 0.5.2 to 0.6.0, `pytest` from 3.6.1 to 3.6.2 and `PyYAML` from 3.12 to 4.1. Except latter, all of them are PATCH and MINOR new changes so I doubt that's the culprit.

escapewindow:
I don't love the fact that this was an intermittent, but it's green now.

PR #240, opened 2018-07-03
Render the full json-e template rather than `tasks[0]`

### LINKED ISSUE #239: Chain of trust shoud extract the tasks from `.taskcluster.yml` after rendering, rather than before.
That would avoid issues like https://bugzilla.mozilla.org/show_bug.cgi?id=1470886#c9 where changing the template in a way that doesn't affect the rendered definition caused scriptworker verification to fail.

escapewindow:
We currently hardcode the assumption that a) the tasks are listed in
.taskcluster.yml (this assumption changes with bug 1470886), and b) the
decision task is the first task in the template.

Instead, we now render the full template, and then verify that the
runtime task definition is part of the `rendered_template['tasks']`.

This should fix #239.


scriptworker/cot/verify.py
```diff
@@ -1294,18 +1292,17 @@ def compare_jsone_task_definition(parent_link, compare_definition):
         compare_definition = remove_empty_keys(compare_definition)
         runtime_definition = remove_empty_keys(parent_link.task)
 
-        log.debug("Compare_definition:\n{}".format(format_json(compare_definition)))
-        log.debug("Runtime definition:\n{}".format(format_json(runtime_definition)))
         diff = list(dictdiffer.diff(compare_definition, runtime_definition))
         if diff:
-            raise AssertionError(pprint.pformat(diff))
+            error_msg = "{} {}: the rebuilt definition doesn't match the runtime task!\n{}".format(
```
tomprince: Should this be a list? If we end up with multiple tasks in `.taskcluster.yml`, the error is likely going to be misleading.

Or, if we think multiple tasks are going to be unlikely (which is probably the case), we should either
- error out early, if there are multiple tasks
- add a note to the error message about multiple tasks and the diff being against one of them.

escapewindow:
We already have multiple tasks, in [Focus](https://github.com/mozilla-mobile/focus-android/blob/master/.taskcluster.yml). Agreed the error message will be misleading; do you have a preference? I could concatenate all the error messages...

tomprince:
I think I'd probably do something like collect the diff and then say `"Task doesn't match any rebuilt task: <list of diffs>"`. That seems like the most useful thing for debugging later.

escapewindow:
`verify_cot --task-type signing --cleanup -- JoyuBbYQQRKGGXoZrPQNiA` works, where `JoyuBbYQQRKGGXoZrPQNiA` is the [busted signing task](https://treeherder.mozilla.org/#/jobs?repo=autoland&revision=66956b121b939bf9a56bc33d6f3b8ffa58df4131&filter-resultStatus=testfailed&filter-resultStatus=busted&filter-resultStatus=exception&filter-resultStatus=success&selectedJob=186223909) from @djmitche 's autoland push.

PR #241, opened 2018-07-04
Fix intermittent test bustage + unclosed session warnings

escapewindow:
Scriptworker tests have been flaky of late, using both pytest-asyncio and pytest-xdist with 4 parallel tests; the symptom is a bunch of tests failed with closed event loop errors. Removing the `event_loop` fixture appears to address this flakiness. I believe some \*scripts import the `event_loop` fixture; we should be able to apply similar fixes downstream. (If we don't, then the imports will fail when we release a new scriptworker with these patches. I can do a pass through the various \*scripts.)

Also, our tests have been peppered with aiohttp unclosed session warnings. I tracked these down to the `rw_context` fixture; by adding an explicit close for `context.session`, `context.queue.session`, and `context.temp_queue.session` if they exist and are `aiohttp.ClientSession` instances, we no longer have those warnings.

### DIFF
--- a/scriptworker/test/__init__.py
+++ b/scriptworker/test/__init__.py
@@ -272,6 +272,16 @@ def tmpdir2():
         yield tmp
 
 
+async def _close_session(obj):
+    """Get rid of all the unclosed session warnings.
+
+    """
+    if not hasattr(obj, 'session'):
+        return
+    if isinstance(obj.session, aiohttp.ClientSession):
+        await obj.session.close()
+
+
 @pytest.mark.asyncio
 @pytest.yield_fixture(scope='function', params=['firefox'])
 async def rw_context(request):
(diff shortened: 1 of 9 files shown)


scriptworker/test/__init__.py
```diff
@@ -236,20 +236,6 @@ def integration_create_task_payload(config, task_group_id, scopes=None,
     }
 
 
-@pytest.yield_fixture(scope='function')
```
MihaiTabara: +1 on testing the other scripts. At least `beetmoverscript` and `bouncerscript` are heavily using `event_loop`. Am happy to help with removing the same way we did here in a subset of the *scripts if you want to divide work. I can take care of {balrog,beetmover,shipit,bouncer}scripts.

scriptworker/test/__init__.py
```diff
@@ -272,6 +258,16 @@ def tmpdir2():
         yield tmp
 
 
+async def _close_session(obj):
+    """Get rid of all the unclosed session warnings.
+
+    """
+    if not hasattr(obj, 'session'):
+        return
+    if isinstance(obj.session, aiohttp.ClientSession):
```
MihaiTabara: Neat way to make this generic 👍

Issue #242, opened 2018-07-04
scriptworker 13.0.0 deploy hangs depsigning scriptworkers


escapewindow:
Pre-deploy, signing tasks were taking some number of minutes to resolve.
Post-deploy, signing tasks are finishing, but not resolving for 2 hours. This needs fixing.

This could be due to one of the other dependencies being bumped (ptyprocess, taskcluster-client) but is more likely due to the `run_tasks` changes introduced in scriptworker 13.0.0.

escapewindow:
I think I figured it out.

We create a [`timeout_task`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L413) that calls [`max_timeout`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L518), and include that in the [set of futures to `await`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L417). Since `max_timeout` doesn't exit for 2 hours, we sit there waiting for it to [sleep the entire 2 hours](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L534) before [returning](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L535).

PR #243, opened 2018-07-05
use `asyncio.wait`'s `timeout` kwarg for `task_max_timeout`

escapewindow:
In #242 , we noticed scriptworker 13.0.0 introduced a hang in running tasks. Because the timeout task now slept for the entire `task_max_timeout` and was one of the tasks we were awaiting, all tasks took the entire `task_max_timeout` time to complete.

This patch refactors `run_task` to use the builtin `timeout` kwarg in `asyncio.wait`. If there are pending futures, we `kill_proc`, which raises a `ScriptWorkerTaskException`. We then clean up in the `finally` block.

This allows us to remove the `max_timeout` and `get_future_exception` functions.

I don't want to roll this out and then go on a long PTO, so I'm currently thinking I'll leave this for review, and roll this out when I get back (assuming it passes review). If we need to release a new scriptworker in the meantime, we should branch off the [12.1.0 tag](https://github.com/mozilla-releng/scriptworker/releases/tag/12.1.0).

...aand it looks like I need a followup fix for the intermittent bustage. I think I'll fix by explicitly passing in an event loop into certain functions from the tests.


escapewindow:
Alternate solutions were here:

* https://github.com/escapewindow/scriptworker/commits/first-completed
* https://github.com/escapewindow/scriptworker/commits/cancel-callback
* and the original `asyncio.call_later(timeout, max_timeout, *args)` in scriptworker 12.1.0 and earlier.

escapewindow:
@Callek , @tomprince what do you think?
As mentioned in the [original comment](https://github.com/mozilla-releng/scriptworker/pull/243#issue-199581863), I think this deploy will probably wait until after I get back on July 16.

scriptworker/test/test_task.py
```diff
@@ -297,14 +298,51 @@ def test_prepare_to_run_task(context):
...
+    # create all 6 files
+    context.config['task_max_timeout'] = 5
+
+    pre = arrow.utcnow().timestamp
+    with pytest.raises(ScriptWorkerTaskException):
+        await swtask.run_task(context)
+    post = arrow.utcnow().timestamp
+    # I don't love these checks, because timing issues may cause this test
+    # to be flaky. However, I don't want a non- or long- running test to pass.
+    # Did this run at all?
```
Callek: could python freezegun help here?

escapewindow: I may be misreading, but afaict, freezegun makes the datetime calls return a predictable time. How would we know if our timeout task took too long or short if both calls return the same timestamp?

Callek: I was imagining using it for https://github.com/spulec/freezegun#manual-ticks and/or https://github.com/spulec/freezegun#moving-time-to-specify-datetime

escapewindow: The time tests in this test are specifically to measure how long we're running the task for before killing: we need to know that the task runs for at least as long as max timeout, but not significantly longer (e.g., if the task runs to completion without being killed, that should be an error). Manually changing the timestamps to pass means those tests do nothing useful, and we should remove them. I know that timestamp tests are potentially flaky, but I also want to know if the kill isn't working properly.

Callek: Fair, I was musing that it may help, not blocking on it, to be extra clear

escapewindow: Ok. Thanks for the pointer; looks like that should be helpful in the future for other time-based tests.

I'm going to push scriptworker 14.0.0 changes here and then merge.

PR #244, opened 2018-07-05
explicitly pass in an `event_loop` in testing.

escapewindow:
Create a `context.event_loop` property to standardize the loop we use.
In regular usage, we default to `asyncio.get_event_loop()`. In testing,
we're using `pytest-asyncio` which creates new `event_loop` fixtures per
test and closes them at the end; and `pytest-xdist` which runs tests in
parallel. When combined, bare `asyncio.get_event_loop()` calls can
result in intermittent bustage (closed loop errors).

By modifying the `rw_context` fixture to populate its `event_loop` with
the `pytest-asyncio` `event_loop`, we get most of the way to the fix.
`main` and `rebuild_gpg_homedirs` are entry points that create their own
`context` objects, so they now take optional `event_loop` kwargs to
override the loop during testing.

### DIFF
--- a/scriptworker/test/test_context.py
+++ b/scriptworker/test/test_context.py
@@ -2,7 +2,9 @@
 # coding=utf-8
 """Test scriptworker.context
 """
+import asyncio
 import json
+import mock
 import os
 import pytest
 import taskcluster
@@ -124,3 +126,23 @@ def test_get_credentials(context):
     expected = {'asdf': 'foobar'}
     context._credentials = expected
     assert context.credentials == expected
+
+
+def test_new_event_loop(mocker):
+    """The default context.event_loop is from `asyncio.get_event_loop`"""
+    fake_loop = mock.MagicMock()
+    mocker.patch.object(asyncio, 'get_event_loop', return_value=fake_loop)
+    context = swcontext.Context()
+    assert context.event_loop == fake_loop
+
+
+def test_set_event_loop(mocker):
+    """`context.event_loop` returns the same value once set.
+
+    (This may seem obvious, but this tests the correctness of the property.)
+
+    """
+    fake_loop = mock.MagicMock()
+    context = swcontext.Context()
+    context.event_loop = fake_loop
+    assert context.event_loop == fake_loop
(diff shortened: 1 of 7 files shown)


escapewindow:
Rebased against 14.0.0.

escapewindow:
@JohanLorenzo do you feel comfy reviewing this? It doesn't block anything, but reducing the number of intermittent test bustages is generally nice to have :)

scriptworker/test/test_context.py
```diff
@@ -124,3 +126,23 @@ def test_get_credentials(context):
...
     context._credentials = expected
     assert context.credentials == expected
+
+
+def test_new_event_loop(mocker):
+    """The default context.event_loop is from `asyncio.get_event_loop`"""
+    fake_loop = mock.MagicMock()
+    mocker.patch.object(asyncio, 'get_event_loop', return_value=fake_loop)
+    context = swcontext.Context()
+    assert context.event_loop == fake_loop
```
JohanLorenzo: Nit: I think we should use the `is` operator in this case. Thus, we're sure this is the same object.

scriptworker/test/test_context.py
```diff
@@ -124,3 +126,23 @@ def test_get_credentials(context):
...
+def test_set_event_loop(mocker):
+    """`context.event_loop` returns the same value once set.
+
+    (This may seem obvious, but this tests the correctness of the property.)
+
+    """
+    fake_loop = mock.MagicMock()
+    context = swcontext.Context()
+    context.event_loop = fake_loop
+    assert context.event_loop == fake_loop
```
JohanLorenzo: Nit: `is` operator

scriptworker/test/test_context.py
escapewindow: Makes sense! Done.

scriptworker/test/test_context.py
escapewindow: Done.

escapewindow:
Closing and reopening to try to poke github...

escapewindow:
Thank you!

PR #248, opened 2018-08-30
Bug 1486532 - consider task['tags']['worker-implementation'] when guessing worker implementation

petemoore:
In [bug 1486532](https://bugzil.la/1486532) empty properties in generic-worker task payloads in gecko decision task generated tasks will be disappearing. Currently, scriptworker guesses worker implementation based on properties that are included in the task payloads. Therefore, if these properties disappear, this guessing function will unfortunately no longer work. However, the good news is, there is trustworthy metadata contained in the gecko decision-task generated task definitions which provides this information, which is the `worker-implementation` tag. I suspect this didn't exist when the worker implementation guess code was written, but now that it does, it should be a reliable source of information.

I'm not sure if scriptworker evaluates tasks that are generated outside of the gecko decision task which may not have the `worker-implementation` tag, such as release tasks? Therefore, rather than assuming the presence of the `worker-implementation` tag, the previous mechanics of guessing the worker implementation have been left in place essentially as a fallback (with the added advantage that if they conflict with the tag, an exception will be raised).

Note, if the `worker-implementation` tag turns out to be present in all tasks that scriptworker evaluates, we could (and probably should) remove the former guessing code.

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -325,6 +325,8 @@ def guess_worker_impl(link):
         worker_impls.append("generic-worker")
     if task['payload'].get("osGroups") is not None:
         worker_impls.append("generic-worker")
+    if task.get('tags', {}).get("worker-implementation", {}):
+        worker_impls.append(task['tags']['worker-implementation'])
 
     for scope in task['scopes']:
         if scope.startswith("docker-worker:"):
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1827,7 +1827,7 @@ async def verify_chain_of_trust(chain):
 
 
 # verify_cot_cmdln {{{1
-def verify_cot_cmdln(args=None):
+def verify_cot_cmdln(args=None, event_loop=None):
     """Test the chain of trust from the commandline, for debugging purposes.
 
     Args:
@@ -1858,14 +1858,14 @@ or in the CREDS_FILES http://bit.ly/2fVMu0A""")
     log = logging.getLogger('scriptworker')
     log.setLevel(logging.DEBUG)
     logging.basicConfig()
-    loop = asyncio.get_event_loop()
+    event_loop = event_loop or asyncio.get_event_loop()
     conn = aiohttp.TCPConnector()
     try:
         session = aiohttp.ClientSession(connector=conn)
         context = Context()
         context.session = session
         context.credentials = read_worker_creds()
-        context.task = loop.run_until_complete(context.queue.task(opts.task_id))
+        context.task = event_loop.run_until_complete(context.queue.task(opts.task_id))
         context.config = dict(deepcopy(DEFAULT_CONFIG))
         context.config.update({
             'cot_product': opts.cot_product,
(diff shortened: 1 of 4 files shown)


scriptworker/cot/verify.py
```diff
@@ -325,6 +325,8 @@ def guess_worker_impl(link):
         worker_impls.append("generic-worker")
     if task['payload'].get("osGroups") is not None:
         worker_impls.append("generic-worker")
+    if task['tags'].get("worker-implementation") is not None:
```
escapewindow: We may want

`if task.get('tags', {}).get("worker-implementation") is not None:`

in case the task doesn't have a `tags` dict. If `tags` is required by the schema, we can keep this, but we'd need to update all the scriptworker test tasks to include `tags`.

tomprince:
scriptworker is used for non-taskgraph based projects as well. I'm not sure what the state of CoT integration for them is, or whether they provide this metadata. They certainly could be modified to provide it. (cc @JohanLorenzo)

scriptworker/cot/verify.py
petemoore: Ahhh, I hadn't thought about that breaking the tests. Sorry about that!

Indeed [tags is a required property](https://github.com/taskcluster/taskcluster-queue/blob/fd8027f1231421e95f11399a45aac3e59ade8525/schemas/v1/task.yml#L192) which is always present in queue responses, but I hadn't considered the tests.

I can push a new commit which updates the tests, or as you say we could switch to your proposed version, which would avoid needing to touch the tests and would still work if `tags` became optional at a point in the future. On reflection, I tend to prefer your proposed alternative if statement.

petemoore:
> scriptworker is used for non-taskgraph based projects as well. I'm not sure what the state of CoT integration for them is, or whether they provide this metadata. They certainly could be modified to provide it. (cc @JohanLorenzo)

That indeed would be nice, for consistency, and allowing the guess function to be removed at some point. ++

escapewindow:
Looks good! We'll need to cover that last line before we can merge.
I can take a look at adding that test...

escapewindow:
I made quite a few changes here -- @tomprince or @JohanLorenzo want to give it an eyeball before I merge and release 15.0.1?

petemoore:
@escapewindow Many thanks for updating the docs/tests which I had not done - apologies for that, I should have covered that at the same time. I was actually surprised you were able to push to the PR branch which is in my personal fork of the project - I never knew that was possible! I appreciate you reacting to the PR so quickly - thank you.

Am I right in thinking once these lines below are bumped to 15.0.1 that this is rolled out globally, and I can land the gecko changes in [bug 1486532](https://bugzil.la/1486532)?

```
pmoore@Petes-iMac:~/git/puppet master $ grep -r 'scriptworker==' .
./modules/signing_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/bouncer_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/transparency_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/addon_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/tree_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/pushsnap_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/shipit_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/beetmover_scriptworker/files/requirements.txt:scriptworker==15.0.0
./modules/balrog_scriptworker/files/requirements-3.txt:scriptworker==15.0.0
./modules/pushapk_scriptworker/files/requirements.txt:scriptworker==15.0.0
```

Or are there any other deployments I'll need to wait for?

Many thanks!

escapewindow:
Closing and reopening to try to un-confuse github.

PR #251, opened 2018-09-05
Allow bouncer to run on mozilla-central.

tomprince:


scriptworker/constants.py
```diff
@@ -224,6 +224,9 @@
                 'project:releng:beetmover:bucket:nightly': 'all-nightly-branches',
                 'project:releng:beetmover:bucket:release': 'all-release-branches',
 
+                'project:releng:bouncer:action:aliases': 'all-release-branches',
```
JohanLorenzo: Discussed with @tomprince over IRC: I'm glad we now enforce these scopes!

escapewindow:
This will probably be problematic for staging releases on Try. We probably want to add a `:server:` scope and only restrict the production servers, and allow the actions across all branches including Try when using a dev bouncer instance.

tomprince:
> This will probably be problematic for staging releases on Try. We probably want to add a :server: scope and only restrict the production servers, and allow the actions across all branches including Try when using a dev bouncer instance.

Looking at the other workers that limit things for nightly-vs.-release, those things are different scopes than are allowed on staging at all. I guess we don't have a way yet to limit actions scopes in production, but not limit those same scopes in staging.

I think we can do better here, but I'd rather unblock updating nightly bouncer entries now, and follow up with better isolation between nightly and release branches.

Callek:
Shouldn't we just use the nightly branches check instead for needs here, or am I misunderstanding? I have a slight worry that this opens up mc for more interactions with release machinery

tomprince:
We do want all of the release branches and mozilla-central to talk to the production bouncer. We'd also like to limit nightly to the `:locations` action, and release to the `:submissions` and `:aliases` actions. However we *don't* want to limit in any what what actions we can do on try.

I don't think there is any way to limit what scopes are used where that doesn't stop those scopes being used on try. I think we should figure out a way of handling this, but I'd like to unblock https://bugzilla.mozilla.org/show_bug.cgi?id=1432656 first.

MihaiTabara:
Thanks for taking care of this whilst I was PTO yesterday!

PR #255, opened 2018-09-07
Bug 1473416: Allow try to talk to staging ship-it and snap workers.

tomprince:
It looks like the `:mock` snap scope doesn't touch external resources, so no reason to limit it.


JohanLorenzo:
Makes sense to me!

PR #256, opened 2018-09-08
bug 1485680 - `jsone.render` action hooks' `hookPayload`

escapewindow:
We define the treeherder symbol for relpro action tasks to be `"${input.release_promotion_flavor}"`. This works for non-hook actions, but for hook actions we end up not rendering that json-e and cot verification fails.

The test for the change is `verify_cot --task-type action --cleanup -- PV3gwVnWSgaGzUQuqVr41w`

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -1150,8 +1150,9 @@ async def get_action_context_and_template(chain, parent_link, decision_link):
         tmpl = {'tasks': [action_defn['task']]}
     else:
         tmpl = await get_in_tree_template(decision_link)
+        rendered_hook_payload = jsone.render(action_defn['hookPayload'], jsone_context)
         for k in ('action', 'push', 'repository'):
-            jsone_context[k] = deepcopy(action_defn['hookPayload']['decision'][k])
+            jsone_context[k] = deepcopy(rendered_hook_payload['decision'][k])
         jsone_context['action']['repo_scope'] = get_repo_scope(parent_link.task, parent_link.name)
 
     return jsone_context, tmpl


scriptworker/cot/verify.py
```diff
@@ -1145,7 +1145,7 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
     else:
         tmpl = await get_in_tree_template(decision_link)
         for k in ('action', 'push', 'repository'):
-            jsone_context[k] = deepcopy(action_defn['hookPayload']['decision'][k])
+            jsone_context[k] = jsone.render(action_defn['hookPayload']['decision'][k], jsone_context)
```
tomprince: Looking at the [spec](https://docs.taskcluster.net/docs/manual/using/actions/spec) and the [tools implementation](https://github.com/taskcluster/taskcluster-tools/blob/2d2897bd5e8cab418fbfd518dbcc9c892abfa066/src/views/UnifiedInspector/ActionsMenu.jsx#L480-L502), I think this is incorrect. My reading is that the returned context should *just* be `hookPayload` rendered with the previous context.

Having written that, looking at the hook definition in [ci-admin](https://hg.mozilla.org/build/ci-admin/file/edad9f8f78d42bc547c1f684b0f9dc116ccbd00a/ciadmin/generate/in_tree_actions.py#l155), it looks like the context under goes some more modification before being passed to the in-tree `.taskcluster.yml`.

escapewindow:
Agreed that while this is a working solution, it may not be the best solution.

Do you have a preferred solution here?

tomprince:
It isn't entirely clear to me what is going on here, or what the correct behavior is. That suggests there should probably comment here explaining what is going on, including how what we are doing here differs from how hooks are implemented.

I could definitely be convinced that the info that we generate lines up with what gets passed into the `.taskcluster.yml`, and that we would rather trust the sources we use. But I'd like to understand how that happens.

escapewindow:
Yeah, it looks like ci-admin prepends a `$let` section at the top, so the template includes information from the `actions.json` `hookPayload`. I'll give that a shot.

escapewindow:
https://github.com/escapewindow/scriptworker/commit/9ec59c1bb8720af1648bb5de4694e0a62bfb957d doesn't appear to work either... I end up with an object for my ACTION_TASK_GROUP_ID instead of a string. My guess is something else is rendering the `hookPayload`, but I don't know what.

@djmitche do you know how [this line](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l162) is getting rendered twice, so given a `payload.decision.action.symbol` of `"${input.release_promotion_flavor}"` we end up with a symbol of, e.g. `firefox_promote` ?

scriptworker/cot/verify.py
djmitche: Yeah, I'm not sure what this line is about.  But rendering some sub-property is usually a red flag that you're doing something wrong.

escapewindow: Ok. The easy workaround is changing the relpro action symbol to a static string, but it's been very useful to have it be templatized. I'm probably going to ask more questions to figure out how action-hooks are able to double-render this payload on the ci-admin / hook / taskgraph side; once we figure that out, hopefully it'll be easier to replicate on the scriptworker side.

djmitche:
Nothing should be rendered twice. I'm not sure what specific hook you're looking for, but they all look pretty much the same, so let's take [this one](https://tools.taskcluster.net/hooks/project-comm/in-tree-action-1-generic%2F894552ca01).  

```json
{
  "$let": {
    "action": {
      ..
      "symbol": "${payload.decision.action.symbol}",
      ..
    },
    ..
  },
  "in": {
    "$let": {
     ..,
     }
    },
    "in": {
      ..,
      "extra": {
        "$merge": [
          {
            "treeherder": {
              "$merge": [
                ..,
                {
                  "$if": "tasks_for == \"hg-push\"",
                  ...,
                  "else": {
                    "$if": "tasks_for == \"action\"",
                    "then": {
                      "symbol": "${action.symbol}",
                      ...
                    },
                    ...,
                  }
                }
              ]
            }
          },
        ]
      },
      ...
    }
  }
}
```

so, given a context with `{payload: {action: {symbol: "${input.release_promotion_flavor}"}}}`, I would expect the rendered result to be

```json
{
  extra: {
    treeherder: {
      symbol: "${input.release_promotion_flavor}",
      ...,
    },
    ...,
  },
  ...,
}
```

That is a particluarly odd symbol, but there's no reason it wouldn't work.

So, that's trying to answer the question you asked.. but I don't understand the rest of this PR, and I dont' see the inputs and outputs to the alleged "double-rendering", so maybe it wasn't the right question.

Just to be clear, ci-admin doesn't do any rendering at all for actions -- it doesn't even `import jsone`. It does wrap `.taskcluster.yml` in a `{$let: .., in: <taskcluster.yml>}` so that the `taskcluster.yml` gets the variables it expects, based on those provided by the action context.

There are two renderings going on, though.  Spelling it out in code:

```python
# invoke the action (from tools UI, treeherder, etc.)
def invoke_action(action):
    action_context = { .. }  # as defined in https://docs.taskcluster.net/docs/manual/using/actions/spec
    triggerHook(action['hookGroupId'], action['hookId'], jsone.render(action['hookPayload'], action_context))

# the triggerHook API implementation (inside the hooks service)
def triggerHook(hookGroupId, hookId, payload):
    hook = getHook(hookGroupId, hookId)
    task = jsone.render(hook['task'], payload)
    queue.createTask(task)
```

or more succinctly,

```python
task = jsone.render(hook['task'],
                    jsone.render(action['hookPayload'], action_context))
```

It seems like `get_action_context_and_template`, and `get_jsone_context_and_template` which calls it, just return a template and context (an "appropriate" one, per the docstring).  The only useful thing to do with a template and context is render them, so presumably that's what the caller does.  That, then, would be the outer `render` in the succinct definition above.  So the inner bit would be `jsone.render(action['hookPayload'], action_context)`.  So it seems that's what `get_action_context_and_template` should do.  I assume there are some complexities in reverse-engineering what `action_context` should be, which is what led to the rendering of only some properties of the hookPayload?

djmitche:
BTW I, too, have a hard time keeping this all in my head.  There are just a lot of places that inputs come from (user, decision task, hook definition).

scriptworker/cot/verify.py
djmitche: There may be a bug - a repro recipe would be helpful!

escapewindow: I've pushed a new change that renders the entirety of `hookPayload` instead of subsections. Does this match what happens in taskcluster? Or is this a red flag still?

escapewindow:
This is essentially the same patch, but it renders all of `hookPayload` like the `invoke_action` here (where is this code?)

```
# invoke the action (from tools UI, treeherder, etc.)
def invoke_action(action):
    action_context = { .. }  # as defined in https://docs.taskcluster.net/docs/manual/using/actions/spec
    triggerHook(action['hookGroupId'], action['hookId'], jsone.render(action['hookPayload'], action_context))

# the triggerHook API implementation (inside the hooks service)
def triggerHook(hookGroupId, hookId, payload):
    hook = getHook(hookGroupId, hookId)
    task = jsone.render(hook['task'], payload)
    queue.createTask(task)
```

It looks like I could, and possibly should, copy everything from `hookPayload.decision` into my context? This patch, as well as the previous patch in this PR, results in `verify_cot --task-type action --cleanup -- PV3gwVnWSgaGzUQuqVr41w` passing. Which unblocks relpro-as-an-action-hook.

djmitche:
I'm still confused about the double-rendering.  I see the link to [the output](https://treeherder.mozilla.org/#/jobs?repo=maple&revision=358b78517a800f02e025a9020ffbf0f8516c204c&searchStr=promote&selectedJob=198186381).  Let me see if I can work out what's going on here, in a live-blogigng format.. :)

actions.json has:
```
    {
      "name": "release-promotion",
      "kind": "hook",
      "hookGroupId": "project-gecko",
      "hookId": "in-tree-action-3-release-promotion/977d669622",
      "hookPayload": {
        "decision": {
          "action": {
            "cb_name": "release-promotion",
            "description": "Promote a release.",
            "name": "release-promotion",
            "symbol": "${input.release_promotion_flavor}",
            "taskGroupId": "E7w_SFN8TouLJXQ-xNVbhw",
            "title": "Release Promotion"
          },
          "parameters": {
            "$eval": "parameters"
          },
          "push": {
            "owner": "mozilla-taskcluster-maintenance@mozilla.com",
            "pushlog_id": "926",
            "revision": "358b78517a800f02e025a9020ffbf0f8516c204c"
          },
          "repository": {
            "level": "3",
            "project": "maple",
            "url": "https://hg.mozilla.org/projects/maple"
          }
        },
        "user": {
          "input": {
            "$eval": "input"
          },
          "taskGroupId": {
            "$eval": "taskGroupId"
          },
          "taskId": {
            "$eval": "taskId"
          }
        }
      },
      ...
}
```

so step one is to render the hookPayload with a context

```
{
  taskGroupId: 'E7w_SFN8TouLJXQ-xNVbhw',  // from the decision task
  taskId: null,                           // run on the task-group
  input: {                                // from action task's task.extra
    "build_number": 5,
    "next_version": "62.0b14",
    "partial_updates": {
      "62.0b10": {
        "buildNumber": 2,
        "locales": ["de", "ja", "ja-JP-mac", "zh-TW"]
      },
      "62.0b12": {
        "buildNumber": 16,
        "locales": [...],
      }
    },
    "previous_graph_ids": ["E7w_SFN8TouLJXQ-xNVbhw"],
    "release_eta": "",
    "release_promotion_flavor": "promote_firefox",
    "version": "62.0b13"
  },
  "parameters": { ... }                   // from the decision task
}
```

the result being:

```
{
  "decision": {
    "action": {
      "cb_name": "release-promotion",
      "description": "Promote a release.",
      "name": "release-promotion",
      "symbol": "promote_firefox",              // ***
      "taskGroupId": "E7w_SFN8TouLJXQ-xNVbhw",
      "title": "Release Promotion"
    },
    "parameters": { ...  },  // copy of those from decision task
    "push": {
      "owner": "mozilla-taskcluster-maintenance@mozilla.com",
      "pushlog_id": "926",
      "revision": "358b78517a800f02e025a9020ffbf0f8516c204c"
    },
    "repository": {
      "level": "3",
      "project": "maple",
      "url": "https://hg.mozilla.org/projects/maple"
    }
  },
  "user": {
    "input": { .. }, // same as above
    "taskGroupId": 'E7w_SFN8TouLJXQ-xNVbhw',
    "taskId": null
  }
}
```

so the symbol is already at its final value with just one render.  That last block would then be used as context for rendering the hook body, putting that symbol into its final position at `task.extra.treenerder.symbol`.

Including JSON-e in the action decorator in [taskcluster/taskgraph/actions/release_promotion.py](https://hg.mozilla.org/projects/maple/file/358b78517a80/taskcluster/taskgraph/actions/release_promotion.py#l57) is clever, which is probably why it led to some confusion, but seems to work and I see no reason why it shouldn't work.

I would expect action-task CoT validation to follow similar lines, and the usual two-part process of (1) guess what initiated this and (2) rigorously validate the guess

 * guess the relevant entry in `actions.json` using `cb_name` (guessing)
 * guess the json-e context (based on comments above)
 * validate that the input is sane
 * verify `render(hookBody, render(hookPayload, guessedContext))` matches the resulting task definition

So given all of that, I'm not sure why the code in this patch is looping over action, push, and repository?  It seems that the `jsone_context` variable is not, in fact, equal to either JSON-e context concerned here, but to the value at the `decision` key of `render(hookPayload, calculatedContext)`.

Looking more deeply at `get_action_context_and_template`, `tmpl` is the decision task's `.taskcluster.yml`, not the hook body. Ah-hah!  So the last step in my list of steps above is actually

 * verify `render(taskcluster_yml, derive_taskcluster_yml_context_from(hookPayload, guessedContext))`

where `derive_taskclsuter_yml_context_from` is responsible for emulating the `$let` which ci-admin wraps around the `.taskcluster.yml`. So basically `emulate_let(render(hookPayload, guessedContext))`.  Thereby avoiding the need to download and trust the hook payload.

That makes sense, but it'd be good to see it broken out into those two functions, so that if the `$let` changes it's clear where to make the modification.  An alternative would be

 * verify `render(wrap_with_let(taskcluster_yml), render(hookPayload, guessedContext))`

where `wrap_with_let` is based closely on [ci-admin](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l155). Then any changes to ci-admin could be reflected directly in CoT. And this arrangement more closely emulates the "real" process, making it a bit easier to understand.

Anyway, I don't think this patch is correct now: `jsone.render(action_defn['hookPayload'], jsone_context)` uses a context that doesn't have the right shape (it should be `{taskGroupId, taskId, input, parameters}`, and then modifies that context moments later.  Or maybe `jsone_context` is more of `dict_with_various_keys_that_things_expect_in_their_context`.  In which case I guess this will work for relpromo actions, and maybe others, but seems like it's going to break if you look at it sideways.

escapewindow:
So https://github.com/escapewindow/scriptworker/commit/9ec59c1bb8720af1648bb5de4694e0a62bfb957d is an attempt to fix this via the same logic in ci-admin [here](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l154). It's likely broken because I didn't render `hookPayload` with anything; once I render `hookPayload` properly it will probably work.

*However*, I don't like the `$let` solution because in both ci-admin and that patch, we're hardcoding a large block of `$let` in two places. If we land the `$let` solution (with fixes), my next patch is going to be to `ci-admin`, to add a comment before [this block](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l154) to say that if you change this block at all, make sure you first land a backwards-compatible change to `scriptworker.cot.verify`.

escapewindow:
I think I'm going to try both solutions, once I get past this other higher priority task that landed on my lap:

1. Keep on this patch, but render `hookPayload` with the proper jsone_context.
2. Switch to the action-hook2 branch (aka `$let` solution), render `hookPayload` with the proper jsone_context, and look a lot closer to what's actually happening, at the expense of hardcoding the `$let` block in two places
3. Float the idea in my head about how we can get rid of that large `$let` block in two places, e.g. can we make `hookPayload` a drop-in `$let` block? Or something?

djmitche:
> add a comment before this block to say that if you change this block at all, make sure you first land a backwards-compatible change to scriptworker.cot.verify.

this sounds great.  I agree that duplication is unfortunate, but in a way expected here: CoT is replaying the work of other components to verify that it was done correctly.  Doing so in a fashion where it's easy to translate a change in one of the two duplicate places into the other will, I think, help future devs not lose their minds.

One way to get rid of the `$let` block would be to embed that logic in `.taskcluster.yml` directly, activated by looking at `tasks_for`.  So that would become a large `$let` with things like
```
"foo": {$if: 'tasks_for == "action"', then: "${some_expression_for_action_jsone_context}", else: "${some_expression_for_mozilla_taskcluster_jsone_context}"},
...
```
The only tricky bit is `cb_name`, which *must* be hardcoded in the hook definitions (otherwise users could run whatever action they wanted..). But both ci-admin and CoT could do some very specific futzing like
```
template['tasks'][0]['$let']['cb_name'] = cb_name
```

That sounds harder than just maintaining two parallel versions of the `$let` block to me, but YMMV :)

escapewindow:
Yay: I got the `action-hook2` branch to pass `verify_cot` on a relpro action-hook!

Boo: I had to put in some ugly hardcodes to do it.
https://github.com/mozilla-releng/scriptworker/compare/master...escapewindow:action-hook2

- where do we get `action_perm` from ? (This issue likely blocks)
- why do I need to set `taskId` to `None` and `ownTaskId` to the action task's `taskId` to get verification to pass? (This seems like action weirdness around `taskId` and `ownTaskId` that may have been there from the beginning.)

escapewindow:
> where do we get action_perm from ? (This issue likely blocks)

Looks like https://hg.mozilla.org/build/ci-configuration/file/tip/actions.yml ?

djmitche:
To *guess* what action_perm is, yes, I think you can look at the hookId.  To *validate* action_perm, assert that it is either `"generic"` or equal to `cb_name`.  The only way to cheat this validation is to run a non-generic action with a generic hook.  That generic hook would run with fewer scopes than the cb_name-specific hook, but more people would have permission to call triggerHook for the generic hook.  I have a hard time seeing how such a thing would be useful to an attacker, but it's also [easy enough to prevent](https://bugzilla.mozilla.org/show_bug.cgi?id=1491186) if it's not already prevented.

[The spec](https://docs.taskcluster.net/docs/manual/using/actions/spec#json-e-context) suggests that `ownTaskId` need not be set.

The `taskId` is None because the action you're working on applies to the whole taskGroup, rather than a single task.

escapewindow:
> To guess what action_perm is, yes, I think you can look at the hookId. To validate action_perm, assert that it is either "generic" or equal to cb_name

Except I have to know what `action_perm` is to populate some strings in the `$let`. I don't want to create 2 different `$let`s, render both, and make sure the task is one of them; I'd rather have a breadcrumb tell me what `action_perm` should be.

If I don't set `ownTaskId`, I think the route changes and the runtime task definition doesn't verify properly. As for `taskId`, I hope I don't have to special case this for relpro vs non-relpro. I think I don't ever set it though.

tomprince:
I think https://github.com/escapewindow/scriptworker/compare/action-hook2...tp-tc:action-fix fixes up the hard-coded things. (though not action-perm)

escapewindow:
Weird, using `tp-tc/action-fix` for `verify_cot --task-type action --cleanup -- PV3gwVnWSgaGzUQuqVr41w` gives me

```
CRITICAL:scriptworker.cot.verify:action PV3gwVnWSgaGzUQuqVr41w: the runtime task doesn't match any rebuilt definition!
["[('change',\n"
 "  'created',\n"
 "  ('2018-09-14T04:09:25.705Z', '2018-09-08T00:40:17.668Z')),\n"
 " ('change',\n"
 "  'deadline',\n"
 "  ('2018-09-15T04:09:25.705Z', '2018-09-09T00:40:17.668Z')),\n"
 " ('change',\n"
 "  'expires',\n"
 "  ('2019-09-14T04:09:26.705Z', '2019-09-08T00:40:18.668Z')),\n"
 " ('change',\n"
 "  'payload.artifacts.public.expires',\n"
 "  ('2019-09-14T04:09:25.705Z', '2019-09-08T00:40:17.668Z'))]"]
```

djmitche:
> Except I have to know what action_perm is to populate some strings in the $let. I don't want to create 2 different $lets, render both, and make sure the task is one of them; I'd rather have a breadcrumb tell me what action_perm should be.

You can find the hookId without rendering anything (it's in the action in `actions.json`).

> If I don't set ownTaskId, I think the route changes and the runtime task definition doesn't verify properly.

That suggests something's going wrong.  [Tools](https://github.com/taskcluster/taskcluster-tools/blob/2d2897bd5e8cab418fbfd518dbcc9c892abfa066/src/views/UnifiedInspector/ActionsMenu.jsx#L455) and treeherder don't supply `onwTaskId` when rendering `action['hookPayload']`.

> As for taskId, I hope I don't have to special case this for relpro vs non-relpro. I think I don't ever set it though.

Some actions run on a taskGroup, while others run on a specific task. You can tell which by looking at `action['context']`.  If that is `[]` then it runs on a task group and `taskId` should be null.  Otherwise, it runs on a specific task and `taskId` should be set.

Also, in talking to Callek yesterday I discovered that `task.extra.context` contains the entire original context used to render `action['hookPayload']`.  That will simplify a lot of the "guessing" we have discussed above.

I'll take [this task](https://tools.taskcluster.net/groups/CRe2ISWVSmiXj6In44a1hg/tasks/OCOen1dBTlmd3dc9BHbmwQ/details) from start to finish.

```javascript
const task = await queue.task('OCOen1dBTlmd3dc9BHbmwQ');
const taskcluster_yml = await fetch_taskcluster_yml('OCOen1dBTlmd3dc9BHbmwQ');
const actions_json = await fetch_actions_json('OCOen1dBTlmd3dc9BHbmwQ');
```

I start by guessing `cb_name` based on `task.payload.env.ACTION_CALLBACK` -- `retrigger-mochitest`.  The [decision task's `actions.json`](https://queue.taskcluster.net/v1/task/CRe2ISWVSmiXj6In44a1hg/runs/0/artifacts/public/actions.json) has an action `action.hookPayload.context.cb_name == 'retrigger-mochitest'`, so

```javascript
const action =  find_action(actions_json, cb_name);  /*{
      "context": [{"test-type": "mochitest"}, {"test-type": "reftest"}],
      "description": "Retriggers the specified mochitest/reftest job with additional options",
      "hookGroupId": "project-gecko",
      "hookId": "in-tree-action-1-generic/a048376fee",
      "hookPayload": {..},
      "kind": "hook",
      "name": "retrigger",
      "schema": {..}
      "title": "Retrigger Mochitest/Reftest"
};*/
```

Based on action['hookId'], I can guess that `const actionPerm = 'generic'` (incidentally, a patch to surface `cb_name` and `actionPerm` at the top level of the action would likely be accepted).  From the above, we have

```javascript
const hookPayload = action.hookPayload;
const guessedContext = task.extra.context;  /*{
      "input": {},
      "parameters": {..},
      "taskGroupId": "CRe2ISWVSmiXj6In44a1hg",
      "taskId": "Bml3eanGREKna9144qhq3w"
};*/
```

Looking back to

  * verify `render(wrap_with_let(taskcluster_yml, actionPerm), render(hookPayload, guessedContext))`

I've added `actionPerm` as an argument to `wrap_with_let`, making that function a trivial derivation from [ci-admin](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l154).

All that remains is to evaluate the above expression and compare it to `task`, to verify that the guesses match the resulting task.  Add in some sanity-checks on the guesses (especially the `guessedContext` since we just took that verbatim) and we're done!

My thoughts on sanity checks:
 * `guessedContext.input` matches `action.schema`
 * additional checks as needed for specific `cb_name`s - for example I'm sure there's more to check for `release-promotion`?
 * `guessedContext.parameters` matches `parameters.yml` from the decision task
 * `guessedContext.taskGroupId` matches the decision task's `taskId`
 * if `guessedContext.taskId` is set
    * the task it points to has taskGroupId == `guessedContext.taskGroupId`
    * `action.context` is not empty
 * else
   * `action.context` is empty

Of course, there's room for more!

escapewindow:
I've gotten [closer](https://github.com/escapewindow/scriptworker/commits/action-hook2). I'll keep poking at the `taskId`/`ownTaskId` thing.

escapewindow:
> (incidentally, a patch to surface `cb_name` and `actionPerm` at the top level of the action would likely be accepted).

Awesome. `cb_name` is easy to get, via `task.payload.env.ACTION_CALLBACK`. The only way I can currently find `actionPerm` is by parsing the hook string. I'll write an in-tree patch.

escapewindow:
>  I'll keep poking at the `taskId`/`ownTaskId` thing.

I currently suspect [this line](https://hg.mozilla.org/mozilla-central/file/6f21533f643e/.taskcluster.yml#l152) should be using `ownTaskId`, not `taskId`. Testing this change on Try.

djmitche:
```
                  ACTION_TASK_ID: {$json: {$eval: 'taskId'}} # taskId of the target task (JSON-encoded)
```

the comment suggests otherwise -- that's the taskId of the "target task", the one the user selected when they elected to perform an action.  It's not the taskId of the action task itself.  A look at `taskcluster/mach_commands.py` confirms this.

escapewindow:
Yeah, @tomprince pointed this out in channel.

Looks like I'll go with [this fix](https://github.com/escapewindow/scriptworker/commit/b9991755d03d50d0bd195b11a413a8826b117637#diff-340a1f42e56ecb68f523d2badfa58686R1213) even though it looks different than [this line](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l191).

escapewindow:
>     * verify `render(wrap_with_let(taskcluster_yml, actionPerm), render(hookPayload, guessedContext))`
> 
> 
> I've added `actionPerm` as an argument to `wrap_with_let`, making that function a trivial derivation from [ci-admin](https://hg.mozilla.org/build/ci-admin/file/edad9f8/ciadmin/generate/in_tree_actions.py#l154).

To be clear: we're rendering twice. From [Dustin's comment](https://github.com/mozilla-releng/scriptworker/pull/256#issuecomment-421361799):

```
    verify render(wrap_with_let(taskcluster_yml, actionPerm), render(hookPayload, guessedContext))
           ^^^^^^                                             ^^^^^^
```

Just wanted to make it clear that the proposed solution contains two renders, since I vaguely remember people being unhappy with me rendering twice previously.

djmitche:
The double-render is correct since the output of one render becomes the *context* for the other, not the template.

escapewindow:
Makes sense, thanks!

PR #258, opened 2018-09-12
use `cb_name` instead of `name` to identify actions in actions.json

### LINKED ISSUE #254: cot.verify assumes there is a single action per `name`
https://github.com/mozilla-releng/scriptworker/blob/63136f130c75fd5115834e59acf45c2f07f01496/scriptworker/cot/verify.py#L1141

We should probably also compare `taskId`s; that may mean we make the comparison in a function call.

escapewindow:
fixes #254.

With this patch, `verify_cot --task-type signing --cleanup -- aKujikq3Qh-jPgUpg6iPGQ` passes. Without this patch, the same command barfs.


escapewindow:
Gotta fix some tests.

scriptworker/task.py
```diff
@@ -80,7 +80,7 @@ def get_run_id(claim_task):
 
 # get_action_name {{{1
 def get_action_name(task):
```
tomprince: Can you update the name of this too.

escapewindow: Done; also fixed tests, added coverage, and added a 16.0.0 changeset.

escapewindow:
@JohanLorenzo , now you should be able to retrigger relpro tasks :)

PR #259, opened 2018-09-14
Handle hook and task actions via callback name.

tomprince:


scriptworker/cot/verify.py
```diff
@@ -1121,10 +1121,20 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
     for defn in all_actions:
-        if defn.get('hookPayload', {}).get('decision', {}).get('action', {}).get('cb_name') == action_name:
-            return defn
+        if defn.get('kind') == 'hook':
+            if defn.get('hookPayload', {}).get('decision', {}).get('action', {}).get('cb_name') == callback_name:
+                return defn
+        elif defn.get('kind') == 'task':
+            if defn.get('task', {}).get('$let', {}).get('action', {}).get('cb_name') == callback_name:
+                return defn
+        else:
```
escapewindow: I'm not sure we need to throw here; an oddly-formatted action doesn't matter to us as long as we find a definition with the right `cb_name`. However, as long as esr60 and release follow these rules, we should be fine keeping this block. The mobile repos may or may not follow these rules; if not, the fix is probably making taskgraph a standalone module.

scriptworker/test/test_cot_verify.py
```diff
@@ -1176,13 +1186,113 @@ def test_verify_link_in_task_graph_exception(chain, decision_link, build_link, i
...
+    link.task = {
+        'taskGroupId': decision_task_id,
+        'provisionerId': 'test-provisioner',
+        'schedulerId': 'tutorial-scheduler',
+        'workerType': 'docker-worker',
+        'scopes': [],
+        'payload': {
+            'image': 'test-image',
+            'env': {
+                'ACTION_CALLBACK': 'non-existant',
```
escapewindow: nit sp: nonexistent ? https://www.dictionary.com/misspelling?term=nonexistant&s=t

scriptworker/cot/verify.py
tomprince: I'd rather be conservative and bail out if we find something we don't know how to interpret. The [spec](https://docs.taskcluster.net/docs/manual/using/actions/spec) only talks about the two kinds here, so I don't think that other products or branches are going to have other kinds. When support is added for something else, we'll can figure out how to support them.

PR #264, opened 2018-10-12
Bug 1486089 - Support "github-release" tasks_for and cron of mobile

JohanLorenzo:
I tested `verify_cot` against a signing task scheduled by:
* a cron task  https://tools.taskcluster.net/groups/fj-ZwGVcT3-eBCTJElE0KQ/tasks/TosAq2dRRI2yxw221ECStQ/details
* and a release one https://tools.taskcluster.net/groups/RfhACY0wTCGdSNu7dLwpnw/tasks/YY2vQg2YSMq97NWsGDJIhw/details

with the following command:
``` sh
verify_cot --task-type signing --cleanup --cot-product mobile -- ${signingTaskId}
```

I didn't run the script against real worker, so I made these adjustments locally: https://gist.github.com/JohanLorenzo/06d11681c400f202dc37baf584a707d6

Both scenario ran fine against the current diff. You can see what `.taskcluster.yml` looks like here: https://github.com/mozilla-mobile/focus-android/compare/master...JohanLorenzo:full-chain-of-trust

I made the choice of not defining new context parameters in the case of cron. This mean, a cron context spoofs a github-release one. This makes the code simpler on chainOfTrust and Firefox Focus already runs the same script in both cases. Please let me know if this choice is too edgy.

### DIFF
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -31,7 +31,7 @@ from scriptworker.utils import (
 
 log = logging.getLogger(__name__)
 
-KNOWN_TASKS_FOR = ('hg-push', 'cron', 'action')
+KNOWN_TASKS_FOR = ('hg-push', 'cron', 'action', 'github-release')
 REPO_SCOPE_REGEX = re.compile("^assume:repo:[^:]+:action:[^:]+$")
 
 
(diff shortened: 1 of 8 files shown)


escapewindow:
> I made the choice of not defining new context parameters in the case of cron. This mean, a cron context spoofs a github-release one. This makes the code simpler on chainOfTrust and Firefox Focus already runs the same script in both cases. Please let me know if this choice is too edgy.

I think we can live with this for now, and when we make taskgraph-as-a-standalone-module a thing, we can revisit.

scriptworker/task.py
```diff
@@ -181,6 +181,40 @@ def get_revision(task, source_env_prefix):
     return revision
 
 
+def get_branch(task, source_env_prefix):
+    """Get the git branch on top of which the graph was made.
```
escapewindow: Hm. Is this git only? (It looks like it, hg doesn't have `_HEAD_BRANCH`.) I'm waffling about naming this `get_git_branch` or removing the `git` reference in the docstring, since hg may have `_HEAD_BRANCH` at some point. Since I can't decide myself, this is optional.

JohanLorenzo: It can be hg too. I removed the reference to git in the docstring

PR #265, opened 2018-10-12
fix verify_cot for taskcluster>=5.0.0

escapewindow:
By populating `context.config` before `context.credentials`, we have a
`taskcluster_root_url` defined for `context.queue`.


escapewindow:
@JohanLorenzo I noticed this while testing your PR. Should help fix if/when you upgrade to taskcluster>=5.0.0

PR #267, opened 2018-10-15
Bug 1486089 - part 2: Whitelist android-components and reference-browser repos

JohanLorenzo:
Found while testing https://github.com/mozilla-mobile/android-components/pull/1049

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -206,6 +206,8 @@ DEFAULT_CONFIG = frozendict({
                 "netlocs": ("github.com", ),
                 "path_regexes": (
                     r"^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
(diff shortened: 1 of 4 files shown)


scriptworker/constants.py
```diff
@@ -206,6 +206,7 @@
                 "netlocs": ("github.com", ),
                 "path_regexes": (
                     r"^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
+                    r"^(?P<path>/mozilla-mobile/android-components)(/|.git|$)",
```
MihaiTabara: Should we add `reference-browser` in advance for the work that's upcoming in a few weeks? https://github.com/mozilla-mobile/reference-browser/

JohanLorenzo: Good thinking, done.

MihaiTabara:
@coveralls Better late than never, thanks for reaching back :)

PR #269, opened 2018-11-02
Pip compile multi

escapewindow:
This PR creates a simple way to use docker to pin python dependencies. A large problem with the dephash-on-laptop approach is we could potentially pull in the wrong wheel if the laptop environment doesn't match the target environment (py36 vs py37, mac vs linux, etc).

- rename `helper_scripts/` to `scripts/`
- create a `requirements/` directory
- rename:
```
    requirements-dev.txt -> requirements/base.in
    requirements-prod.txt -> requirements/base.txt
    requirements-test-dev.txt -> requirements/test.in
    requirements-test-prod.txt -> requirements/test.txt
```
- add a `scripts/pip` helper script. I'm open to suggestions about the name.

@Callek I know you have used pip-compile-multi. @tomprince you've given me helpful pointers about dockerfiles before. Thoughts?


escapewindow:
@tomprince Do you want to take a look? Or should I treat @Callek's comments as an r+?

scripts/pip
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -e
+# Use docker + pip-compile-multi to pin scriptworker python requirements
+TOP_DIR="$(dirname "$( cd "$(dirname "$0")" && pwd -P)")"
+PIP_DOCKER_TAG="${PIP_DOCKER_TAG:-pip}"
```
tomprince: I think this should be something like `mozillareleases/scriptworker-pip-compile:${PY_DOT_VERSION}`, since the namespace is global to the machine?

scripts/pip
```diff
@@ -0,0 +1,37 @@
...
+    exit 1
+fi
+
+case $1 in
+    build)
+        docker build -t "$PIP_DOCKER_TAG" "$TOP_DIR/requirements/."
+        ;;
+    pin)
+        docker run -i -v "$TOP_DIR/requirements":/requirements "$PIP_DOCKER_TAG" \
+            pip-compile-multi -d /requirements/. --upgrade -i in -o txt -g base -g test
```
tomprince: nit: Is the trailing `.` in `/requirements/.` required?

escapewindow: It isn't. Fixed.

scripts/pip
escapewindow: Makes sense. I'm going to figure out how to push to mozillareleases.

escapewindow:
>     how does this interact with pyup?

I believe Callek has some pyup/pip-compile-multi interactions in addonscript; we should follow suit in scriptworker after this lands.

>     is it possible to test the pinning in CI?

I think so. I'm picturing that we'll bump deb repos in CI along with pyup/python-bumping, and we can run tests and generate docker images in CI as well. This is just the first step towards that future.

Thanks both for the review+feedback!

PR #271, opened 2018-11-09
Bug 1512631 - Enable mozilla-mobile staging release from Github forks

JohanLorenzo:
Discovered in https://github.com/mozilla-mobile/android-components/pull/1284#pullrequestreview-173430757

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -9,6 +9,7 @@ Attributes:
 """
 from frozendict import frozendict
 import os
+import re
 
 # STATUSES {{{1
 STATUSES = {
@@ -21,6 +22,14 @@ STATUSES = {
     'superseded': 6,
     'intermittent-task': 7,
 }
+
+_ALLOWED_MOBILE_GITHUB_OWNERS = (
+    'mozilla-mobile',
+    # Owners below are allowed to run staging releases
+    'JohanLorenzo',
+    'MihaiTabara',
+    'mitchhentges',
+)
 
 # DEFAULT_CONFIG {{{1
 # When making changes to DEFAULT_CONFIG that may be of interest to scriptworker
@@ -162,9 +171,13 @@ DEFAULT_CONFIG = frozendict({
         "gecko-1-decision",
         "gecko-2-decision",
         "gecko-3-decision",
-        # gecko-focus is for mozilla-mobile releases. It's named this way because it's a worker type
-        # but using the gecko ChainOfTrust keys. See bug 1455290 for more details.
+        # gecko-focus was for mozilla-mobile releases (bug 1455290) for more details.
+        # TODO: Remove it once not used anymore
         "gecko-focus",
+        "mobile-1-decision",
+        # We haven't had the need for mobile-2-decision yet
+        # https://bugzilla.mozilla.org/show_bug.cgi?id=1512631#c6
+        "mobile-3-decision",
     ),
 
     # docker-image cot
@@ -204,11 +217,16 @@ DEFAULT_CONFIG = frozendict({
             'mobile': (frozendict({
                 "schemes": ("https", "ssh", ),
                 "netlocs": ("github.com", ),
-                "path_regexes": (
-                    r"^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/android-components)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/reference-browser)(/|.git|$)",
-                ),
+                "path_regexes": tuple([
+                    # XXX We allow Github forks to run staging releases. Please make sure to
+                    # restrict production scopes to the origin repo (with cot_restricted_scopes)!
+                    r"^(?P<path>/{repo_owner}/{repo_name})(/|.git|$)".format(
+                        repo_owner=re.escape(username),
+                        repo_name=re.escape(repo_name)
+                    )
+                    for username in _ALLOWED_MOBILE_GITHUB_OWNERS
+                    for repo_name in ('android-components', 'focus-android', 'reference-browser')
+                ]),
             }),),
         }),
     },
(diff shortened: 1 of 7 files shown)


scriptworker/constants.py
```diff
@@ -165,6 +165,8 @@
         # gecko-focus is for mozilla-mobile releases. It's named this way because it's a worker type
         # but using the gecko ChainOfTrust keys. See bug 1455290 for more details.
         "gecko-focus",
+        # staging releases can run on the regular github-worker.
+        "github-worker",
```
JohanLorenzo: To be honest, I'm concerned about this line. If I remember correctly, "gecko-1-decision" was created because we want segregation between taskcluster workers that anybody can use and workers only level-1 people can use. I wonder if we should do something similar for github.

scriptworker/constants.py
```diff
@@ -205,9 +207,11 @@
                 "schemes": ("https", "ssh", ),
                 "netlocs": ("github.com", ),
                 "path_regexes": (
-                    r"^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/android-components)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/reference-browser)(/|.git|$)",
+                    # XXX We allow Github forks to run staging releases. Please make sure to
+                    # restrict production scopes to the origin repo (with cot_restricted_scopes)!
+                    r"^(?P<path>/[A-Za-z0-9-]+/android-components)(/|.git|$)",
```
JohanLorenzo: I checked: github "Username may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen". This regex is a little bit more permissive.

scriptworker/constants.py
```diff
@@ -354,9 +360,12 @@
                 ),
             }),
             'mobile': frozendict({
-                'focus-repo': (
+                'focus-origin-repo': (
                     '/mozilla-mobile/focus-android',
                 ),
+                'android-components-origin-repo': (
+                    '/mozilla-mobile/android-components',
+                ),
```
MihaiTabara: Should we prep things in advance by adding the`reference-browser` counterpart too, given upcoming work?

JohanLorenzo: I'd prefer @mitchhentges to handle this change. This looks like a good first patch in scriptworker, to me.

MihaiTabara: Good point.

JohanLorenzo: By the way, that makes me wonder if shouldn't remove https://github.com/mozilla-releng/scriptworker/pull/271/files#diff-4f43b126671fd39d1f9e922fbb72b839R214 too. That may be confusing during the development of `reference-browser`.

MihaiTabara: In the interest of good-first-patch, that makes sense.

JohanLorenzo: Done

scriptworker/constants.py
tomprince: I think maybe we want to have a configuration like `verify_cot_signature` that conditionally (i.e. for staging workers) either doesn't check this list, or checks against a more permissive

scriptworker/constants.py
tomprince: Does it make sense to allow any github repository? Forks don't necessarily need to use the same name as the base repository.

scriptworker/constants.py
```diff
@@ -354,9 +359,12 @@
                 ),
             }),
             'mobile': frozendict({
-                'focus-repo': (
+                'focus-origin-repo': (
```
tomprince: I'm not sure that adding `-origin-` doesn't make thing clearer.

scriptworker/constants.py
JohanLorenzo: Discussed with Aki over IRC. One clean solution would be to create `mobile-{1..3}-decision worker types. This way, we don't allow `github-worker` anymore. We're using a more trusted subset of workers. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1512631

scriptworker/constants.py
JohanLorenzo: That's a great point. Allowing all github reposity sounds risky for now. @escapewindow suggested to limit the list down to known users (who are willing to make staging releases). Let's see how this goes.

scriptworker/constants.py
JohanLorenzo: Okay. Removed.

scriptworker/constants.py
```diff
@@ -205,9 +217,14 @@
...
-                    r"^(?P<path>/mozilla-mobile/focus-android)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/android-components)(/|.git|$)",
-                    r"^(?P<path>/mozilla-mobile/reference-browser)(/|.git|$)",
+                    # XXX We allow Github forks to run staging releases. Please make sure to
+                    # restrict production scopes to the origin repo (with cot_restricted_scopes)!
+                    r"^(?P<path>/({})/android-components)(/|.git|$)".format(
+                        '|'.join(_ALLOWED_MOBILE_GITHUB_OWNERS)
+                    ),
+                    r"^(?P<path>/({})/focus-android)(/|.git|$)".format(
+                        '|'.join(_ALLOWED_MOBILE_GITHUB_OWNERS)
```
tomprince: In an abundance of caution, I think we should run these through `map(re.escape, ...)`.

JohanLorenzo:
The latest changes worked in staging: https://tools.taskcluster.net/groups/X-d_JJWpSoGXJ6QwzLs7OQ/tasks/CTcr1xXOTvC9vPywK4gThg/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L72

scriptworker/constants.py
JohanLorenzo: https://github.com/mozilla-releng/scriptworker/commit/f45ea14a383b55987f3bfa6562ecfa038f1ee728 ended up being live for quite a long time. Reference browser has now nightlies. I'm putting reference browser back.

scriptworker/constants.py
JohanLorenzo: Done

JohanLorenzo:
In the latest push, I:
* rebased on top of master
* expanded the code coverage to ensure constants.py has some mobile specific tests
* and I bumped the version

scriptworker/constants.py
```diff
@@ -22,6 +23,17 @@
     'intermittent-task': 7,
 }
 
+_ALLOWED_MOBILE_GITHUB_OWNERS_PATTERN = '|'.join([
```
mitchhentges: IMHO, we should have this data in its raw-est format up here: when we're adding new users, this is the only place we want to have to change. We should move the regex-y bit (escaping, joining with `|`) down to where it's needed.
Then, if we are using these names down below in a non-regex-y way, we don't need to parse out the `|` :smile:

JohanLorenzo: True. That variable dated back before I turned the mobile regexes into a list-comprehension. It doesn't provide much value anymore. I implemented your suggestion. Thank for spotting it!

PR #272, opened 2018-11-13
Add nagios info

nthomas-mozilla:
I'd like to hold off merging this until we know if the new type docs are correct via work in [bug 1484880](https://bugzilla.mozilla.org/show_bug.cgi?id=1484880). Comments welcome in the meantime.


escapewindow:
lgtm; thank you!

nthomas-mozilla:
We landed the change which validates these docs so I'm ready to merge.

The test failures after merging master seem to be unrelated to the changes here. FWIW, we use filterwarnings to suppress DeprecationWarnings in our Balrog deps, [eg](https://github.com/mozilla/balrog/blob/master/tox.ini#L26).

PR #278, opened 2018-11-29
Fix json-e context for mobile projects

JohanLorenzo:


scriptworker/cot/verify.py
```diff
@@ -1110,21 +1110,26 @@ def _get_additional_github_releases_jsone_context(parent_link, decision_link):
         jsone_context.update(
             _get_additional_github_releases_jsone_context(parent_link, decision_link)
         )
+
```
Callek: I had to look this up, since I thought cot_product would say 'mobile' for Firefox-for-Android (from hg), I was wrong.

Callek:
I'd love increased coverage as well, but I'm not sure this is worth blocking on better coverage.

scriptworker/cot/verify.py
JohanLorenzo: I'm sorry for the confusion. We currently have 3 different `cot_product`s:

* `firefox` -> for all mozilla-* hg repos
* `comm` -> for comm-* hg repos
* `mobile` -> for all new mobile projects that are on GitHub.

I'll try to make this clearer in the tests.

JohanLorenzo:
> I'd love increased coverage as well, but I'm not sure this is worth blocking on better coverage.

I expanded the coverage, but it doesn't provide enough context yet, to me. I couldn't get what I wanted in a reasonable amount of time, so I just ended up putting coverage back to 100% 

Thank you very much for the fast review!

PR #282, opened 2018-12-11
No concurrent tests

escapewindow:
Back-burnering the dockerfile for now, since the /dev/random -> /dev/urandom fix doesn't stick.

This fixes a bunch of intermittents and also speeds up travis testing.


escapewindow:
Thanks! Looks like we'll have some followups.

PR #283, opened 2018-12-11
docker entrypoint for unittests + gpg testing

escapewindow:


escapewindow:
@nthomas-mozilla do you feel comfy reviewing this? Copy/edited the balrog run.sh

escapewindow:
Yeah. I'm waffling between sharing the same image between prod and dev, in which case we want to only run the dev commands in the script, or having separate images, in which case we can run the dev commands in the Dockerfile.

Issue #277, opened 2018-11-29
CoT regression on Firefox Focus: Cannot download HG pushlog


JohanLorenzo:
2 days ago, scriptworker 17.0.0 was deployed to all workers (https://github.com/mozilla-releng/build-puppet/pull/311), including the mobile ones. This caused [this CoT regression](https://tools.taskcluster.net/groups/YDFdZBm_QuC5ENF0uNIFyA/tasks/d-gkFlVKQSW0cJDzF0x-LQ/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L125) on Focus Nightly:

```
2018-11-29T14:08:16  WARNING - retry_async: download_file: too many retries!
2018-11-29T14:08:16 CRITICAL - Chain of Trust verification error!
Traceback (most recent call last):
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 1406, in verify_parent_task
    await verify_parent_task_definition(chain, link)
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 1266, in verify_parent_task_definition
    chain, parent_link, decision_link, tasks_for
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 1232, in get_jsone_context_and_template
    chain, parent_link, decision_link, tasks_for
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 1126, in populate_jsone_context
    await _get_additional_cron_jsone_context(parent_link, decision_link)
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 1059, in _get_additional_cron_jsone_context
    pushinfo = await _get_additional_hgpush_info(decision_link)
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 994, in _get_additional_hgpush_info
    pushlog_info = await get_pushlog_info(decision_link)
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/cot/verify.py", line 950, in get_pushlog_info
    context, pushlog_url, file_path, overwrite=False
  File "/builds/scriptworker/lib/python3.6/site-packages/scriptworker/utils.py", line 547, in load_json_or_yaml_from_url
    retry_exceptions=(DownloadError, aiohttp.ClientError),
[...]

PR #284, opened 2018-12-12
Json-e followup

JohanLorenzo:
Follows #277 and #278 up

### DIFF
--- a/scriptworker/test/__init__.py
+++ b/scriptworker/test/__init__.py
@@ -269,27 +269,45 @@ async def _close_session(obj):
 
 
 @pytest.mark.asyncio
-@pytest.yield_fixture(scope='function', params=['firefox'])
-async def rw_context(request, event_loop):
+@pytest.yield_fixture(scope='function')
+async def rw_context(event_loop):
     with tempfile.TemporaryDirectory() as tmp:
-        config = get_unfrozen_copy(DEFAULT_CONFIG)
-        config['cot_product'] = request.param
-        context = Context()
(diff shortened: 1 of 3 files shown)


scriptworker/test/__init__.py
```diff
@@ -269,27 +269,45 @@ def tmpdir2():
...
-        context.event_loop = event_loop
+        context = _craft_rw_context(tmp, event_loop, cot_product='mobile')
         yield context
-        await _close_session(context)
-        await _close_session(context.queue)
-        await _close_session(context.temp_queue)
+        await _close_context(context)
+
+
+def _craft_rw_context(tmp, event_loop, cot_product):
```
escapewindow: Smart. I was a little concerned that changing the Context creation coroutine to a synchronous function might give us more `aiohttp.ClientSession` warnings. If we did start getting these warnings, the following may help address:

* changing `_craft_rw_context` to a coroutine
* defining `context.event_loop` before creating `context.config`, since we [use the `event_loop` to create the session](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/context.py#L118).

However, I don't see any of these warnings in Travis, so we may be good. Still, if we went back to pytest-xdist or other concurrent tests, defining `event_loop` before the config may still be a good fix to make.

JohanLorenzo: Cool! Thanks for the context. I'll keep an eye out for the warnings. In the meantime, I'm landing this patch as is.

PR #286, opened 2018-12-19
bug 1485680 - convert release-promotion to an action hook

escapewindow:
This passes `verify-cot --task-type action --cleanup -- $TASKID` for the following tasks:

```
FBoKTObySZeq25MnvJcSAA  # try relpro hook
E93EolXNSRuV2gOy7yMZEA  # try retrigger hook
V4Zls-rmShmRubT_vyf_Kw  # beta retrigger hook
fmqrnpNNSdOmtX5kG2oTyQ  # beta relpro task
LyJAxMeaRFid175eXWGdBw  # esr relpro task
```
More context and discussion in #256.

@tomprince @djmitche how does this look?


scriptworker/cot/verify.py
```diff
@@ -1204,10 +1258,27 @@ def _get_action_from_actions_json(all_actions, callback_name):
...
+                action_defn, jsone_context, parent_link
+            ),
+            'taskId': parent_link.task_id,
+            'now': jsone_context['now'],
+            'as_slugid': jsone_context['as_slugid'],
+        }
     else:
+        # We know this block is `task` because we've already tested for action kind in
+        # `_get_action_from_actions_json`.
+        # XXX Get rid of this block when all actions are hooks
```
tomprince: I'd rather explicitly check for `kind = 'task'` here, and error if the kind is neither.

escapewindow: I had that, and removed it, because it was unreachable code. (we check `kind` validity earlier in the function, so i was never able to reach the else)

escapewindow: - I believe in [100% test coverage, without `#pragma: nocover`](https://escapewindow.dreamwidth.org/247980.html)
- We are going to remove support for this else block when all actions are hooks in the near future
- I disagree that not having the explicit check makes it hard to reason about the code, especially since there are comments in the block which explicitly say what the block does

Do you think the explicit check blocks landing?

tomprince: > I disagree that not having the explicit check makes it hard to reason about the code, especially since there are comments in the block which explicitly say what the block does

The fact that we are having this discussion is evidence that it is hard to reason about. I don't see anything in the diff that ensure that other case can't occur here, and even if I expand the entire function, it isn't obvious where the checks is. There is a comment, but to verify the correctness of the code, I still need go and look at that it is doing what this code expects it is.

> I had that, and removed it, because it was unreachable code. 

I find the fact that you also originally thought there should be a check here argues strongly that we should keep the check.

escapewindow: > > I disagree that not having the explicit check makes it hard to reason about the code, especially since there are comments in the block which explicitly say what the block does
> 
> The fact that we are having this discussion is evidence that it is hard to reason about. I don't see anything in the diff that ensure that other case can't occur here, and even if I expand the entire function, it isn't obvious where the checks is. There is a comment, but to verify the correctness of the code, I still need go and look at that it is doing what this code expects it is.

Does this block landing, especially since we're going to remove `kind: task` support once we have relpro action hooks?

> > I had that, and removed it, because it was unreachable code.
> 
> I find the fact that you also originally thought there should be a check here argues strongly that we should keep the check.

I originally added it due to one of your earlier review comments, so I still only see one person pushing for this. Do you consider this blocking?

Issue #287, opened 2018-12-21
schema_file should not be a recommended part of config


mitchhentges:
See the [original pushapkscript ticket](https://github.com/mozilla-releng/pushapkscript/issues/59)

The `schema_file` defines what is a valid task structure for the _current_ version of each `$script`. Being able to configure this schema file means that our code can't trust that validation is ensuring incoming data is in the right format. As source code changes to expect a different schema, the `schema_file` should be updated in tandem, and should be expected to be applied in tandem, too.

So, we should statically refer to the version of `schema_file` that ships with each `$script`. This has two benefits:
* Less variability at runtime
* Less configuration for end-users

-----

This ticket requires the following:
1. ~~`scriptworker` should be able to have the schema handed directly from the dependent scripts via arguments~~
2. ~~Each `$script` should be updated to provide `schema_file` directly, and should remove `schema_file` from the configuration/context~~
3. ~~`scriptworker` should no longer be able to pull `schema_file` from configuration~~

[EDIT](https://github.com/mozilla-releng/scriptworker/issues/287#issuecomment-449768770):
* All scripts will provide `schema_file` as part of the "default config" passed to `scriptworker`.
* All scripts will no longer advertise `schema_file` as an editable property in the config file (though, at a technical level, it still will be)
* Puppet configuration will be updated to no-longer provide `schema_file`
* No changes to the `scriptworker` API

escapewindow:
@mitchhentges , I see a lot of work around changing script configs here and elsewhere. You're aware that releng controls puppet which controls script_config.json, correct?

mitchhentges:
Affirmative - once all the scripts have been updated to no longer require having the schema provided in the config, I'll update the `script_config.json` for each script in puppet :+1:

escapewindow:
I mean removing the ability to specify things in script_config.json and scriptworker.yaml does very little to improve security, since we control all the pieces. Let's make sure the other rationales are strong enough for these changes to be worth it.

mitchhentges:
The goal of this change is not to enhance security, it's to improve confidence that our tasks are properly validated against the running version of `$script`

JohanLorenzo:
Looks like the agreement happened in https://github.com/mozilla-releng/signingscript/pull/99#issuecomment-450679904. I'll good with the plan too! :slightly_smiling_face:

escapewindow:
[...]

PR #288, opened 2018-12-21
Adds "schema" and "task" parameters to validate_task_schema

mitchhentges:
Step 1 of #287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Upcoming]
+### Changed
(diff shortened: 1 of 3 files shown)


mitchhentges:
As much as I'd like to push file operations up and out of logic that should (IMHO) be about validating data structures, this change can be deferred - it's not necessary for #287.
Even after the work of deprecating/updating all usages, we're in the unfortunately scenario where all usages _do_ pull from a file, so they'll all "look" more unclean since they'll all be loading-from-file right before calling `validate_task_schema`, instead of having that "load-from-file" logic shared.

There's a few functions of `scriptworker` that take larger data structures, then peel out the data they need, rather than just having the needed-data directly as parameters. I'd like to address this, but it isn't pressing for the more-beneficial change of removing `schema_file` from config

PR #293, opened 2019-01-04
bug 1492617 - Add ECDSA signature support to cot

### LINKED ISSUE #124: bug 1492617 - add ecdsa support and deprecate gpg support
GPG 2.0.x will EOL EOY 2017.

https://www.gnupg.org/download/index.en.html

"GnuPG 2.0 is an older branch of GnuPG. This branch will reach end-of-life on 2017-12-31."

GPG 2.2 is the new stable branch. https://lists.gnu.org/archive/html/info-gnu/2017-08/msg00012.html

escapewindow:
This PR adds ECDSA CoT signature generation and verification support to scriptworker; GPG support is deprecated.

It also:
- adds a new artifact, `public/chain-of-trust.json`
- adds a new artifact, `public/chain-of-trust.json.sig`, if `sign_chain_of_trust` is `True`. We verify these two files if they exist, and fall back to gpg verification.
- makes `public/chainOfTrust.json.asc` a mandatory download during cot verification.
- adds `read_from_file` and `write_to_file` helper functions
- fixes pytest-random-order

Integration production cot verification tests are slower; I'm wondering if we're retrying the `chain-of-trust.json{,.sig}` downloads even if they're 404s. Going to poke at this. I also want to try running this, maybe on signing-linux-dev1, and make sure it behaves like I think it does.

Rollout plan:

- [ ] ecdsa support across all worker implementations
  - [ ] land scriptworker ecdsa cot generation + verification
  - [ ] add generic-worker ecdsa cot generation support
  - [ ] add docker-worker ecdsa cot generation support
  - [ ] add docs for ecdsa key maintenance and rotation
- [ ] audit: cot should be using ecdsa across all trees
  - [ ] point in-tree code at `chain-of-trust.json` rather than `chainOfTrust.json.asc`
- [ ] kill gpg support
  - [ ] remove gpg support from scriptworker
  - [ ] remove gpg support from generic-worker
  - [ ] remove gpg support from docker-worker
  - [ ] remove gpg support from puppet
  - [ ] remove gpg nagios checks
  - [ ] retire cot-gpg-keys repo

Resolves #124 .


tomprince:
I haven't yet looked at the details of how this is implemented, by I'm curious what the reasoning for choosing ecdsa as the signature algorithm is. Reading [this](https://latacora.micro.blog/2018/04/03/cryptographic-right-answers.html) suggests ed25519 is probably what we should be picking (probably using [pynacl](https://pynacl.readthedocs.io/en/stable/)). Maybe @Alex can chime in.

alex:
Apologies since I don't have a ton of context here, but general notes on signature schemes:

- Yes, for new protocols ed25519 (or ed448) is preferred to ECDSA. This is because ed25519 is simpler to implement, has fewer edge cases, (I believe is faster, but I can't find any citations), and most importantly, has fewer catastrophic failure modes -- `k` reuse in ECDSA permanently compromises the private key, which can't happen in ed25519, and ed25519 will survive a significant break in the hash function (SHA512), ECDSA will not
- If you're implementing something for interop with pre-existing protocols you might be stuck with ECDSA, which is unfortunate, but not the end of the world.
- For ed25519 you can get it either via pynacl or soon via pyca/cryptography (https://github.com/pyca/cryptography/pull/4114)
- If you are going to use ECDSA, I recommend getting it via pyca/cryptography, which will be faster and better audited than using the ecdsa library.

escapewindow:
It looks like there are ed25519 solutions for node and go, which is needed to add support to docker-worker and generic-worker.

jvehent:
`ed25519` support is still incomplete in a lot of crypto libraries which is why I still learn towards ECDSA, but if you only need implementation in one language that supports it, I agree with @alex ed25519 is the better choice.

escapewindow:
@alex do you know the state of the golang and node.js ed25519 implementations? or know someone that would?

alex:
For Go it's available in the x/crypto package: https://godoc.org/golang.org/x/crypto/ed25519

I don't know about node, sorry.

escapewindow:
`nacl.signing.SigningKey` doesn't appear to have the `.to_pem()` functionality that the `ecdsa.SigningKey` has.
Is there a preferred way of creating a portable text version of the signing key? I'm currently using

```
signing_key = nacl.signing.SigningKey.generate()
signing_key_string = base64.encodestring(bytes(signing_key)).strip().decode('utf-8')
assert base64.decodestring(signing_key_string.encode('utf-8')) == bytes(signing_key)
```

to test exporting and importing the plaintext key.

alex:
Yeah, just using a base64 version of the raw key is pretty common.

escapewindow:
I have a working version of ed25519 support in https://github.com/escapewindow/scriptworker/tree/ed25519 .

escapewindow:
Closing this in favor of #294 . If we find any blocking issues with ed25519 rollout, this is still here to fall back on.

PR #294, opened 2019-01-05
bug 1492617 - Add ed25519 signature support to cot

### LINKED ISSUE #124: bug 1492617 - add ecdsa support and deprecate gpg support

escapewindow:
This PR adds ed25519 CoT signature generation and verification support to scriptworker; GPG support is deprecated. If this lands, we should close #293.

It also:
- adds a new artifact, `public/chain-of-trust.json`
- adds a new artifact, `public/chain-of-trust.json.sig`, if `sign_chain_of_trust` is `True`. We verify these two files if they exist, and fall back to gpg verification.
- makes `public/chainOfTrust.json.asc` a mandatory download during cot verification.
- adds `read_from_file` and `write_to_file` helper functions
- fixes pytest-random-order

Integration production cot verification tests are slower; I'm wondering if we're retrying the `chain-of-trust.json{,.sig}` downloads even if they're 404s. Going to poke at this. I also want to try running this, maybe on signing-linux-dev1, and make sure it behaves like I think it does.

Rollout plan:

- [ ] ed25519 support across all worker implementations
  - [ ] land scriptworker ed25519 cot generation + verification
  - [ ] add generic-worker ed25519 cot generation support
  - [ ] add docker-worker ed25519 cot generation support
  - [ ] add docs for ed25519 key maintenance and rotation
- [ ] audit: cot should be using ed25519 across all trees
  - [ ] point in-tree code at `chain-of-trust.json` rather than `chainOfTrust.json.asc`
- [ ] kill gpg support
  - [ ] remove gpg support from scriptworker
  - [ ] remove gpg support from generic-worker
  - [ ] remove gpg support from docker-worker
  - [ ] remove gpg support from puppet
  - [ ] remove gpg nagios checks
  - [ ] retire cot-gpg-keys repo

Resolves #124 .


scriptworker.yaml.tmpl
```diff
@@ -93,6 +93,19 @@ gpg_path: gpg
 my_email: "scriptworker@example.com"
 
 
+#-----------------------------------------------------------------------------------------------
+# ed25519 settings.
+#-----------------------------------------------------------------------------------------------
+ed25519_private_key_path: /tmp/ed25519_privkey.pem
+ed25519_public_key_paths:
```
escapewindow: Hm. These public keys are tiny. For example, the scriptworker throwaway test public key is `KxYrV3XAJ3uOyAUX0Wcl1Oeu6GSMrI/5hOn39q8Lf0I=`. I wonder if we shouldn't just bake these into the config, rather than create plaintext files for each.

scripts/gen_ed25519_key.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+from __future__ import print_function
+import base64
+from nacl.signing import SigningKey, VerifyKey
+import sys
+
+
+def to_string(key):
+    """Convert an ed25519 key to a base64-encoded string."""
+    return base64.encodestring(bytes(key)).strip().decode('utf-8')
```
tomprince: https://pynacl.readthedocs.io/en/stable/encoding/ suggests that we don't need to implement these ourself.

scriptworker/cot/verify.py
```diff
@@ -803,36 +918,17 @@ def verify_cot_signatures(chain):
...
-            with open(path, "r") as fh:
-                contents = fh.read()
-        except OSError as exc:
-            if is_task_required_by_any_mandatory_artifact(chain, link.task_id):
-                raise CoTError("Can't read mandatory {}: {}!".format(path, str(exc)))
-            else:
-                log.warning("Could not read optional {}. Continuing. Error gotten: {}!".format(path, str(exc)))
-                continue
+        unsigned_path = link.get_artifact_full_path('chain-of-trust.json')
+        ed25519_signature_path = link.get_artifact_full_path('chain-of-trust.json.sig')
```
tomprince: Do we want to somehow encode the format of the signature? Either in the filename, or by having the signature file be structured somehow? @alex, is there any standard practice around this, for detached signatures?

scriptworker/ed25519.py
```diff
@@ -0,0 +1,87 @@
...
+    """Convert an ed25519 key to a base64-encoded string.
+
+    Args:
+        key (nacl.signing.SigningKey or nacl.signing.VerifyKey): the key to write to the file.
+
+    Returns:
+        str: the key representation as a str
+
+    """
+    return base64.encodebytes(bytes(key)).strip().decode('utf-8')
```
tomprince: As mentioned above, I think we can use pynacl functions for this.

scriptworker/utils.py
```diff
@@ -465,6 +465,73 @@ def load_json_or_yaml(string, is_path=False, file_type='json',
...
+        raise ScriptWorkerException("Unknown file_type {} not in {}!".format(file_type, FILE_TYPE_MAP))
+    try:
+        with open(path, FILE_TYPE_MAP[file_type]) as fh:
+            return fh.read()
+    except (OSError, FileNotFoundError) as exc:
+        message = "Can't read_from_file {}: {}".format(path, str(exc))
+        if exception is not None:
+            raise exception(message)
+        else:
+            log.warning(message)
```
tomprince: I don't see this case being used in production code, so I think we should always raise an exception. (And the one use in tests appears to expect the function to be infallible.

escapewindow:
Thanks! I'm currently working on a wip branch that has a few minor changes. I think I'll get things a bit smoother there, then:

- rebase on master and force-push. Github is now showing the diff in force-pushes, so leaving any other changes out of the first will help minimize the noise.
- add in changes from the wip branch, and address your review comments.

escapewindow:
It looks like `chain-of-trust.json.sig` contains the signature and the body, which is different from ecdsa... when verifying I can just `binary_json = verify_key.verify(sig_contents)`. I'm also seeing a bunch of 404 retries on missing `chain-of-trust.json{,.sig}` artifacts when verifying cot, which I want to address.

scriptworker/cot/verify.py
escapewindow: It looks like the signature file has the body in it.
That raises the question of whether we publish a single file (the signature+body) or two (unsigned, and signature+body). I do foresee some automation using the unsigned json directly. If we're able to roll out a tool that can verify the signature and return the json (or ignore the signature and return the json), we could publish one file.

scripts/gen_ed25519_key.py
escapewindow: Nice, thanks for the pointer!

scriptworker/constants.py
```diff
@@ -31,6 +31,16 @@
...
 
+_COT_VERSION_DESCRIPTIONS = frozendict({
+    1: 'Initial Chain of Trust implementation with GPG signatures',
+    2: 'Json-e verification support',
+    3: 'Action hook support',
+    4: 'Release promotion action hook support',
+    5: 'Initial ed25519 support; GPG deprecated',
+    6: 'Drop support for non-hook actions',
+    7: 'Drop support for GPG',
+})
```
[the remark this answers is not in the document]
escapewindow: We have `min_cot_version` which corresponds to the keys. The values are human documentation. I can remove this, but I'd like this documented somewhere. Do you think this works, or do you think it belongs elsewhere?

scriptworker/cot/verify.py
```diff
@@ -601,28 +605,45 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
+        mandatory_urls.append(gpg_url)
+        unsigned_url = get_artifact_url(chain.context, task_id, 'public/chain-of-trust.json')
+        optional_urls.append(unsigned_url)
+        if chain.context.config['verify_cot_signature']:
+            optional_urls.append(
+                get_artifact_url(chain.context, task_id, 'public/chain-of-trust.json.sig')
             )
-        )
 
-        if is_task_required_by_any_mandatory_artifact(chain, task_id):
```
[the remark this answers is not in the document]
escapewindow: Removed. This means this will be a major version bump, which is probably a good thing anyway.

scriptworker/constants.py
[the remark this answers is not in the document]
escapewindow: 4 just landed. This is 5. 6 and 7 aren't ready to land yet, and yes, they could switch order depending on when they land, but this is the current plan. I sometimes reference cot versions by number; this is the map from number to what happened in that version.

scriptworker.yaml.tmpl
escapewindow: I'm currently staying with files, because we're less likely to log the contents of a private key file than the running config. When we're verifying these signatures in more places than just scriptworker, it might make sense to publish some json or something that contains all the valid pubkeys.

scriptworker/cot/generate.py
```diff
@@ -111,9 +118,14 @@ def generate_cot(context, path=None):
     )
     validate_json_schema(body, schema, name="chain of trust")
     body = format_json(body)
-    path = path or os.path.join(context.config['artifact_dir'], "public", "chainOfTrust.json.asc")
+    asc_path = path or os.path.join(context.config['artifact_dir'], "public", "chainOfTrust.json.asc")
+    unsigned_path = os.path.join(os.path.dirname(asc_path), 'chain-of-trust.json')
```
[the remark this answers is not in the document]
escapewindow: +1. It looks like we might want to keep it since we use a non-tmpdir `artifact_dir` in some tests, but I could change `path` to `parent_path`, especially since we already have a good reason to bump the major version.

scriptworker/utils.py
escapewindow: Fixed.

scriptworker/cot/generate.py
```diff
@@ -111,9 +118,14 @@ def generate_cot(context, path=None):
...
+    write_to_file(unsigned_path, body)
     if context.config['sign_chain_of_trust']:
+        ed25519_signature_path = "{}.sig".format(unsigned_path)
```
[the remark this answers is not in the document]
escapewindow: Easy enough.

scriptworker/constants.py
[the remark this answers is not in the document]
escapewindow: It's easy enough to move this to the docs though.

scriptworker.yaml.tmpl
escapewindow: Per IRC, we're leaving private keys in files and putting public keys in config.
In the future, we may move the public keys into their own json file if/when we share the public keys across multiple points of verification.

scriptworker/cot/verify.py
escapewindow: Ah, the signature file has the body in it if I use `signing_key.sign(binary_contents)`, and is a detached signature if I use `signing_key.sign(binary_contents).signature`.

I'm open to standard practices.

escapewindow:
> Thanks! I'm currently working on a wip branch that has a few minor changes. I think I'll get things a bit smoother there, then:
> 
>     * rebase on master and force-push. Github is now showing the diff in force-pushes, so leaving any other changes out of the first will help minimize the noise.
> 
>     * add in changes from the wip branch, and address your review comments.

These are both done.

> It looks like `chain-of-trust.json.sig` contains the signature and the body, which is different from ecdsa... when verifying I can just `binary_json = verify_key.verify(sig_contents)`. I'm also seeing a bunch of 404 retries on missing `chain-of-trust.json{,.sig}` artifacts when verifying cot, which I want to address.

I addressed the 404s in #300. I switched over to detached signatures, though I'm happy to switch back to attached signatures if we have a good reason to. I still want to at least get a working generic-worker implementation, possibly also docker-worker. before fully releasing ed25519 support, to avoid too much churn. I'll focus on those unless there are other changes needed here.

scriptworker/constants.py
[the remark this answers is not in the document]
tomprince: I was looking at a `chain-of-trust.json` file, and I noticed that that has a version number in it, which still appears to be v1, so that is perhaps a source of confusion around what CoT versioning is.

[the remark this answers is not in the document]
escapewindow: The json schema is still backwards compatible with the first cot artifact schemas, hence v1. However, the logic behind how to verify the chain has changed drastically: from guessing at whether the task looks like it might have come from the tree, to using json-e to actually regenerate the task, to supporting action hooks, to supporting action hooks for relpro. Not all of these versions are backwards compatible.

scriptworker/cot/generate.py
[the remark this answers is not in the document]
escapewindow: This appears to have changed the filename to `chain-of-trust.sig` instead of `chain-of-trust.json.sig`, which isn't what we want. I'm probably going to revert this change.

scriptworker/test/test_utils.py
```diff
@@ -312,6 +312,53 @@ def test_rm_dir():
...
+}"""
+), (
+    'binary', b'asdfasdf', b'asdfasdf'
+), (
+    'text', 'asdfasdf', 'asdfasdf'
+)))
+def test_write_to_file(tmpdir, file_type, contents, expected):
+    path = os.path.join(tmpdir, 'foo')
+    utils.write_to_file(path, contents, file_type=file_type)
+    mode = 'rb' if file_type == 'binary' else 'r'
```
tomprince: I think this should always be binary, to test exactly what got written. And maybe there should be some non-ascii data, and non-UTF-8 binary data.

scriptworker/test/test_utils.py
```diff
@@ -312,6 +312,53 @@ def test_rm_dir():
...
+    'text', 'asdfasdf', 'asdfasdf', ScriptWorkerException, False
+), (
+    'text', None, None, Exception, True
+), (
+    'bad_file_type', 'asdfasdf', 'asdfasdf', Exception, True
+)))
+def test_read_from_file(tmpdir, file_type, contents, expected, exception, raises):
+    path = os.path.join(tmpdir, 'foo')
+    if contents is not None:
+        mode = 'wb' if file_type == 'binary' else 'w'
```
tomprince: And similarly here.

scriptworker/test/test_utils.py
escapewindow: > I think this should always be binary, to test exactly what got written.

How do we verify that it was written as non-binary, then?

scriptworker/constants.py
```diff
@@ -31,6 +30,8 @@
     'mitchhentges',
 )
 
+_SW_LIB_DIR = os.path.dirname(__file__)
```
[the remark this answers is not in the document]
escapewindow: Removed.

scriptworker/ed25519.py
```diff
@@ -0,0 +1,120 @@
...
+from nacl.encoding import Base64Encoder
+from nacl.exceptions import BadSignatureError
+from nacl.signing import SigningKey, VerifyKey
+from scriptworker.exceptions import ScriptWorkerEd25519Error, ScriptWorkerException
+from scriptworker.utils import read_from_file
+
+log = logging.getLogger(__name__)
+
+
+def ed25519_key_from_string(obj, string):
```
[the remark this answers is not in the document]
escapewindow: Done.

scriptworker/test/test_utils.py
escapewindow: I've added additional parameters to test, but haven't yet changed the mode.

scriptworker/test/test_utils.py
escapewindow: Here as well.

scriptworker/test/test_utils.py
tomprince: > > I think this should always be binary, to test exactly what got written.
> 
> How do we verify that it was written as non-binary, then?

I think there is an incorrect assumption behind this question. In every case, things end up as bytes on disk. Opening in binary or not changes how things get converted from the internal representation to bytes on disk; opening as text means you can write `str` but not `bytes` and opening as binary vice-versa.

escapewindow: Done.

escapewindow:
@tomprince https://github.com/escapewindow/scriptworker/compare/ed25519...escapewindow:ed25519-wip?expand=1 switches us from `PyNaCl` to `cryptography` for ed25519. Should I land those changes here?

escapewindow:
Once I added the `verify_ed25519_signature` entry point, the cryptography branch became way more appealing than the pynacl branch. We can back out the last 6 commits if we want to revert.

escapewindow:
Added some more logging so we know what's going on. [Here](https://gist.github.com/escapewindow/8d20d9504558a2aaca41ccf6bfb0ff1b)'s a log from maple. It'll get a little less ugly when we're ed25519 across the board :) (We'll need to rebuild the various docker images and toolchains.)

I think we're good here.

escapewindow:
Also with [green generic-worker cot signature verification](https://taskcluster-artifacts.net/Ctlh9aOXQQ-13cCoaNzn2Q/1/public/logs/chain_of_trust.log).

scripts/gen_ed25519_key.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+
+def b64_from_private_key(key):
+    """Get the base64 string from an Ed25519PrivateKey."""
+    return base64.b64encode(key.private_bytes(
+        encoding=serialization.Encoding.Raw,
+        format=serialization.PrivateFormat.Raw,
+        encryption_algorithm=serialization.NoEncryption()
+    ), None).decode('utf-8')
```
alex: I believe `None` is the default on `b64encode()` so it shouldn't be necessary to pass it. If I'm wrong about that, I think it'd be good to add a comment.

scripts/gen_ed25519_key.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def b64_from_public_key(key):
+    """Get the base64 string from an Ed25519PublicKey."""
+    return base64.b64encode(key.public_bytes(
+        encoding=serialization.Encoding.Raw,
+        format=serialization.PublicFormat.Raw,
+    ), None).decode('utf-8')
```
alex: Same

scripts/gen_ed25519_key.py
```diff
@@ -0,0 +1,62 @@
...
+privkey_str = b64_from_private_key(privkey)
+pubkey_str = b64_from_public_key(pubkey)
+
+# test
+privkey2 = private_key_from_string(privkey_str)
+pubkey2 = public_key_from_string(pubkey_str)
+assert b64_from_private_key(privkey2) == privkey_str
+assert b64_from_public_key(pubkey2) == pubkey_str
+
+open("{}private_key".format(prefix), "w").write(privkey_str)
```
alex: Nit: use a `with` statement.

scriptworker/ed25519.py
```diff
@@ -0,0 +1,183 @@
...
+        signature (bytes): the signature to verify
+        message (str): the error message to raise.
+
+    Raises:
+        ScriptWorkerEd25519Error: on failure
+
+    """
+    try:
+        public_key.verify(signature, contents)
+    except (UnsupportedAlgorithm, InvalidSignature) as exc:
```
alex: I don't think `UnsupportedAlgorithm` is possible here.

scriptworker/ed25519.py
```diff
@@ -0,0 +1,183 @@
...
+                    ed25519_public_key_from_string(seed), contents, signature,
+                    "didn't work with {}".format(seed)
+                )
+                log.info("Verified good with {} seed {} !".format(
+                    key_type, seed
+                ))
+                sys.exit(0)
+            except ScriptWorkerEd25519Error:
+                pass
+    else:
```
alex: The `else` clause isn't needed here and I think it's slightly clearer if you just fall through to it.

escapewindow:
Thanks!

scripts/gen_ed25519_key.py
escapewindow: That's right :) Fixed.

scripts/gen_ed25519_key.py
escapewindow: Done.

scripts/gen_ed25519_key.py
escapewindow: Done.

scriptworker/ed25519.py
escapewindow: Removed.

scriptworker/ed25519.py
escapewindow: Removed.

PR #300, opened 2019-01-12
Don't retry on 404s.

escapewindow:
I noticed that we were retrying 404s for downloading the currently largely nonexistent `chain-of-trust.json{,.sig}` files in #294... this means it takes us a few minutes to realize that yes, the file is in fact missing. I *think* not retrying on 404 is ok; if the file is there but we have other issues, it'll be a non-404 status.

If we think this might have larger repercussions, I can retry on `BaseDownloadError` in most cases, and override that with just `DownloadError` specifically for downloading `chain-of-trust.json{,.sig}` until we create those files across all 3 worker implementations.

What do you think, @JohanLorenzo , @tomprince ?

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -290,8 +290,8 @@ async def download_artifacts(context, file_urls, parent_dir=None, session=None,
         list: the full paths to the files downloaded
 
     Raises:
-        scriptworker.exceptions.DownloadError: on download failure after
-            max retries.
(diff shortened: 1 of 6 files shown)


escapewindow:
(This mainly works because [this line](https://github.com/mozilla-releng/scriptworker/blob/2f72b66adfbedd83412b119e377e4a340f238aca/scriptworker/artifacts.py#L313) doesn't include `Download404` in the list of exceptions to retry on).

escapewindow:
Let's give it a try.

PR #302, opened 2019-01-22
Refactors some functions to use a sub-set of context

mitchhentges:
This is a small discussion-y PR to coax out a code style discussion.
Within `scriptworker`, we have a `Context` object that, IMHO, is venturing into ["god object"](https://en.wikipedia.org/wiki/God_object) territory. This is noticeable in a couple ways:

* Tests are less isolated: almost all of them depend on the `context` fixtures. When debugging why a test is failing, the information for that test isn't all local: it's spread throughout some shared test setup that's highly generic. There's a lot of setup that's irrelevant for individual tests (e.g.: setting up a session when a test is just checking config). For example, see the changes to the `test_get_single_upstream_artifact_full_path` test in this PR to see what kind of improvements I'm imagining
* It's harder to find out where a variable is used. For example, `context.proc` (in [task.py](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L436)) is mostly only used in `task.py`, but since it's on the context object, it _could_ be used in `*scripts`, or perhaps accessed in a not-easily-greppable way elsewhere in `scriptworker`
* It's making it harder to call some functions: they require `context` as a parameter, but perhaps only need `context.config`. This is forcing my calling function to need to set up a _whole_ `context`, even if only a subset is needed.

I think that a good remedy to this situation is to only pass around the subset of `context` that we need. Additionally, we should peel off chunks of `context` that don't actually need to be a part of it, such as [`write_json`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/context.py#L163-L175)

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -224,11 +224,11 @@ def _craft_artifact_put_headers(content_type, encoding=None):
 
 
 # get_artifact_url {{{1
-def get_artifact_url(context, task_id, path):
+def get_artifact_url(queue, task_id, path):
     """Get a TaskCluster artifact url.
 
     Args:
-        context (scriptworker.context.Context): the scriptworker context
+        queue (taskcluster.aio.Queue): taskcluster queue
         task_id (str): the task id of the task that published the artifact
         path (str): the relative path of the artifact
 
@@ -240,9 +240,9 @@ def get_artifact_url(context, task_id, path):
 
     """
     if path.startswith("public/"):
-        url = context.queue.buildUrl('getLatestArtifact', task_id, path)
+        url = queue.buildUrl('getLatestArtifact', task_id, path)
     else:
-        url = context.queue.buildSignedUrl(
+        url = queue.buildSignedUrl(
             'getLatestArtifact', task_id, path,
             # XXX Can set expiration kwarg in (int) seconds from now;
             # defaults to 15min.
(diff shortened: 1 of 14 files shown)


escapewindow:
> This is a small discussion-y PR to coax out a code style discussion.

After a quick look: I think this is overall a good change. Agreed that the functions can be simpler. I need to dig a bit deeper for a full review.

Do you have a particular timeframe for this? I'm guessing that this PR doesn't hard-block any OKRs and is only meant to make things nicer (thank you!).

This is definitely a non-backwards compatible change. There may be some *scripts that import more of the scriptworker functions than others; if so, they would need PRs to match, and coordinated rollout. If this lands before #294, I can unbitrot. But if we wanted to, say, completely overhaul scriptworker, we may want to schedule those changes to avoid placing too many yaks in front of OKRs.

mitchhentges:
There isn't a timeframe for this, it's just a little shenanigan on the side :smile:

> There may be some *scripts that import more of the scriptworker functions than others

That's true, though I did do a `grep` through our `*scripts` to ensure that they're not affected (though, this is probably worth a second pass)

For future reference, the affected functions are:

* `artfacts.get_artifact_url`
* `artifacts.download_artifacts`
* `artifacts.get_and_check_single_upstream_artifact_full_path`
* `artifacts.get_single_upstream_artifact_full_path`
* `client._init_logging`
* `log.update_logging_config`
* `utils.download_file`
* `utils.load_json_or_yaml_from_url`

-----

Definitely land #294 first, I don't want this to impact higher-value PRs, I'll handle bitrot on this end :+1:.
It'll probably be worth waiting for the [potential monorepo](https://github.com/mozilla-releng/releng-rfcs/pull/8) before pushing this further, since that'll make it easier to update all references :)

mitchhentges:
I'm not worried about this, I'll wait until we [potentially have a monorepo](https://github.com/mozilla-releng/releng-rfcs/pull/8) before I look into reducing the usage of `context`

PR #304, opened 2019-01-26
Use `git_path` for git binary

rail:
This is very similar to `gpg_path`. We will need this because the `git` binary is not in PATH in the nix generated docker images.

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -113,7 +113,7 @@ DEFAULT_CONFIG = frozendict({
     "base_gpg_home_dir": "...",
     "gpg_lockfile": os.path.join(os.getcwd(), "gpg_homedir.lock"),
     # The path to the git executable.
-    "git_path": None,
+    "git_path": 'git',
     "git_key_repo_dir": "...",
     "git_key_repo_url": "https://github.com/mozilla-releng/cot-gpg-keys.git",
     "last_good_git_revision_file": os.path.join(os.getcwd(), "git_revision"),
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -112,6 +112,8 @@ DEFAULT_CONFIG = frozendict({
 
     "base_gpg_home_dir": "...",
     "gpg_lockfile": os.path.join(os.getcwd(), "gpg_homedir.lock"),
+    # The path to the git executable.
+    "git_path": None,
     "git_key_repo_dir": "...",
     "git_key_repo_url": "https://github.com/mozilla-releng/cot-gpg-keys.git",
     "last_good_git_revision_file": os.path.join(os.getcwd(), "git_revision"),
--- a/scriptworker/version.py
+++ b/scriptworker/version.py
@@ -52,7 +52,7 @@ def get_version_string(version):
 
 # 1}}}
 # Semantic versioning 2.0.0  http://semver.org/
-__version__ = (17, 2, 3)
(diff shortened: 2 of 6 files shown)


scriptworker/version.py
```diff
@@ -52,7 +52,7 @@ def get_version_string(version):
 
 # 1}}}
 # Semantic versioning 2.0.0  http://semver.org/
-__version__ = (17, 2, 2)
+__version__ = (17, 2, 3)
```
escapewindow: We're changing the arguments to functions in a non-backwards compatible way, so this would be `18.0.0`.

scriptworker/constants.py
```diff
@@ -112,6 +112,8 @@
 
     "base_gpg_home_dir": "...",
     "gpg_lockfile": os.path.join(os.getcwd(), "gpg_homedir.lock"),
+    # The path to the git executable.
+    "git_path": None,
```
escapewindow: Let's just set this to `"git"`, and then we don't need `guess_git_path`.

rail: Do you mean I should use `context['git_path']` instead of `guess_git_path(context)`. It's kind of makes sense (less tests!), just double checking.

escapewindow: Yes :) `context.config['git_path']`

mitchhentges:
> We will need this because the git binary is not in PATH in the nix generated docker images.

FWIW, is it possible to modify the nix generated docker images to ensure that `git` is in the `PATH`?

mitchhentges:
:thinking: Huh, I've never seen seen a case where we couldn't trust the `$PATH` before. Since we control the container, we aren't worried about confused users having garbage in their `$PATH`, so I'm a little confused. 

Oh, is this because we're bundling software twice in our `nix` container:
* Underlying software bundled with the system itself
* A second version of the software (e.g.: an older version that we have compatibility with)

Is this the scenario that we're handling here?

escapewindow:
I'm open: this can go into the scriptworker config, or we could have a wrapper that launches scriptworker that sets PATH explicitly (e.g. git path is before system paths). Do we want to continue discussion about this before merging?

mitchhentges:
I don't think that this needs to block merging, I'm just getting context :)
Since we control the container, I'd love to lean more on OS primitives rather than having `$PATH` workarounds leak into `scriptworker`, but I'm more curious than concerned :smile:

rail:
It's more about addressing a hypothetical situation where you may want to have 2 different versions of git/gpg/whatever installed in the same container. Having both versions in `PATH` would not easily work (you'd need to rename the binaries/libs/etc). I'm less concerned about the trust, it's more about being explicit.

mitchhentges:
Hmm, I suppose that with this PR, we're encoding our intent of "use `$binary` at `$location`" within the `scriptworker` config file when we do our `nix`-ing.

We can keep the same benefit (explicit intent) by moving this to the part of `nix`-ing where we define the `$PATH`: explicitly define that the custom binary location is a higher priority than the system installation. In this case, we shouldn't lose explicitness while also not needing to affect `scriptworker`.

-----

Following the pattern implied by this PR, I'm nervous about how, as we depend on additional binaries (or `*scripts` depend on binaries), this has potential to affect `scriptworker` more significantly

escapewindow:
We're going to tear out git and gpg support this quarter, and with them, `git_path` and `gpg_path`, so I'm not terribly worried.

PR #306, opened 2019-01-28
Portable mimetypes tests

### LINKED ISSUE #305: Make tests more portable
(disclaimer: this ticket is not about dependency requirements, such as requiring that `gpg` must be on version `2.0.x`)

We have some tests that perform some mime type checks that depend on the user's platform. These should be updated to assert `scriptworker`'s behaviour, not the behaviour of the underlying environment.

I believe the only env-specific tests are:
* `test_artifacts.test_compress_artifact_if_supported`
* `test_artifacts.test_guess_content_type`

mitchhentges:
Fixes #305.
Also addresses a test failure if the working directory was the root of the repository

### DIFF
--- a/scriptworker/artifacts.py
+++ b/scriptworker/artifacts.py
@@ -4,6 +4,8 @@ Importing this script updates the mimetypes database. This maps some known exten
 in S3.
 
 """
+import posixpath
+
 import aiohttp
 import arrow
 import asyncio
@@ -30,22 +32,17 @@ log = logging.getLogger(__name__)
 
 
 _GZIP_SUPPORTED_CONTENT_TYPE = ('text/plain', 'application/json', 'text/html', 'application/xml')
-_EXTENSIONS_TO_FORCE_TO_PLAIN_TEXT = ('.asc', '.diff', '.log')
 
 
-def _force_mimetypes_to_plain_text():
-    """Populate/Update the mime types database with supported extensions that we want to map to 'text/plain'.
-
-    These extensions can then be open natively read by browsers once they're uploaded on S3. It doesn't affect artifacts
-    once they're downloaded from S3.
-
-    """
-    for extension in _EXTENSIONS_TO_FORCE_TO_PLAIN_TEXT:
-        mimetypes.add_type('text/plain', extension)
-        log.debug('Extension "{}" forced to text/plain'.format(extension))
-
-
-_force_mimetypes_to_plain_text()
+_EXTENSION_TO_MIME_TYPE = {
+    'txt': ('text/plain', None),
+    'tgz': ('application/x-tar', 'gzip'),
+    'dmg': ('application/x-apple-diskimage', None),
+    'log': ('text/plain', None),
+    'asc': ('text/plain', None),
+    'diff': ('text/plain', None),
+    'xml': ('application/xml', None),
+}
 
 
 # upload_artifacts {{{1
@@ -135,6 +132,11 @@ def guess_content_type_and_encoding(path):
         str: the content type of the file
 
     """
+    #  TODO: is mimetypes needed? Maybe we can depend purely on our own _EXTENSION_TO_MIME_TYPE mapping
+    _, extension = posixpath.splitext(path)
+    if _EXTENSION_TO_MIME_TYPE.get(extension[1:]):
+        return _EXTENSION_TO_MIME_TYPE.get(extension[1:])
+
     content_type, encoding = mimetypes.guess_type(path)
     content_type = content_type or "application/binary"
     return content_type, encoding
--- a/scriptworker/test/__init__.py
+++ b/scriptworker/test/__init__.py
@@ -13,7 +13,6 @@ import os
 import pytest
 import tempfile
 import taskcluster.exceptions
-from contextlib import contextmanager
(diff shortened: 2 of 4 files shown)


scriptworker/test/__init__.py
```diff
@@ -241,6 +242,14 @@ def integration_create_task_payload(config, task_group_id, scopes=None,
     }
 
 
+@contextmanager
+def working_directory(path):
```
mitchhentges: This isn't in the `pytest` base package. and that makes me :thinking:: is this not the right way of doing this? Let me know if there's a better way doing this :)

escapewindow:
What errors are you getting without this? (wfm, on mac + docker linux)

scriptworker/test/test_artifacts.py
```diff
@@ -46,21 +47,12 @@ def test_force_mimetypes_to_plain_text(extension):
...
-}
-
-
-# guess_content_type {{{1
-@pytest.mark.parametrize("mime_types", [(k, v) for k, v in sorted(MIME_TYPES.items())])
-def test_guess_content_type(mime_types):
-    path, (mimetype, encoding) = mime_types
-    assert guess_content_type_and_encoding(path) == (mimetype, encoding)
+@pytest.mark.parametrize("mimetypes_response, expected_content_type",
+                         (('text/plain', 'text/plain'), (None, 'application/binary')))
```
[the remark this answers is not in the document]
escapewindow: Looks like we have explicit tests for `_force_mimetypes_to_plain_text`, so that's less of a concern than otherwise.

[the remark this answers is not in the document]
mitchhentges: I'd argue that for this test, we were testing the underlying `mimetypes` functionality more than `guess_content_type_and_encoding(...)`.

Assuming that the tested standard library module works properly (which should be a safe assumption), shouldn't this test just be verifying that `guess_content_type_and_encoding(...)` works correctly? And, the two branches of that function are:
* `mimetypes` guessed a value
* `mimetypes` couldn't guess a value

I'm running Arch Linux - are you running Ubuntu? My Ubuntu laptop passes these tests, but I bet that there's some extra bundled mime types

scriptworker/test/__init__.py
mitchhentges: Nice! that's exactly what I was looking for :)

scriptworker/test/test_artifacts.py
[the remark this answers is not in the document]
escapewindow: I'm on OSX. I think the docker image is a flavor of debian.

I'm almost leaning towards renaming `_force_mimetypes_to_plain_text` and having a dictionary of suffix to mimetype that we want to force. We really want to have a consistent set of mime types when we upload, especially for the files we ship like `.dmg`.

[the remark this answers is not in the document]
mitchhentges: Yeah, that sounds great!
I'll add it as part of this PR, so it can go hand-in-hand with the test change, so we don't lose coverage?

scriptworker/artifacts.py
```diff
@@ -135,6 +132,11 @@ def guess_content_type_and_encoding(path):
         str: the content type of the file
 
     """
+    #  TODO: is mimetypes needed? Maybe we can depend purely on our own _EXTENSION_TO_MIME_TYPE mapping
```
escapewindow: I thought about this a bit too. We mostly upload a known set of file extensions, and falling back to `application/binary` isn't the worst. But there are no guarantees we won't upload a new file extension, or a dozen, in the future. So we could:

- have a small subset of known file extensions and potentially adjust this map every time we add a new extension
- have a full set of known file extensions, and potentially adjust this map more rarely, if we upload a brand new file extension that doesn't exist in the full set
- have a small subset of file extensions to override the system mimetypes, but fall back to the system definitions. This is what we have here.

In each of these cases, we could require a scriptworker release, or we could configure the known file extensions via the `scriptworker.yaml` config file per scriptworker instance.

In practice I'm not sure the win is large enough to warrant further change, but I could be wrong.

mitchhentges: Part of why this is so challenging is that we pass `content_type` to Taskcluster, at which point any number of things consume it. Unfortunately, this means that all those downstream components depend on the content-type detection of `scriptworker`.

In the mean time, I think that option #3 (have small subset, fall back to system definitions) makes the most sense. Though it's ~inconsistent, it seems to have generally worked so far.

-----

Hmm, we might have a 4th option. If the only reason we're overriding mimetypes is for the gzip logic (and **not** for downstream consumption, e.g.: browsers when clicking artifacts), then we can just do:
* Use mimetypes + extension-matching to determine if something is gzippable. But, when providing `content_type` to TC, just give it the system mimetype without overriding.

I don't think that this discussion is blocking the PR, so I'll merge it :)

escapewindow: > Use mimetypes + extension-matching to determine if something is gzippable. But, when providing content_type to TC, just give it the system mimetype without overriding.

I think that means you end up with files like blah.log or foo.asc that are cleartext, but you have to download to read instead of being able to read them in your browser.

mitchhentges: Ew :)
Yeah, I agree that option 3 is probably our best bet :smile:

PR #307, opened 2019-01-29
Allow mobile to create in-tree docker images

JohanLorenzo:
Blocks https://github.com/mozilla-mobile/android-components/pull/1874

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -96,6 +96,9 @@ DEFAULT_CONFIG = frozendict({
     "cot_version": 3,
     "min_cot_version": 2,
     "max_chain_length": 20,
+    # Calls to Github API are limited to 60 an hour. Using an API token allows to raise the limit to
+    # 5000 per hour. https://developer.github.com/v3/#rate-limiting
+    "github_oauth_token": "",
 
     # Specify a default gpg home other than ~/.gnupg
     "gpg_home": None,
@@ -189,6 +192,9 @@ DEFAULT_CONFIG = frozendict({
         "gecko-1-images",
         "gecko-2-images",
         "gecko-3-images",
+
+        "mobile-1-images",  # there is no mobile level 2.
+        "mobile-3-images",
     ),
 
     # for trace_back_to_*_tree.  These repos have access to restricted scopes;
@@ -230,6 +236,30 @@ DEFAULT_CONFIG = frozendict({
                     for repo_name in ('android-components', 'focus-android', 'reference-browser', 'fenix')
                 ]),
             }),),
+        }),
+    },
+
+    'valid_tasks_for': {
+        'by-cot-product': frozendict({
+            'firefox': ('hg-push', 'cron', 'action',),
+            'thunderbird': ('hg-push', 'cron', 'action',),
+            'mobile': (
+                'cron',
+                # On staging releases, level 1 docker images may be built in the pull-request graph
+                'github-pull-request',
+                # Similarly, docker images can be built on regular push. This is usually the case
+                # for level 3 images
+                'github-push',
+                'github-release',
+            ),
+        }),
+    },
+
+    'official_github_repos_owner': {
+        'by-cot-product': frozendict({
+            'firefox': '',
+            'thunderbird': '',
+            'mobile': 'mozilla-mobile',
         }),
     },
 
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -29,15 +29,19 @@ from scriptworker.config import read_worker_creds, apply_product_config
 from scriptworker.constants import DEFAULT_CONFIG
 from scriptworker.context import Context
 from scriptworker.exceptions import CoTError, BaseDownloadError, ScriptWorkerGPGException
+from scriptworker.github import GitHubRepository
 from scriptworker.gpg import get_body, GPG
 from scriptworker.log import contextual_log_handler
 from scriptworker.task import (
+    extract_github_repo_owner_and_name,
     get_action_callback_name,
     get_and_check_project,
     get_and_check_tasks_for,
     get_commit_message,
     get_decision_task_id,
     get_parent_task_id,
+    get_pull_request_number,
+    get_push_date_time,
     get_repo,
     get_repo_scope,
     get_revision,
@@ -45,7 +49,7 @@ from scriptworker.task import (
     get_triggered_by,
     get_task_id,
     get_worker_type,
-    is_try,
+    is_try_or_pull_request,
     is_action,
 )
 from scriptworker.utils import (
@@ -117,16 +121,16 @@ class ChainOfTrust(object):
         """
         return [x.task_id for x in self.links]
 
-    def is_try(self):
+    def is_try_or_pull_request(self):
         """Determine if any task in the chain is a try task.
 
         Returns:
             bool: True if a task is a try task.
 
         """
-        result = is_try(self.task, source_env_prefix=self.context.config['source_env_prefix'])
+        result = is_try_or_pull_request(self.context, self.task)
         for link in self.links:
-            if link.is_try:
+            if link.is_try_or_pull_request:
                 result = True
                 break
         return result
@@ -184,7 +188,7 @@ class LinkOfTrust(object):
         context (scriptworker.context.Context): the scriptworker context
         decision_task_id (str): the task_id of self.task's decision task
         parent_task_id (str): the task_id of self.task's parent task
-        is_try (bool): whether the task is a try task
+        is_try_or_pull_request (bool): whether the task is a try or a pull request task
         name (str): the name of the task (e.g., signing.decision)
         task_id (str): the taskId of the task
         task_graph (dict): the task graph of the task, if this is a decision task
@@ -226,7 +230,7 @@ class LinkOfTrust(object):
         """dict: the task definition.
 
         When set, we also set ``self.decision_task_id``, ``self.parent_task_id``,
-        ``self.worker_impl``, and ``self.is_try`` based on the task definition.
+        and ``self.worker_impl`` based on the task definition.
 
         """
         return self._task
@@ -238,7 +242,11 @@ class LinkOfTrust(object):
         self.decision_task_id = get_decision_task_id(self.task)
         self.parent_task_id = get_parent_task_id(self.task)
         self.worker_impl = guess_worker_impl(self)
-        self.is_try = is_try(self.task, source_env_prefix=self.context.config['source_env_prefix'])
+
+    @property
+    def is_try_or_pull_request(self):
+        """bool: the task is either a try or a pull request one."""
+        return is_try_or_pull_request(self.context, self.task)
 
     @property
     def cot(self):
@@ -1030,11 +1038,61 @@ async def _get_additional_hg_push_jsone_context(parent_link, decision_link):
     }
 
 
-def _get_additional_github_releases_jsone_context(parent_link, decision_link):
+async def _get_additional_github_releases_jsone_context(decision_link):
+    context = decision_link.context
+    source_env_prefix = context.config['source_env_prefix']
+    task = decision_link.task
+    repo_url = get_repo(task, source_env_prefix)
+    repo_owner, repo_name = extract_github_repo_owner_and_name(repo_url)
+    tag_name = get_revision(task, source_env_prefix)
+
+    release_data = GitHubRepository(
+        repo_owner, repo_name, context.config['github_oauth_token']
+    ).get_release(tag_name)
+
+    # The release data expose by the API[1] is not the same as the original event[2]. That's why
+    # we have to rebuild the object manually
+    #
+    # [1] https://developer.github.com/v3/repos/releases/#get-a-single-release
+    # [2] https://developer.github.com/v3/activity/events/types/#releaseevent
+    return {
+        'event': {
+            'repository': {
+                # TODO: Append ".git" to clone_url in order to match what GitHub actually provide.
+                # This can't be done at the moment because some mobile projects still rely on the
+                # bad value
+                'clone_url': repo_url,
+                'html_url': repo_url,
+            },
+            'release': {
+                'tag_name': tag_name,
+                'target_commitish': release_data['target_commitish'],
+                # Releases are supposed to be unique. Therefore, we're able to use the latest
+                # timestamp exposed by the API (unlike PRs, for instance)
+                'published_at': release_data['published_at'],
+            },
+            'sender': {
+                'login': release_data['author']['login'],
+            },
+        },
+    }
+
+
+def _get_additional_git_cron_jsone_context(decision_link):
     source_env_prefix = decision_link.context.config['source_env_prefix']
     task = decision_link.task
     repo = get_repo(task, source_env_prefix)
+
+    # TODO remove the call to get_triggered_by() once Github repos don't define it anymore.
+    user = get_triggered_by(task, source_env_prefix)
+    if user is None:
+        # We can't default to 'cron' (like in hg) because https://github.com/cron is already taken
+        user = 'TaskclusterHook'
+
+    # Cron context mocks a GitHub release one. However, there is no GitHub API to call since this
+    # is built on Taskcluster.
     return {
+        'cron': load_json_or_yaml(decision_link.task['extra']['cron']),
         'event': {
             'repository': {
                 # TODO: Append ".git" to clone_url in order to match what GitHub actually provide.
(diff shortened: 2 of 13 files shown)


JohanLorenzo:
This was tested end-to-end in almost all possible scenarios:

* a cron graph that fetches a docker image built in a PR: https://tools.taskcluster.net/groups/COzRCwrmTAiRVyFSvT0G0A/tasks/MbtI7uX-Sg6QV_FTxOQ47g/runs/2/logs/public%2Flogs%2Fchain_of_trust.log#L137
* a cron graph that gets the docker image from a github-push graph: https://tools.taskcluster.net/groups/UduvnE9JSzKDzwqTZIZ06w/tasks/LEwtEPP4TE6IXaqGnr7OTA/runs/0/logs/public%2Flogs%2Fchain_of_trust.log#L135
* a release graph based off a PR: https://tools.taskcluster.net/groups/aEgavtwJT7Oh9w9jeMbMmw/tasks/UHkB7QknSqqlPFWMswTVAg/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L140
* a release based off a github-push: https://tools.taskcluster.net/groups/GSS2W84PTAO3-mZSB1GLAw/tasks/Iui8d4aYRiyiU2bW3jExSA/runs/0/logs/public%2Flogs%2Fchain_of_trust.log#L138

The remaining scenarios would when the docker image is built within the release/cron graph. This scenario is unlikely, and it's actually going through the same path tested above. So I think the test coverage is good enough.

scriptworker/github.py
```diff
@@ -0,0 +1,90 @@
+"""GitHub helper functions."""
```
tomprince: I have a couple of suggestions about the code here:
- I think the functions here should be at a higher level of abstraction. In other words, they should fetch the data themselves, and provide it in a useful form. That way, `.cot.verify` doesn't need to know details of the structure of github's API, or the results it generates.
- I would consider using an existing binding to the github api, such as [github3.py](https://github3py.readthedocs.io/en/stable/), rather than rolling our own. (Though still possibly worth while to wrap it here).

scriptworker/cot/verify.py
```diff
@@ -1030,11 +1041,77 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
+def _get_additional_git_cron_jsone_context(decision_link):
     source_env_prefix = decision_link.context.config['source_env_prefix']
     task = decision_link.task
     repo = get_repo(task, source_env_prefix)
+
+    # TODO remove the call to get_triggered_by() once Github repos don't define it anymore.
+    user = get_triggered_by(task, source_env_prefix)
+    if user is None:
+        # We can't default to 'cron' (like in hg) because https://github.com/cron is already taken
+        user = 'TaskclusterHook'
```
tomprince: Is there any reason we can't have `None` here?

Can you explain a little bit how this is used, and what the context is? If we aren't using `cron` because that already exists, I worry if someone were to create this user, they might be able to subvert some part of our checks. We could, of course, create the user, but I wonder if there is some way to structure things to avoid that.

scriptworker/github.py
tomprince: Having looked at the changes in `.cot.verify`, it looks like these are use mainly to synthesize github generated events. One thought that occurred to me (though I'm not convinced it is correct) is that functions here should do the work of synthesizing the event data, and the code in `.cot.verify` can wrap that to create the context.

If we do go this direction, I suspect that the cron context should be handled differently, since the pseudo-event isn't actually related to github.

scriptworker/task.py
```diff
@@ -311,6 +357,20 @@ def get_repo_scope(task, name):
...
+
+    Args:
+        repo_url (str): The URL to the GitHub repository
+
+    Returns:
+        str, str: the owner of the repository, the repository name
+
+    """
+    elements = repo_url.split('/')
+    return elements[3], elements[4]
```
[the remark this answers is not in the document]
JohanLorenzo: That's a good idea. Fixed

scriptworker/cot/verify.py
JohanLorenzo: Yeah, we need to provide a user name so `metadata.owner` is defined in the task definition. `metadata.owner` must be an email address. For instance: https://github.com/mozilla-mobile/fenix/blob/b8ab51318e99706656f54669594ef8263cf560b6/.taskcluster.yml#L114

This value is hardcoded in the scripts that spin the cron graphs. For instance: https://github.com/mozilla-mobile/fenix/blob/b8ab51318e99706656f54669594ef8263cf560b6/automation/taskcluster/schedule_nightly_graph.py#L64

An alternative solution could be to use `cron@noreply.mozilla.org` instead of an `@users.noreply.github.com` address.

scriptworker/github.py
JohanLorenzo: Done. I used github3.py, even that's not an async library. The reason is: none of [the highlighted libraries](https://developer.github.com/v3/libraries/) is async. If we want to make this module async, we could use [aioify](https://github.com/yifeikong/aioify). However, the lib is quite unstable and I'm not certain it's worth the effort yet => we don't parallelize any call to github in `cot.verify`. 

Regarding the cron context, we can define something that doesn't try to be a github event. It was implemented this way because it allows some factorization in https://github.com/mozilla-mobile/android-components/blob/b2c827e66d89f25688a15e7b27f7494ad2c718dc/.taskcluster.yml#L153 and in the scripts that handles the decision task. If we want to revisit this context, I would suggest to do it in a follow up.

JohanLorenzo:
You can review the last 4 commits. I'll squash them afterwards.

JohanLorenzo:
Tested end-to-end: https://tools.taskcluster.net/groups/DOOc9bWqTVKvT-nfPr9wwQ

scriptworker/task.py
```diff
@@ -311,6 +360,28 @@ def get_repo_scope(task, name):
...
+        repo_url (str): The URL to the GitHub repository
+
+    Raises:
+        ValueError: on repo_url that aren't from github
+
+    Returns:
+        str, str: the owner of the repository, the repository name
+
+    """
+    repo_url = repo_url.rstrip('.git')
```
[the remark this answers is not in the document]
JohanLorenzo: Oh right! I had forgotten about how `rstrip()` works. Thanks for catching that1

scriptworker/task.py
```diff
@@ -311,6 +360,28 @@ def get_repo_scope(task, name):
...
+    elements = repo_url.split('/')
+
+    if elements[2] != 'github.com':
+        raise ValueError('{} is not a GitHub URL!'.format(repo_url))
```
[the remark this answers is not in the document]
JohanLorenzo: Good thinking. Done.

scriptworker/cot/verify.py
```diff
@@ -1030,11 +1034,61 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
 
 
-def _get_additional_github_releases_jsone_context(parent_link, decision_link):
+async def _get_additional_github_releases_jsone_context(decision_link):
+    context = decision_link.context
+    source_env_prefix = context.config['source_env_prefix']
+    task = decision_link.task
+    repo_url = get_repo(task, source_env_prefix)
+    repo_owner, repo_name = extract_github_repo_owner_and_name(repo_url)
+    tag_name = get_revision(task, source_env_prefix)
```
escapewindow: Is there ever a chance that this isn't a tag? If so, should we catch that here or in `GithubRepository`, or will we just fail out in chain of trust verification?

scriptworker/cot/verify.py
```diff
@@ -1030,11 +1034,61 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
+    # we have to rebuild the object manually
+    #
+    # [1] https://developer.github.com/v3/repos/releases/#get-a-single-release
+    # [2] https://developer.github.com/v3/activity/events/types/#releaseevent
+    return {
+        'event': {
+            'repository': {
+                # TODO: Append ".git" to clone_url in order to match what GitHub actually provide.
+                # This can't be done at the moment because some mobile projects still rely on the
+                # bad value
```
escapewindow: We should move to taskgraph everywhere; that may provide an opportunity to standardize stuff like this.
You have a similar comment below; should we call a noop function that we'll add `.git` to later? Non-blocking.

scriptworker/cot/verify.py
```diff
@@ -1030,11 +1034,61 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
...
+                'clone_url': repo_url,
+                'html_url': repo_url,
+            },
+            'release': {
+                'tag_name': tag_name,
+                'target_commitish': release_data['target_commitish'],
+                # Releases are supposed to be unique. Therefore, we're able to use the latest
+                # timestamp exposed by the API (unlike PRs, for instance)
```
escapewindow: I'm picturing someone deleting release X and then recreating. They'd probably cause a lot of noise [?]

scriptworker/cot/verify.py
escapewindow: If this is purely a metadata thing, using a Mozilla email address seems good. Non-blocker.

scriptworker/task.py
```diff
@@ -311,6 +360,29 @@ def get_repo_scope(task, name):
...
+    Returns:
+        str, str: the owner of the repository, the repository name
+
+    """
+    if not repo_url.startswith('https://github.com/'):
+        raise ValueError('{} is not a supported GitHub URL!'.format(repo_url))
+
+    if repo_url.endswith(".git"):
+        repo_url = repo_url[:-len(".git")]
+    elements = repo_url.split('/')
```
escapewindow: There are enough weird ways to specify urls that it may be safest to use [`urlparse`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) and then act on the path.

scriptworker/cot/verify.py
```diff
@@ -1046,9 +1100,94 @@ def _get_additional_github_releases_jsone_context(parent_link, decision_link):
...
             },
             'sender': {
-                'login': get_triggered_by(task, source_env_prefix),
+                'login': user,
+            },
+        },
+    }
+
+
+async def _get_additional_github_pull_request_jsone_context(decision_link):
```
escapewindow: Overall, awesome that we're supporting pull requests; that will help a great deal. We do need to be a bit careful about permissions.

In a pull request, which `.taskcluster.yml` do we use to generate the tasks? The pull request's `.taskcluster.yml`? We can't necessarily trust the code or config in the pull request, especially since this is pre-review from an arbitrary user. We may want something like `is_try` to notice that our `tasks_for` specifies a pull request.

scriptworker/cot/verify.py
```diff
@@ -29,15 +29,19 @@
...
 from scriptworker.task import (
+    extract_github_repo_owner_and_name,
     get_action_callback_name,
     get_and_check_project,
     get_and_check_tasks_for,
     get_commit_message,
     get_decision_task_id,
     get_parent_task_id,
+    get_pull_request_number,
+    get_push_date_time,
```
escapewindow: These functions seem like they equally belong in `scriptworker.github` and `scriptworker.task`. If we end up supporting, say, bitbucket as well, and if it has different behavior, we might end up moving these functions.

scriptworker/task.py
```diff
@@ -31,7 +31,17 @@
 
 log = logging.getLogger(__name__)
 
-KNOWN_TASKS_FOR = ('hg-push', 'cron', 'action', 'github-release')
+KNOWN_TASKS_FOR = (
```
escapewindow: Hm. I wonder if we should change this to have a set of known `tasks_for` per `cot-product`. We're likely good for now, since gecko will barf if we try to use github cot verification there.

scriptworker/task.py
```diff
@@ -211,8 +217,51 @@ def get_triggered_by(task, source_env_prefix):
...
+    Returns:
+        str: the string when the event was pushed. It's usually formated as ISO 8601. However, it
+            may be an epoch timestamp, (known case: github-push events).
+        None: if not defined for this task.
+
+    """
+    return _extract_from_env_in_payload(task, source_env_prefix + '_PUSH_DATE_TIME')
+
+
+def _extract_from_env_in_payload(task, key, default=None):
```
escapewindow: Smart.

scriptworker/version.py
```diff
@@ -52,7 +52,7 @@ def get_version_string(version):
 
 # 1}}}
 # Semantic versioning 2.0.0  http://semver.org/
-__version__ = (18, 1, 0)
+__version__ = (19, 0, 0)
```
escapewindow: This might be 19.1.0 or 20.0.0 now. I think the only potentially backwards-incompatible change is the new dependency.

scriptworker/cot/verify.py
JohanLorenzo: Yeah, we need to rely on the PR's `.taskcluster.yml`. That's useful to test out staging releases. I like the idea of `is_try`. I just implemented something like this!

scriptworker/task.py
JohanLorenzo: Good idea! Done.

scriptworker/task.py
JohanLorenzo: Fixed!

scriptworker/cot/verify.py
JohanLorenzo: Okay. I'll address it in a followup

JohanLorenzo:
Retested end-to-end:
* staging release: https://tools.taskcluster.net/groups/ejmq0cldR56SKq8RJX9efg
* staging cron: https://tools.taskcluster.net/groups/JqmGpS0zRkSVYcawGIs9bg

scriptworker/cot/verify.py
escapewindow: I thought about this -- since we have a limited number of repos that run cron, we can change the email in `.taskcluster.yml` as soon as the new email is rolled out in `cot.verify`, and then we should be good.

scriptworker/cot/verify.py
```diff
@@ -242,7 +242,11 @@ def task(self, task):
         self.decision_task_id = get_decision_task_id(self.task)
         self.parent_task_id = get_parent_task_id(self.task)
         self.worker_impl = guess_worker_impl(self)
-        self.is_try = is_try(self.task, source_env_prefix=self.context.config['source_env_prefix'])
+
+    @property
+    def is_try_or_pull_request(self):
+        """bool: the task is either a try or a pull request one."""
+        return is_try_or_pull_request(self.context, self.task)
```
escapewindow: We may end up calling this multiple times, but most likely that won't be too expensive.

scriptworker/constants.py
```diff
@@ -239,6 +239,30 @@
         }),
     },
 
+    'valid_tasks_for': {
```
escapewindow: 👍

scriptworker/task.py
```diff
@@ -377,21 +364,30 @@ def extract_github_repo_owner_and_name(repo_url):
...
+    return parts
+
+
+def _is_try_url(url):
+    return 'try' in _get_parts_of_url_path(url)[0]
+
+
+def _does_url_come_from_official_repo(context, url):
+    official_repo_owner = context.config['official_github_repos_owner']
+    return not official_repo_owner or official_repo_owner == _get_parts_of_url_path(url)[0]
```
escapewindow: Hm. Do we really want this to be `True` for non-mobile? I suppose we're checking this elsewhere. I might lean towards `True` / `False` / `None` for no owner specified. Not a blocker, just curious.

scriptworker/cot/verify.py
JohanLorenzo: Yeah, I was worried about that too. I don't think it is either

scriptworker/task.py
JohanLorenzo: Hmmm, I could use `None` and check for it later. I'll address this in a followup

JohanLorenzo:
> This looks good, thank you! We might have to add a couple of branch test coverage, depending on what coveralls tells us. (`coverage html` is telling us 99% on a few files.)

FYI: https://travis-ci.org/mozilla-releng/scriptworker/jobs/493004995#L1226 and https://travis-ci.org/mozilla-releng/scriptworker/jobs/493004995#L1240 are in master. I filed https://github.com/mozilla-releng/scriptworker/issues/311 to track this

scriptworker/task.py
```diff
@@ -312,14 +348,46 @@ def get_repo_scope(task, name):
...
+        ValueError: on repo_url that aren't from github
+
+    Returns:
+        str, str: the owner of the repository, the repository name
+
+    """
+    if not repo_url.startswith('https://github.com/'):
+        raise ValueError('{} is not a supported GitHub URL!'.format(repo_url))
+
+    parts = _get_parts_of_url_path(repo_url)
```
tomprince: If you change it to 
```suggestion
    repo_owner, repo_name = _get_parts_of_url_path(repo_url)
```
then it will automatically error if there are not exactly two parts, which is correct.

scriptworker/cot/verify.py
tomprince: I don't think we can trust `tasks_for` to determine if something is a pull request, since it comes from the untrusted `.taskcluster.yml`.

JohanLorenzo: I agree. This can easily be hardcoded. Just like `is_try`, I implemented a couple checks on the URL used in the decision task https://github.com/mozilla-releng/scriptworker/pull/307/files#diff-27eeb4b1649924d24d3d4b07fc7da4d5R448. These can be changed too, but if `metadata.source` is forged, then `.taskcluster.yml` must match something that was shipped on the official repository. This makes the attack surface lower, IMO.

scriptworker/task.py
JohanLorenzo: That doesn't work if the URL is something like: https://github.com/JohanLorenzo/android-components/raw/0e8222165d3ec11e932d5f900e8851dc98f62e98/.taskcluster.yml (tested at https://github.com/mozilla-releng/scriptworker/pull/307/files#diff-6e00aba980aa1d91133b0b8bfe689d88R198), which is still valid, IMO. Maybe it shouldn't. In which case, I'd prefer to handle this in a followup

JohanLorenzo:
I don't see any major thing to change, anymore. I squashed the commits. I'm going to land the PR as is and deploy it today (in order to avoid a deployment on a Friday). I'll address the remaining nits in a followup.

Thanks all for the input!

PR #309, opened 2019-01-31
Bug 1524358 - also provide taskId in json-e contexts

djmitche:


scriptworker/cot/verify.py
```diff
@@ -1277,6 +1278,7 @@ def _get_action_perm(action_defn):
             'taskId': parent_link.task_id,
             'now': jsone_context['now'],
             'as_slugid': jsone_context['as_slugid'],
+            'taskId': json_context['as_slugid']('decision'),
```
escapewindow: This should probably be `jsone_context`

scriptworker/cot/verify.py
```diff
@@ -1277,6 +1278,7 @@ def _get_action_perm(action_defn):
             'taskId': parent_link.task_id,
```
escapewindow: Also, we have `taskId` set twice now?

scriptworker/cot/verify.py
```diff
@@ -1102,6 +1102,7 @@ def _get_additional_github_releases_jsone_context(parent_link, decision_link):
     jsone_context = {
         'now': parent_link.task['created'],
         'as_slugid': lambda x: task_ids.get(x, task_ids['default']),
+        'taskId': task_ids['decision'],
```
escapewindow: This change also adds a 2nd conflicting `taskId` (see line 1111)

scriptworker/cot/verify.py
tomprince: I don't think this wants to change, as this is the action task json-e context being generated.

scriptworker/cot/verify.py
tomprince: With the conflicting `taskId` removed, this looks correct.

scriptworker/cot/verify.py
```diff
@@ -1270,6 +1273,8 @@ def _get_action_perm(action_defn):
         action_perm = _get_action_perm(action_defn)
         tmpl = _wrap_action_hook_with_let(in_tree_tmpl, action_perm)
 
+        # define the JSON-e context with which the hook's task template was rendered,
+        # working backward from the json-e context the .taskcluster.yml expects
```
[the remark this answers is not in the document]
tomprince: No. This is the context described [here](https://docs.taskcluster.net/docs/reference/core/taskcluster-hooks/docs/firing-hooks#triggerhook), which does not have `ownTaskId`. Maybe there should be a linke to that here, though?

djmitche:
I'll fix up the flakes.

scriptworker/cot/verify.py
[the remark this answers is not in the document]
djmitche: Yeah, a link to that doc would be good.

escapewindow:
[b7e1259](https://github.com/mozilla-releng/scriptworker/pull/309/commits/b7e125921f386c7363f65e01f3c272e6f3724b73) should fix up the tests.

djmitche:
OK, now it's green :)

I'll get a working try push and ping you in the bug.

PR #310, opened 2019-02-02
Stops scriptworker upon SIGTERM being received

### LINKED ISSUE #260: On SIGTERM, cancel current task and stop
[pushsnapscript](https://github.com/mozilla-releng/pushapkscript) is the kind of worker that doesn't output anything for several minutes (this is a limitation of [snapcraft](https://github.com/snapcore/snapcraft)). It also takes a while to complete. Therefore, scriptworker has to reclaim the task, at least once.

RyanVM discovered: https://tools.taskcluster.net/groups/YndHp56KRGiPUmmt0tNd0Q/tasks/dzTJrSbLQ96gqd7sVZuk1A/runs/0

There is no logs for run 0 on Taskcluster, but here are the ones on the worker:

```
2018-09-21T13:37:29     INFO - Pushing target.snap
[                                                    ]   0% Pushing target.snap
[===                                                 ]   6% Pushing target.snap
[======                                              ]  12% Pushing target.snap
[=========                                           ]  18% Pushing target.snap
[============                                        ]  24% Pushing target.snap
[===============                                     ]  30% Pushing target.snap
[==================                                  ]  36% Pushing target.snap
[=====================                               ]  42% Pushing target.snap
[=========================                           ]  48% Pushing target.snap
[============================                        ]  54% Pushing target.snap
[===============================                     ]  60% Pushing target.snap
[==================================                  ]  66% Pushing target.snap
[=====================================               ]  72% Pushing target.snap
[========================================            ]  78% Pushing target.snap
[===========================================         ]  84% Pushing target.snap
[==============================================      ]  90% Pushing target.snap
[==================================================  ]  96% Pushing target.snap
[====================================================]  100% Pushing target.snap
2018-09-21T13:39:35    DEBUG - Reclaiming task...
2018-09-21T13:39:36    DEBUG - Reclaim task response:
{'credentials': '{********}',
 'runId': 0,
 'status': {'deadline': '2018-09-22T13:34:18.351Z',
            'expires': '2019-09-21T13:34:18.351Z',
            'provisionerId': 'scriptworker-prov-v1',
            'retriesLeft': 5,
            'runs': [{'reasonCreated': 'scheduled',
                      'runId': 0,
                      'scheduled': '2018-09-21T13:34:34.263Z',
                      'started': '2018-09-21T13:34:35.194Z',
                      'state': 'running',
                      'takenUntil': '2018-09-21T13:59:35.934Z',
                      'workerGroup': 'pushsnap-v1',
                      'workerId': 'pushsnapworker-1'}],
            'schedulerId': 'gecko-level-3',
            'state': 'running',
            'taskGroupId': 'YndHp56KRGiPUmmt0tNd0Q',
            'taskId': 'dzTJrSbLQ96gqd7sVZuk1A',
            'workerType': 'pushsnap-v1'},
 'takenUntil': '2018-09-21T13:59:35.934Z',
 'workerGroup': 'pushsnap-v1',
 'workerId': 'pushsnapworker-1'}
2018-09-21T13:39:36    DEBUG - waiting 300 seconds before reclaiming...
2018-09-21T13:43:27     INFO - SIGTERM received; shutting down after next task
2018-09-21T13:43:39     INFO - Scriptworker starting up at 2018-09-21 13:43:39+00:00 UTC
2018-09-21T13:43:39    DEBUG - rm(/builds/scriptworker/work)
2018-09-21T13:43:39    DEBUG - makedirs(/builds/scriptworker/work)
2018-09-21T13:43:39    DEBUG - rm(/builds/scriptworker/artifacts)
2018-09-21T13:43:39    DEBUG - makedirs(/builds/scriptworker/artifacts)
2018-09-21T13:43:39    DEBUG - makedirs(/builds/scriptworker/artifacts/public/logs)
2018-09-21T13:43:39    DEBUG - Calling claimWork...
2018-09-21T13:43:40     INFO - rebuild_gpg_homedirs()...
2018-09-21T13:43:40     INFO - Updating git repo
2018-09-21T13:43:40     INFO - rebuilding gpg_home for /home/cltbld/pubkey at /tmp/tmp7bj8zlsv...
```

As you can see, once SIGTERM is received, the task is immediately killed. That explains why no logs were uploaded. Moreover the 300 seconds weren't waited on.

mitchhentges:
Fixes #260 .
`cancellable_verify_chain_of_trust` and `to_cancellable_process` aren't the nicest way of doing this, but somehow we need to be able to communicate shutdowns to CoT verification and the subprocess.

I'm following this up with a `build-puppet` change so that we have 30 seconds to perform this shutdown operation, because sometimes propagating the "build expired" information to Taskcluster takes dangerously close to 10 seconds (the default).

Note: is it possible to be alerted when `supervisord` is forced to kill our processes? Would be awesome to know for applications where there's side-effects (e.g.: 20 minute hanging builds) if cleanup doesn't finish


scriptworker/test/test_worker.py
```diff
@@ -60,27 +61,32 @@ def test_main(mocker, context, event_loop):
...
     """Test that sending SIGTERM causes the main loop to stop after the next
     call to async_main."""
-    config = dict(context.config)
-    config['poll_interval'] = 1
-    creds = {'fake_creds': True}
-    config['credentials'] = deepcopy(creds)
+    run_tasks_cancelled = event_loop.create_future()
 
-    async def async_main(*args):
+    class MockRunTasks:
```
escapewindow: There's a `constants helpers and fixtures` section at the top; do these belong there?

scriptworker/test/test_worker.py
```diff
@@ -254,3 +260,296 @@ def _mocker_run_tasks_helper(mocker, exc, func_to_raise):
     mocker.patch.object(worker, "complete_task", new=noop_async)
     status = await worker.run_tasks(context)
     assert status == context.config['task_max_timeout_status']
+
+
+_MOCK_CLAIM_WORK_RETURN = {
```
escapewindow: Same with these. If these are potentially useful to tests for other modules, these might belong in `test/__init__.py`

scriptworker/test/test_worker.py
```diff
@@ -254,3 +260,296 @@ def _mocker_run_tasks_helper(mocker, exc, func_to_raise):
...
+        return result
+    return fn
+
+
+@pytest.mark.asyncio
+async def test_run_tasks_no_cancel(context, mocker):
+    mocker.patch('scriptworker.worker.claim_work', create_async(_MOCK_CLAIM_WORK_RETURN))
+    mocker.patch.object(asyncio, 'sleep', create_async())
+    mocker.patch('scriptworker.worker.prepare_to_run_task', create_async())
+    mocker.patch('scriptworker.worker.reclaim_task', create_async())
```
escapewindow: I've got `noop_async` and `noop_sync` sprinkled throughout the test code, possible these are combinable.

Also, agreed that `test_worker.py` has a *ton* of mocks. But even if we follow clean architecture best practices, somewhere we're dealing with the i/o and complexity. In scriptworker, that's `scriptworker.worker`.

scriptworker/test/test_worker.py
```diff
@@ -60,27 +61,32 @@ def test_main(mocker, context, event_loop):
...
-    config['poll_interval'] = 1
-    creds = {'fake_creds': True}
-    config['credentials'] = deepcopy(creds)
+    run_tasks_cancelled = event_loop.create_future()
 
-    async def async_main(*args):
+    class MockRunTasks:
+        @staticmethod
+        def cancel():
+            nonlocal run_tasks_cancelled
```
escapewindow: I think this works without the `nonlocal`, though I don't think it hurts.

scriptworker/worker.py
```diff
@@ -90,6 +94,87 @@
...
+
+            # Assume only a single task, but should more than one fall through,
+            # run them sequentially.  A side effect is our return status will
+            # be the status of the final task run.
+            status = None
+            for task_defn in tasks.get('tasks', []):
+                prepare_to_run_task(context, task_defn)
+                reclaim_fut = context.event_loop.create_task(reclaim_task(context, context.task))
+                status = await do_run_task(context, self._cancellable_verify_chain_of_trust, self._to_cancellable_process)
+                status = worst_level(status, await do_upload(context))
```
escapewindow: Does this mean that if we cancel the task during `do_run_task` we don't upload the logs at all?
We may want to change this so we set a flag on upload. If we hit `CancelledError` and this flag isn't set, we upload the minimum: possibly `cot.log` and `live_backing.log` if they exist; probably skip cot generation + signing?

scriptworker/test/test_worker.py
```diff
@@ -254,3 +255,296 @@ def _mocker_run_tasks_helper(mocker, exc, func_to_raise):
...
+
+
+@pytest.mark.asyncio
+async def test_run_tasks_cancel_cot(context, mocker):
+    context.config['verify_chain_of_trust'] = True
+
+    slow_function_called, slow_function = create_slow_async()
+    mocker.patch('scriptworker.worker.verify_chain_of_trust', slow_function)
+
+    mock_prepare_task = mocker.patch('scriptworker.worker.prepare_to_run_task')
```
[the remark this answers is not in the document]
escapewindow: These functions are pretty core to the worker, so as mentioned below? even clean arch can't save us from lots of i/o in the core functions. We may be able to mitigate this with kwargs as mentioned in chat.

It may be worthwhile adding some integration tests around cancelling tasks. We already have `test_cancel_task`, but we could verify uploads happen as expected etc.

scriptworker/test/test_worker.py
```diff
@@ -60,27 +61,32 @@ def test_main(mocker, context, event_loop):
...
         # Send SIGTERM to ourselves so that we stop
         os.kill(os.getpid(), signal.SIGTERM)
-        return True
 
     try:
         _, tmp = tempfile.mkstemp()
         with open(tmp, "w") as fh:
-            json.dump(config, fh)
-        del(config['credentials'])
+            json.dump(context.config, fh)
```
escapewindow: do you mean `context` or `main_context` here?

scriptworker/task.py
```diff
@@ -7,23 +7,23 @@
...
 import logging
 import os
 import pprint
 import re
-import signal
+from asyncio.subprocess import PIPE
+from copy import deepcopy
 from urllib.parse import unquote, urlparse
 
+import aiohttp
```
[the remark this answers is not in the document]
mitchhentges: I blame Pycharm: I probably hit "optimize imports", which appears to like to group all the `from * import *` together

scriptworker/task.py
```diff
@@ -414,13 +414,14 @@ def prepare_to_run_task(context, claim_task):
 
 
 # run_task {{{1
-async def run_task(context):
+async def run_task(context, to_cancellable_process):
```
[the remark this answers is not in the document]
mitchhentges: Do any dependent packages call `run_task`? That isn't part of the `scriptworker` client :thinking:

scriptworker/worker.py
```diff
@@ -5,30 +5,34 @@
...
 import logging
 import os
-import sys
 import signal
+import sys
+import types
+from asyncio import CancelledError
+
+import aiohttp
+import arrow
```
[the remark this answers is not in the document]
mitchhentges: Thanks pycharm :clap:

scriptworker/test/test_worker.py
mitchhentges: I'm nervous about putting some of these in `__init__.py`, since they're not complete.
For example, if I moved `_MOCK_CLAIM_WORK_RETURN` to `__init__.py`, then either:
* Someone uses it in another test before realizing that the internal `tasks` value isn't fully filled out
* I add complete, correct mock data for `_MOCK_CLAIM_WORK_RETURN`, but not all of it is used. Upon future updates, we need to update (potentially unneeded) mock data, and it increases the "input" for the test (as a reader, it's less obvious what is needed to make the test pass)

My personal preference is to lean towards duplication if it makes individual tests smaller and easier to comprehend. When there are multiple tests that require full mocks, perhaps those full mocks could be shared?

scriptworker/worker.py
mitchhentges: I'm not sure I fully understand.
FWIW, our cancellable states are:
* `claim_work`: When cancelled, we jump to `except CancelledError` and return
* `sleep`: When cancelled, we jump to `except CancelledError` and return
* `verify_chain_of_trust`: When cancelled, skip running the task itself. Upload artifacts to Taskcluster and inform of `worker-shutdown` status (note that the `CancelledError` is caught in `do_run_task`!)
* Task process (within `run_task`): When cancelled, kill the process. Upload artifacts to Taskcluster and inform of `worker-shutdown` status (note that `CancelledError` is not raised! The process is killed, which triggers the follow-up behaviour)

I agree that perhaps the post-task behaviour should perhaps be different if the task was cancelled partially through a build - perhaps just: {upload `cot.log`, upload `live_backing.log`, set task status}, yeah. Should that be handled here, or in follow-up? I'm flexible :)

scriptworker/test/test_worker.py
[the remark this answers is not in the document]
mitchhentges: True that these functions are pretty core! Sometimes I just need to complain after setting up a bunch of mocks to let off steam :wink:
I'll update the cot-cancellation and task-cancellation tests such that they verify that `do_upload` is called, too :)
I'll add an integration test as well, good call

scriptworker/test/test_worker.py
mitchhentges: I'm pretty sure `context` is what I want here - `main_context` is a bit of a workaround since we need to write everything to a file, so that `worker.main` can read that file into a new `context` variable. I needed to slap `running_tasks = MockRunTasks()` onto it, and I can't encode `MockRunTasks()` into a file, so I couldn't put it on beforehand. The solution was to:
* Differentiate context before being written to file (`context`) and context after being read from the file (`main_context`)
* Modify the instance of context used in `_handle_sigterm` to that it would trigger `cancel()`

scriptworker/task.py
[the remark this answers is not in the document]
escapewindow: It's a minor nit, but I'd prefer it.

scriptworker/task.py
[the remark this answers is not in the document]
escapewindow: They shouldn't, but any non-underscore-prefixed functions are public.

scriptworker/test/test_worker.py
escapewindow: I didn't necessarily mean all of them, just the ones that make sense. The `create_{a,}sync` methods seem generic and useful enough that we could either share them or share them and replace `noop_{a,}sync` usage.

scriptworker/worker.py
escapewindow: I think you understand with the last paragraph. I basically meant that, unless I'm misreading, a cancel in the middle of the task will not hit the `do_upload` call. We may want an upload call after a cancel.

Maybe the upload call post-cancel should be a best effort upload:
* allow for missing artifacts (one of the logs may not exist?)
* allow for already-uploaded artifacts, in case we were killed mid- or post- upload

I'm flexible as well, since I think current behavior is `claim-expired` up to 20 min later with no logs. If the rollout process is lengthy enough for this change, it might be worth having both changes landed at the same time, but that's not a hard requirement.

scriptworker/test/test_worker.py
escapewindow: Ok. This isn't clear during a casual read of the test. Could you add a comment, maybe similar to the above, to the test?

scriptworker/worker.py
mitchhentges: Ah, gotcha!
With the implementation in this PR:
* If there's a cancel in the middle of a task, `do_upload` **is** still called
* `do_upload` can't be cancelled. Only those four states (`claim_work`, `sleep`, `verify_chain_of_trust` and the child task process can be cancelled)

escapewindow: Ok, misread, I thought we raised a `CancelledTask` which would jump to the `except`.
If we run `do_upload` regardless, we may want to change its behavior based on whether we cancelled (don't upload artifacts other than logs)

mitchhentges: Ah, yeah, very much on the same page!

I'm less confident that the cost-vs-benefit is worth it for attempting to add artifacts up until being killed - if the job failed due to `worker-shutdown`, and is being rescheduled to be done properly, I think that the potentially-incomplete/corrupt artifacts won't have much value. Besides, once we start dancing purposely close to being killed, it becomes tougher (what if we're killed in the middle of uploading? Are our operations atomic?).

Eh, as you said, this would be follow-up/discussion for later anyways :)

escapewindow: > I'm less confident that the cost-vs-benefit is worth it for attempting to add artifacts up until being killed

This already happens: signingscript copies artifacts into `artifact_dir` as they're ready, and `do_upload` uploads everything in `artifact_dir`. We'd have to add logic to not upload the artifacts if that risks hitting the 30s mark. Potential followup, yes.

scriptworker/test/test_worker.py
mitchhentges: I'm going to leave this one here: It's only used in the one test, and has `cancel()`'s behaviour set accordingly. I'm not sure it's useful to share, and I think there's benefit in it being within the test definition: it hints to the next dev that it's only used in that one test.

Let me know what you think

scriptworker/worker.py
mitchhentges: I've adjusted the behaviour such that we only upload `chain_of_trust.log` and `live_backing.log` if the worker shuts down :+1:

scriptworker/test/test_worker.py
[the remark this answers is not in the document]
mitchhentges: Integration test added!

mitchhentges:
I'll update any failing tests/missing coverage/weird lints tomorrow, but this should be ready for re-review :)
As for roll-out, I'm considering:
1. Roll out to mobile workers (signing + pushapk)
2. (??? Anything in-between? Can we roll out to dep workers? Up to you)
3. Roll out to remaining workers

escapewindow:
It looks like the `do_upload` calls are broken in the tests.

scriptworker/worker.py
```diff
@@ -12,23 +12,25 @@
...
-from scriptworker.utils import cleanup, rm
+from scriptworker.task_process import TaskProcess
+from scriptworker.utils import cleanup, rm, filepaths_in_dir
 
 log = logging.getLogger(__name__)
 
 
 # do_run_task {{{1
-async def do_run_task(context):
+async def do_run_task(context, run_cancellable, to_cancellable_process):
```
escapewindow: We'll need to update the docstring here.

scriptworker/worker.py
```diff
@@ -90,8 +96,86 @@
...
+        if self.future is not None:
+            self.future.cancel()
+        if self.task_process is not None:
+            log.warning("Worker is shutting down, but a task is running. Terminating task")
+            await self.task_process.worker_shutdown_stop()
+
+
 # run_tasks {{{1
-async def run_tasks(context, creds_key="credentials"):
+async def run_tasks(context):
```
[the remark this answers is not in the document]
mitchhentges: I removed `creds_key` because it's [not being used in `run_tasks(...)`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/worker.py#L94-L132)

scriptworker/worker.py
mitchhentges: Surprised `flake8` hasn't caught that - maybe it will, once tests pass

scriptworker/test/test_integration.py
```diff
@@ -164,7 +165,7 @@ def get_temp_creds(context):
         assert result['status']['state'] == 'pending'
         async with remember_cwd():
             os.chdir(os.path.dirname(context.config['work_dir']))
-            status = await worker.run_tasks(context, creds_key="integration_credentials")
+            status = await worker.run_tasks(context)
```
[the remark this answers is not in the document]
mitchhentges: I removed these due to the parameter [not being used in the `run_tasks(...)` function](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/worker.py#L94-L132)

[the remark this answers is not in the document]
escapewindow: You're right. Looks like it's been useless since https://github.com/mozilla-releng/scriptworker/commit/9b8dcb3f5f8e8f3ff5f76927694de972ad769334#diff-9baccc65625cbceb2579ef438f4e18c2 =\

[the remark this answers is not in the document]
mitchhentges: Based on the top of `test_integration.py`, it's definitely [using `integration_credentials` from the `secrets.json` file](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/test/test_integration.py#L40) :)

scriptworker/worker.py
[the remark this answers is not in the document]
escapewindow: https://github.com/mozilla-releng/scriptworker/pull/310#discussion_r256623365 . [`read_integration_creds`](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/test/test_integration.py#L39) appears to do that for us. So... docstring update, and fix the tests, and then we're good :)

mitchhentges:
Builds are green, just need your official Github "approval" before I can press the big cool button

mitchhentges:
At the same time as scriptworker is updated in `build-puppet`, merge [this PR](https://github.com/mozilla-releng/build-puppet/pull/387)

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [19.0.0] - 2019-02-13
+### Added
+- added `running_tasks` property to `Context`
+- added `WorkerShutdownDuringTask` exception
```
mitchhentges: Hmm, I know that this exception is "visible" to implementers, but it should never be raised by them, nor will they every catch a raised `WorkerShutdownDuringTask`. 

I suppose splitting the scriptworker client and worker will resolve this :)

escapewindow: tree, signing, shipit, pushsnap, bouncer, beetmover, addon all show up in a `ag scriptworker *script | grep import | grep -v client`

Issue #273, opened 2018-11-20
Chain of Trust: Restrict git/hg branches 


JohanLorenzo:
Today, @rpappalax came to me with an interesting case: he has some big changes to make on the [focus-android](https://github.com/mozilla-mobile/focus-android/) taskgraph and he's willing to use a github branch (living the focus-android repo). This branch needs some taskcluster scopes.
He asked @imbstack to expand the following role: https://tools.taskcluster.net/auth/roles/repo%3Agithub.com%2Fmozilla-mobile%2Ffocus-android%3Abranch%3A*. Without too much context, all the scopes needed to sign and publish a Focus build ended up in the role (I actually wonder if these scopes were on the master role). 

## Immediate fix

I updated to role to strip all the signing/publishing scopes out. As of now, no one can publish Focus from a random branch on the main repo.

## What scriptworker can do

In hg.mozilla.org, this is safeguarded by [cot_restricted_scopes](https://github.com/mozilla-releng/scriptworker/blob/ae4cc4de36bd59a235c7a48e17d15c827caa3024/scriptworker/constants.py#L217-L269). Sadly, scriptworker doesn't allow to restrict branches within an existing repo. This case has never popped up on hg.mozilla.org, but we now have a case on github.

To me, scriptworker can enforce that. scriptworker knows what's the commit under checks. We can let it know what's the branch it's on. scriptworker will then verify against:
* github.com by fetching and parsing https://github.com/:user/:repo/branch_commits/:commit (for instance: https://github.com/mozilla-mobile/focus-android/branch_commits/1d0f51cd43f45a25b34191a5276d3f5d98023899). I [couldn't find](https://stackoverflow.com/questions/29992441/github-api-identify-the-branches-that-contain-a-given-commit) an API endpoint that exposes this piece of data.
* hg.mozilla.org withhttps://hg.mozilla.org/releases/mozilla-release/json-pushes?full=1&changeset=tip 

This way scriptworker knows the branch against which the graph was triggered, and it knows the branch isn't a lie. Therefore, scriptworker can restrict by git/hg branches.

@escapewindow @tomprince, how does this look to you?

imbstack:
The new role is just a copy-paste of the old role that was restricted to the master branch. I was suspicious that these scopes were not appropriate and did ask a good bit about whether or not those scopes were appropriate for every branch. rpappa/colee decided that it was fine. We can restrict the role to the master branch again if these scopes are not appropriate for every branch.

[...]

PR #313, opened 2019-02-22
Bug 1529990 - Fix Mobile bustage

JohanLorenzo:
And add more checks to detect PRs

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -121,19 +121,18 @@ class ChainOfTrust(object):
         """
         return [x.task_id for x in self.links]
 
-    def is_try_or_pull_request(self):
+    async def is_try_or_pull_request(self):
         """Determine if any task in the chain is a try task.
 
         Returns:
             bool: True if a task is a try task.
 
         """
-        result = is_try_or_pull_request(self.context, self.task)
-        for link in self.links:
-            if link.is_try_or_pull_request:
-                result = True
-                break
-        return result
+        tasks = [asyncio.ensure_future(link.is_try_or_pull_request()) for link in self.links]
+        tasks.insert(0, asyncio.ensure_future(is_try_or_pull_request(self.context, self.task)))
+
+        conditions = await raise_future_exceptions(tasks)
+        return any(conditions)
 
     def get_link(self, task_id):
         """Get a ``LinkOfTrust`` by task id.
@@ -243,10 +242,9 @@ class LinkOfTrust(object):
         self.parent_task_id = get_parent_task_id(self.task)
         self.worker_impl = guess_worker_impl(self)
 
-    @property
-    def is_try_or_pull_request(self):
+    async def is_try_or_pull_request(self):
         """bool: the task is either a try or a pull request one."""
-        return is_try_or_pull_request(self.context, self.task)
+        return await is_try_or_pull_request(self.context, self.task)
 
     @property
     def cot(self):
@@ -2072,7 +2070,7 @@ async def trace_back_to_tree(chain):
                     obj.name, obj.task_id
                 ))
     # Disallow restricted privs on is_try_or_pull_request.  This may be a redundant check.
-    if restricted_privs and chain.is_try_or_pull_request():
+    if restricted_privs and await chain.is_try_or_pull_request():
         errors.append(
             "{} {} has restricted privilege scope, and is_try_or_pull_request()!".format(
                 chain.name, chain.task_id
--- a/scriptworker/github.py
+++ b/scriptworker/github.py
@@ -1,6 +1,17 @@
 """GitHub helper functions."""
 
+import re
+
 from github3 import GitHub
+
+from scriptworker.exceptions import ConfigError
+from scriptworker.utils import (
+    get_single_item_from_sequence,
+    get_parts_of_url_path,
+    retry_request,
+)
+
+_GIT_FULL_HASH_PATTERN = re.compile(r'^[0-9a-f]{40}$')
 
 
 class GitHubRepository():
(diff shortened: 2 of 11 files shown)


scriptworker/cot/verify.py
```diff
@@ -244,9 +243,9 @@ def task(self, task):
         self.worker_impl = guess_worker_impl(self)
 
     @property
-    def is_try_or_pull_request(self):
+    async def is_try_or_pull_request(self):
```
escapewindow: https://stackoverflow.com/questions/36666151/asynchronous-property-setter

scriptworker/task.py
```diff
@@ -462,10 +455,33 @@ def is_try_or_pull_request(context, task):
...
+        task (dict): the task definition to check.
+
+    Returns:
+        bool: True if a piece of data refers to GitHub
+
     """
     return any((
-        is_try(task, context.config['source_env_prefix']),
-        is_pull_request(context, task),
+        # XXX Cron tasks don't usually define 'taskcluster-github' as their schedulerId as they
```
escapewindow: Once we use standalone taskgraph in our github repos, we can look at standardizing how the taskgraph generates cron tasks so we can detect it here.

JohanLorenzo:
I briefly tested this PR against production: 

* Cron jobs: 
 * of the reference-browser https://tools.taskcluster.net/groups/NnVqnbC-QAWXHrYq3xlfKw/tasks/Wo2ZpFRjSZCT5WlhQ772xQ/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L113
 * of Fenix https://tools.taskcluster.net/groups/RBRW5HYQR7KCs5Te-kz3mA/tasks/YX31_U6mRmuQKzK4RI7bew/runs/2/logs/public%2Flogs%2Fchain_of_trust.log#L114
 * of Focus for Android https://tools.taskcluster.net/groups/MCqjnEZyQomPglWXNlqICA/tasks/e_qxXAdOTMG1jGelJnu_KA/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L119 
 *  https://tools.taskcluster.net/groups/SecY7jv7Q4GU5PjFPyfdsw/tasks/dMA6kPu0S6OTjTZHtJSJAQ/runs/6/logs/public%2Flogs%2Fchain_of_trust.log#L102

scriptworker/github.py
```diff
@@ -65,3 +76,152 @@ def get_release(self, tag_name):
...
+
+    Args:
+        url (str): The URL to test. It can be None.
+
+    Returns:
+        bool: False if the URL is not a string or if it doesn't match a Github URL
+
+    """
+    if isinstance(url, str):
+        return url.startswith('https://github.com/')
```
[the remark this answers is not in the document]
JohanLorenzo: As of not, I'm not aware of the usage of `git@github.com` nor `ssh://` within chain of trust. Let's add it whenever the need comes up, I believe

scriptworker/task.py
```diff
@@ -422,35 +391,59 @@ def is_try(task, source_env_prefix):
...
-    This goes further than checking ``tasks_for``.  We may or may not want
+    This goes further than checking ``tasks_for``. We may or may not want
     to keep this.
 
     This checks for the following things::
 
         * ``task.extra.env.tasks_for`` == "github-pull-request"
         * ``task.payload.env.MOBILE_HEAD_REPOSITORY`` doesn't come from an official repo
         * ``task.metadata.source`` doesn't come from an official repo, either
+        * The last 2 items are landed on the official repo
```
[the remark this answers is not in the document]
JohanLorenzo: Agreed. This implementation paves the way to https://github.com/mozilla-releng/scriptworker/issues/273. We just need to parse the bits of HTML returned by the `branch_commits` endpoint and we can know what branch the commit lives on. 

I'm not sure how we can  check if it's part of the last N commits of a branch, yet. The real github APIv3 may have something already.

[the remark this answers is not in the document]
escapewindow: Ah, I thought it might be easier to detect the last N commits of a branch, rather than just knowing which branch(es) a commit is on. If the latter is easier, it's preferable.

scriptworker/github.py
```diff
@@ -65,3 +76,152 @@ def get_release(self, tag_name):
...
+    if not official_repo_owner:
+        raise ConfigError(
+            'This worker does not have a defined owner for official GitHub repositories. '
+            'Given "official_github_repos_owner": {}'.format(official_repo_owner)
+        )
+
+    return official_repo_owner == repo_owner
+
+
+def _is_git_full_hash(commit_hash):
```
mitchhentges: Nit: since the input of this function isn't always a commit hash (it may be a branch or a tag), perhaps this should be called "revision" :)

scriptworker/github.py
```diff
@@ -65,3 +76,152 @@ def get_release(self, tag_name):
...
+    return _GIT_FULL_HASH_PATTERN.match(commit_hash) is not None
+
+
+def _check_github_url_is_supported(url):
+    if not is_github_url(url):
+        raise ValueError('"{}" is not a supported GitHub URL!'.format(url))
```
mitchhentges: Nit that isn't relevant for this PR: I wonder if it would make sense to have earlier validation for things like this?
Currently, we don't need `if` checks to assert that our variables are strings/numbers/whatever deep within our functions because we perform earlier validation of the task, and that allows the functions that operate on the task to be so much easier to read (and write).

In the same way. it could be beneficial to have validation of github urls in advance? It _could_ make logic cleaner, since we could leverage expectations with a data type (e.g.: "my function parameter is `GithubUrl`, so it better be a Github URL") rather than a function call ("what does `_check_github_url_is_supported(...)` do? Oh, it can throw an error, gotcha").

-----

This is a potentially subjective note. Not that big of a concern :)

scriptworker/task.py
```diff
@@ -422,35 +391,59 @@ def is_try(task, source_env_prefix):
...
 
     Args:
         context (scriptworker.context.Context): the scriptworker context.
         task (dict): the task definition to check.
 
     Returns:
-        bool: True if it's a pull-request
+        bool: True if it's a pull-request. False if it either comes from the official repos or if
+        the origin can't be determined. In fact, some valid scriptworker tasks don't
```
mitchhentges: I think that this comment is cut off?

scriptworker/task.py
```diff
@@ -422,35 +391,59 @@ def is_try(task, source_env_prefix):
...
+        extract_github_repo_and_revision_from_source_url(metadata_source_url)
+
+    conditions = [tasks_for == 'github-pull-request']
+    urls_revisions_and_can_skip = (
+        (repo_url_from_payload, revision_from_payload, True),
+        (repo_from_source_url, revision_from_source_url, False),
+    )
+    for repo_url, revision, can_skip in urls_revisions_and_can_skip:
+        # XXX In the case of scriptworker tasks, neither the repo nor the revision is defined
+        if not repo_url and can_skip:
```
mitchhentges: What happens if `not repo_url` and `can_skip is False`? We expect `extract_github_repo_and_name(...)` to throw an error?

scriptworker/task.py
```diff
@@ -422,35 +391,59 @@ def is_try(task, source_env_prefix):
...
+            continue
+
+        repo_owner, repo_name = extract_github_repo_owner_and_name(repo_url)
+        conditions.append(not is_github_repo_owner_the_official_one(context, repo_owner))
+
+        if not revision and can_skip:
+            continue
 
+        github_repository = GitHubRepository(repo_owner, repo_name, context.config['github_oauth_token'])
+        conditions.append(not await github_repository.has_commit_landed_on_repository(context, revision))
```
mitchhentges: Nice :)

scriptworker/github.py
tomprince: If, as you suggest, we weren't using strings everywhere (particularly if we could lean on type-checking), we could get away without these checks.

However, given that this is security sensitive code, I'd rather err on the side of having more sanity checks, rather than less.

mitchhentges: FWIW, I'm not recommending removing these checks, but rather moving them to the entry-point of untrusted data and doing validation there.
To represent that the string has been validated as the specialized type (github url string), we could use a type at that point.

But yeah, I agree, I don't think that we should lessen our sanity checks :)

scriptworker/task.py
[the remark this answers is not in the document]
JohanLorenzo: Mmm, I thought there was a way to get the last N commits of a branch, but this doesn't seem to be exposed after all:  https://developer.github.com/v3/repos/branches/

`branch_commits` is our best alternative.

scriptworker/github.py
JohanLorenzo: Good call. Fixed.

scriptworker/github.py
```diff
@@ -65,3 +76,152 @@ def get_release(self, tag_name):
...
+            context (scriptworker.context.Context): the scriptworker context.
+            revision (str): the commit hash or the tag name.
+
+        Returns:
+            bool: True if the commit is present in one of the branches of the main repository
+
+        """
+        # Revision may be a tag name. `branch_commits` doesn't work on tags
+        if not _is_git_full_hash(commit_hash=revision):
+            revision = self.get_tag_hash(tag_name=revision)
```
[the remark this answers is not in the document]
JohanLorenzo: At the moment, that's not a supported case. `get_tag_hash()` will raise an error because the tag doesn't exist. I'm not aware of a case where we want to support it. If that comes up, I'm fine adding this support.

scriptworker/github.py
```diff
@@ -65,3 +76,152 @@ def get_release(self, tag_name):
...
+
+        repo = self._github_repository.html_url
+
+        url = '/'.join([repo.rstrip('/'), 'branch_commits', revision])
```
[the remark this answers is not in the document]
JohanLorenzo: Yeah, that's exactly what happened! :sweat_smile:

scriptworker/task.py
JohanLorenzo: Oops. Fixed.

scriptworker/task.py
JohanLorenzo: That's right. `repo_url` won't be a valid github URL and then `_check_github_url_is_supported()` raises a `ValueError`

scriptworker/github.py
JohanLorenzo: Agreed on keeping this check here, for now.

JohanLorenzo:
Chatted with @tomprince, yesterday on Slack: he's fine with me landing the patch.

PR #314, opened 2019-03-08
fix integration tests for taskcluster-client.py>=7.0.0 and slugid>=2.0.0

escapewindow:


escapewindow:
> This LGTM!
> TIL there is taskcluster-client >= 7 😳

I just released it earlier today :)
Thanks!

PR #315, opened 2019-03-08
verify_cot --verify-sigs option

escapewindow:
ed25519 + the known level 3 pubkeys makes this a viable thing to do locally. Currently it only works with level 3 graphs, since we don't have the level 1 pubkeys anywhere.

This will fail if we have to fall back to gpg signature verification.


escapewindow:
Included in #320.

PR #317, opened 2019-03-17
bug 1400258 - write cot logs to live_backing.log

escapewindow:


escapewindow:
Essentially #232 again. Github made me open a new PR since I force-pushed.

escapewindow:
@tomprince do you still have objections to this?
We seem to follow this pattern in other worker implementations, e.g. the docker-worker task live_backing.log also containing vcs log information before the task command runs.

tomprince:
I'm still sad that this seems to be the best solution, but given that treeherder doesn't have support for multiple logs, this is probably the correct solution.

jcristau:
Superseded by #796.

PR #252, opened 2018-09-06
Fix verify_cot command line script by making sure to instantiate the …

Callek:
…Taskcluster Queue.

PR #320, opened 2019-03-26
bug 1534506 - remove gpg support from chain of trust

### LINKED ISSUE #311: Put branch coverage back to 100%
There are 2 places we miss, currently:
* https://travis-ci.org/mozilla-releng/scriptworker/jobs/493004995#L1226
* https://travis-ci.org/mozilla-releng/scriptworker/jobs/493004995#L124

Spotted in #307 by @escapewindow

escapewindow:
This is a fairly large change to remove gpg support from the chain of trust.
Now that we have ed25519 signature support across all release trains, we can remove gpg support, and its associated gpg key management complexity.

For the most part, this is removal. I have also:

- added a `verify_cot --verify-sigs` option. This was not an option in gpg, since local gpg homedir management was too much overhead. With the ed25519 level 3 pubkeys in `scriptworker.constants`, we can easily verify these signatures locally.
- added cot signature verification in integration tests. Because we're looking at level 3 release train indexes, we can assume that these tasks all have valid level 3 ed25519 signatures.
  - added an esr60 integration test.
  - switched integration from using linux64 repackage-signing tasks which appear to be obsoleted with the mar-signing kind. Instead, use win64 repackage-signing tasks.
- ported the fix from #252, which I couldn't replicate until I moved away `./secrets.json` and `~/.scriptworker`.
- made the `chain-of-trust.json` cot artifact mandatory. `chain-of-trust.json.sig` is also mandatory if signature verification is on. (`chainOfTrust.json.asc` is removed with the other gpg refs.)
- got us back to 100% branch coverage in https://github.com/escapewindow/scriptworker/commit/54f7e29a480427bfa5d9c8c8ee7ba9fc9bec575b (fixes #311 )


scriptworker/worker.py
```diff
@@ -239,6 +239,8 @@ def main(event_loop=None):
         done = True
         if context.running_tasks is not None:
             await context.running_tasks.cancel()
+        # for test coverage
+        sys.exit(1)
```
escapewindow: @mitchhentges what do you think about this? This allows us to get full test coverage. Just want to make sure you don't think it'll cause any weirdness.

escapewindow:
@tomprince do you have time/headspace for this review? I'm happy to answer questions or split the commits differently as desired.

scriptworker/worker.py
mitchhentges: Hey, thanks for pinging me!

I'm not sure that this is a good idea: this signal handler is called when the main process is interrupted (and we don't know where the process was interrupted at). I think it's a lot safer to let the application exit on its own.

I wonder how this `sys.exit(1)` is improving test coverage: is it because it impedes the test runner from exiting normally? :thinking:

escapewindow: It's safer, but is it harmful? If it is, I can remove.
It's been a while, but we see the lack of coverage [here](https://travis-ci.org/mozilla-releng/scriptworker/jobs/493004995#L1240). Also, I vaguely remember the test just running forever.

Sounds like I might need to remove this line and see what comes up.

escapewindow: It looks like it [works without the `sys.exit`](https://github.com/escapewindow/scriptworker/commit/4d7f838e82a8a3b073fb2b194af52a20da4551c3). We're able to more clearly see which branch we're hitting in the test, because we have slightly different behavior in the `running` branch than if not running. But we can see that we have full coverage, and that we exit the main loop, so perhaps that's sufficient.

docs/cot_verify.rst
```diff
@@ -14,42 +14,13 @@ The decision task is a special task that generates a taskgraph, then submits it
...
-For flat directories, scriptworker imports all pubkeys from the corresponding directory, and signs them to mark the pubkeys as valid.  This allows us to verify the signature on the signed chain of trust json artifacts.
-
-``signed`` directories
-^^^^^^^^^^^^^^^^^^^^^^
-
-This is currently only for scriptworker.
-
-- scriptworker imports all pubkeys in the ``trusted/`` subdirectory, signs them, and marks them as trusted.
-- scriptworker imports all pubkeys in the ``valid/`` subdirectory.  They should already be signed by one of the keys in the ``trusted/`` subdirectory, so scriptworker doesn't otherwise sign or mark them as valid.
+We keep the level 3 gecko pubkeys in ``scriptworker.constants.ed25519_public_keys``, as base64-encoded ascii string. Once decoded, these are the seeds for the ed25519 public keys. These are tuples of valid keys, to allow for key rotation.
```
tomprince: string**s**?

scriptworker/cot/verify.py
```diff
@@ -608,24 +606,18 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
 
     """
     mandatory_artifact_tasks = []
```
tomprince: I'd consider removing the `mandatory_` here, since that is the only thing now (but also understand if you'd rather do that separately).

docs/cot_gpg_key_mgmt.rst
```diff
@@ -1,103 +0,0 @@
...
--  [taskcluster team] worker keys, which are unsigned pubkeys for
-   docker- and generic- workers.
--  [releng team] scriptworker keys, which are signed pubkeys for
-   scriptworkers.
--  [releng team] scriptworker trusted keys, which are the pubkeys of
-   releng team members who are allowed to generate and sign scriptworker
-   keys.
--  [various] git commit signing keys. We keep the above pubkeys in a git
-   repo, and we sign the commits. These are the pubkeys that are allowed
-   to sign the git commits.
```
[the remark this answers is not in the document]
escapewindow: It's almost all gpg specific, but I can add a few sentences about ed25519 key management.

scriptworker/worker.py
mitchhentges: Weird :thinking: 

FWIW, regarding safer/harmful: I believe that the harm of having the `sys.exit` would be that if/when it affects things negatively, it may be really tricky to determine the cause of the negative behavour. Like, since the main process is interrupted when the signal handler is called, that interruption could happen in the middle of a volatile function (something OS-related? Who knows what the possibilities are). I attribute a high cost to any risks that might cause issues that are incredibly difficult to debug.

PR #322, opened 2019-03-27
make paths unique in get_all_artifacts_per_task_id

### LINKED ISSUE #321: only download each taskId/path combination once
In [this task](https://taskcluster-web.netlify.com/tasks/H5zF-GnQRBWe6ZSXss3kIA/runs/0/) , we download [this artifact](https://queue.taskcluster.net/v1/task/ToJ-6HK2RmaB9Sd7eNGM8g/artifacts/public%2Ftarget-langpack.checksums) 5 times, because it's specified in upstreamArtifacts 5 times under different `locale`s. When this works, it's wasteful in terms of network and compute time; when it doesn't work, it has 5x as many chances to fail (esp if we check the hash while another coroutine overwrites the artifact).

We should create a set of `taskId`/`path` combinations to ensure they're unique, and then download those cot artifacts.

escapewindow:
also use setdefault in add_enumerable_item_to_dict
fixes #321

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -795,6 +795,10 @@ def get_all_artifacts_per_task_id(chain, upstream_artifacts):
                 dict_=all_artifacts_per_task_id, key=upstream_dict['taskId'], item=upstream_dict['paths']
             )
 
+    # Avoid duplicate paths per task_id
+    for task_id, paths in all_artifacts_per_task_id.items():
+        all_artifacts_per_task_id[task_id] = sorted(list(set(paths)))
+
     return all_artifacts_per_task_id
 
 
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -770,7 +770,7 @@ def get_all_artifacts_per_task_id(chain, upstream_artifacts):
         upstream_artifacts: the list of upstream artifact definitions
 
     Returns:
-        dict: list of paths to downloaded artifacts ordered by taskId
+        dict: sorted list of paths to downloaded artifacts ordered by taskId
 
     """
     all_artifacts_per_task_id = {}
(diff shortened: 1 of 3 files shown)


escapewindow:
@JohanLorenzo I think you wrote most of this original code -- we're trying to avoid duplicate downloads. I also revisited `add_enumerable_item_to_dict`.

@tomprince we discussed this in #releaseduty.

Non-urgent. If either of you have the time and headspace to review, this should help reduce unnecessary churn on tasks with duplicate paths per taskId (we currently know some beetmover-checksums tasks do this). Thanks!

scriptworker/cot/verify.py
```diff
@@ -795,6 +795,10 @@ def get_all_artifacts_per_task_id(chain, upstream_artifacts):
                 dict_=all_artifacts_per_task_id, key=upstream_dict['taskId'], item=upstream_dict['paths']
             )
 
+    # Avoid duplicate paths per task_id
+    for task_id, paths in all_artifacts_per_task_id.items():
+        all_artifacts_per_task_id[task_id] = sorted(list(set(paths)))
```
tomprince: This can be simplified to `sorted(set(paths)`; `sorted` accepts an iterable an returns a list.

scriptworker/utils.py
```diff
@@ -713,16 +713,11 @@ def add_enumerable_item_to_dict(dict_, key, item):
         item (whatever): The item to add to the list associated to the key
 
     """
+    dict_.setdefault(key, [])
```
JohanLorenzo: 👍

scriptworker/cot/verify.py
JohanLorenzo: TIL :)

scriptworker/cot/verify.py
```diff
@@ -795,6 +795,10 @@ def get_all_artifacts_per_task_id(chain, upstream_artifacts):
                 dict_=all_artifacts_per_task_id, key=upstream_dict['taskId'], item=upstream_dict['paths']
```
JohanLorenzo: Nit: Add `sorted in the pydoc. For instance: 

```py
"""
    Returns:
        dict: list of sorted paths to downloaded artifacts ordered by taskId
"""
```

scriptworker/cot/verify.py

scriptworker/cot/verify.py

escapewindow:
Thanks!

PR #325, opened 2019-04-10
Bug 1514903 - Fix .tar.gz uploads

escapewindow:
Currently, when we download a gzip-compressed tarball (e.g. target.tar.gz or target.tgz) that has been uploaded by scriptworker, the encoding is `gzip`, so the tarball is uncompressed on download. This breaks chain of trust verification, because the sha256 changes when the tarball is uncompressed.

To fix, we should upload the gzip tarball with `('application/x-tar', None)`, so we download the compressed tarball without uncompressing. I stopped using `posixpath.splitext` because that returns `.gz` for a `.tar.gz` (not what we want), and instead use `.endswith`.

There are two ridealong fixes I can unbundle if preferred:

- `s,scriptharness,scriptworker,` in the docs, and
- a fix for `verify_cot` without a secrets.json or ~/.scriptworker defined. This would error out on a missing `rootUrl` in the `Queue` object. We should grab this from the environment, with a sane default.

@mitchhentges , you touched the mimetypes code last. Are you up for reviewing?

(This is tested on Try: when signing-linux-dev1 has this patchset applied, .tar.gz files pass CoT verification.)


scriptworker/cot/verify.py
```diff
@@ -2154,7 +2154,12 @@ def format(self, record):
         context.session = session
         context.config = dict(deepcopy(DEFAULT_CONFIG))
         context.credentials = read_worker_creds()
-        context.queue = context.queue or Queue(session=session)
+        context.queue = context.queue or Queue(
+            session=session,
+            options={
+                'rootUrl': os.environ.get('TASKCLUSTER_ROOT_URL', 'https://taskcluster.net'),
```
mitchhentges: Nice

escapewindow:
Thanks!

PR #329, opened 2019-04-16
Bug 1536079 - enable scriptworker support for appservices

MihaiTabara:
Initially we thought of considering this as `mobile` but it's not quite a mobile product. Nor is a gecko one, hence for now it's its own thing.

### DIFF
--- a/scriptworker/constants.py
+++ b/scriptworker/constants.py
@@ -155,6 +155,8 @@ DEFAULT_CONFIG = frozendict({
         # We haven't had the need for mobile-2-decision yet
         # https://bugzilla.mozilla.org/show_bug.cgi?id=1512631#c6
         "mobile-3-decision",
+        # this app is neither mobile nor gecko so for now has its own thing
+        "application-services-r",
     ),
 
     # docker-image cot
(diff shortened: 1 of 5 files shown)


scriptworker/constants.py
```diff
@@ -167,6 +169,8 @@
 
         "mobile-1-images",  # there is no mobile level 2.
         "mobile-3-images",
+        # TODO: to change this to something else?
```
JohanLorenzo: I agree to change this to something like `services-X-decision`, or `services-X-images`, in a followup

scriptworker/constants.py
```diff
@@ -402,27 +425,35 @@
...
             'mobile': 'any',  # all allowed
+            'application-services': 'any',  # all allowed
         }),
     },
     'source_env_prefix': {
         'by-cot-product': frozendict({
             'firefox': 'GECKO',
             'thunderbird': 'COMM',
             'mobile': 'MOBILE',
+            'application-services': 'APPSERVICES',
```
JohanLorenzo: I wonder if `APPSERVICES` is too narrow. Do you know what kind of projects would fall under this `cot_product`?

scriptworker/cot/verify.py
```diff
@@ -1061,10 +1060,7 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
     return {
         'event': {
             'repository': {
-                # TODO: Append ".git" to clone_url in order to match what GitHub actually provide.
-                # This can't be done at the moment because some mobile projects still rely on the
-                # bad value
-                'clone_url': repo_url,
+                'clone_url': "{}.git".format(repo_url),
```
JohanLorenzo: 👍

PR #330, opened 2019-04-23
Send OAuth token to GitHub

eoger:
`has_commit_landed_on_repository` doesn't go through `github3` and therefore does not enjoy the rate-limiting upgrade provided by our OAuth token.

This should fix the 429 errors seen [here](https://tools.taskcluster.net/groups/F-eFLR1YTD6dUcnYIxAg6g/tasks/SIrT2zqXSKamP5FEIOq75w/runs/0/logs/public%2Flogs%2Fchain_of_trust.log).

### DIFF
--- a/scriptworker/github.py
+++ b/scriptworker/github.py
@@ -114,7 +114,8 @@ class GitHubRepository():
         repo = self._github_repository.html_url
 
         url = '/'.join([repo.rstrip('/'), 'branch_commits', revision])
-        html_data = await retry_request(context, url)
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
In the meantime, a workaround can be to rerun failed tasks.

JohanLorenzo:
I filed #331 to track the real fix. @MihaiTabara, do you think spinning up a new beetmover instance makes sense?

eoger:
Work to fix the issue in continuing in https://github.com/mozilla-releng/scriptworker/pull/332, thanks for the review @JohanLorenzo !

Issue #331, opened 2019-04-24
`cot.verify` should cache requests made to Github


JohanLorenzo:
See https://github.com/mozilla-releng/scriptworker/pull/330#pullrequestreview-229932915

We should fix errors like:

```
2019-04-23T20:16:03    DEBUG -  GET https://github.com/mozilla/application-services/branch_commits/0a49320a24b2fc826f0153c0627baba1aa94d429
2019-04-23T20:16:03    DEBUG -  GET https://github.com/mozilla/application-services/branch_commits/0a49320a24b2fc826f0153c0627baba1aa94d429
2019-04-23T20:16:04    DEBUG -  GET https://github.com/mozilla/application-services/branch_commits/0a49320a24b2fc826f0153c0627baba1aa94d429
2019-04-23T20:16:04    DEBUG -  GET https://github.com/mozilla/application-services/branch_commits/0a49320a24b2fc826f0153c0627baba1aa94d429
2019-04-23T20:16:04    DEBUG -  Status 429
2019-04-23T20:16:05    DEBUG -  Status 429
2019-04-23T20:16:05    DEBUG -  Status 429
2019-04-23T20:16:05    DEBUG -  Status 429
2019-04-23T20:16:05    ERROR - Hit ScriptWorkerException: Bad status 429
```

One way to solve this is to cache the request made here https://github.com/mozilla-releng/scriptworker/blob/30a54b0c8092698f23b570cf81560f8e1af3322c/scriptworker/github.py#L117. In fact, we don't need to call this URL 4 times in a row.

JohanLorenzo:
For the record, android-components saw about a dozen identical failures, yesterday. For instance:  https://tools.taskcluster.net/groups/JpNiL1RTQau4tf_yYnMriA/tasks/Mj94usLBTo-vXnvL-gQ8YA/runs/1/logs/public%2Flogs%2Fchain_of_trust.log#L105

PR #332, opened 2019-04-24
Cache has_commit_landed_on_repository results

eoger:
It should fix https://github.com/mozilla-releng/scriptworker/issues/331

### DIFF
--- a/scriptworker/github.py
+++ b/scriptworker/github.py
@@ -1,5 +1,6 @@
 """GitHub helper functions."""
 
+import asyncio
 import re
 
 from github3 import GitHub
@@ -12,6 +13,8 @@ from scriptworker.utils import (
 )
 
 _GIT_FULL_HASH_PATTERN = re.compile(r'^[0-9a-f]{40}$')
+_branch_commits_cache = {}
+_branch_commits_cache_lock = asyncio.Lock()
 
 
 class GitHubRepository():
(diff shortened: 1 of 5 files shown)


scriptworker/test/test_github.py
```diff
@@ -103,27 +103,37 @@ def test_get_tag_hash(github_repository, tags, raises, expected):
...
+    '0129abcdef012345643456789abcdef012345678',
+    'https://github.com/some-user/some-repo/branch_commits/0129abcdef012345643456789abcdef012345678',
+    '\r\n\r\n',
+    True,
+    False,
+    False,
 ), (
-    '0123456789abcdef0123456789abcdef01234567',
-    'https://github.com/some-user/some-repo/branch_commits/0123456789abcdef0123456789abcdef01234567',
+    '0f0123456789abcdef012123456789abcde34565',
```
eoger: I made these different because the cache was kicking in and failing the tests.

scriptworker/github.py
```diff
@@ -12,6 +13,8 @@
 )
 
 _GIT_FULL_HASH_PATTERN = re.compile(r'^[0-9a-f]{40}$')
+_branch_commits_cache = {}
```
JohanLorenzo: I figured out why Chain of Trust didn't contact github.com anymore. This variable lives as long as the program runs. Some parts of beetmover_scriptworker are killed after the worker stopped running the task, but not this very one. Therefore, the first time CoT encounters the URL, it does cache the response, but then it keeps it for waaaaay to long. I think we need to cache it differently. 

I know it must be done at the class the level, because the 4 requests we see are spawned by https://github.com/mozilla-releng/scriptworker/blob/646d478420a213544aa1954c24c53893dc1c4f67/scriptworker/task.py#L441

Maybe the easier way is to store the result in scriptworker's work_dir. If I remember correctly, this one gets nuked after a task is processed.

scriptworker/github.py
```diff
@@ -114,8 +117,15 @@ def get_tag_hash(self, tag_name):
         repo = self._github_repository.html_url
 
         url = '/'.join([repo.rstrip('/'), 'branch_commits', revision])
-        html_data = await retry_request(context, url)
-        html_text = html_data.strip()
+        from scriptworker.task import get_decision_task_id
```
eoger: If imported at the top level this will result in a circular dependency problem.

eoger:
@JohanLorenzo wanna have another look?

MihaiTabara:
@JohanLorenzo is PTO until mid-week next-week. I wanted to get to the bottom of this this week, but I'll have to wait until Monday. Sorry about the delays.

MihaiTabara:
Dealing with other fires yesterday, I'll attempt to get to this today. If not, I'll defer to @JohanLorenzo.

scriptworker/github.py
```diff
@@ -114,8 +117,15 @@ def get_tag_hash(self, tag_name):
...
+        cache_key = '{}-{}'.format(get_decision_task_id(context.task), url)
```
JohanLorenzo: I think we might eventually hit an issue because the cache isn't invalidated over time, or because the dictionary takes too much memory. We may want to eventually use https://github.com/tkem/cachetools.

That said, the patch looks good enough to me for now. I'll deploy it and see if fixes the immediate issues we're seeing.

PR #336, opened 2019-05-02
Allow tasks to use indexed-images for non-restricted-scope tasks.

tomprince:


scriptworker/cot/verify.py
```diff
@@ -501,6 +502,7 @@ def verify_docker_image_sha(chain, link):
             else:
                 log.debug("Found matching docker-image sha {}".format(upstream_sha))
     else:
+        # FIXME: Indexed image sould be verified by CoT as well
```
escapewindow: Nit typo: *should

I imagine we'll need to add a check for indexed docker image tasks, and add them to the chain in `find_sorted_task_dependencies`. Then we'll need to compare the artifact hash that the docker image task uploaded with the artifact hash that the downstream task used, to make sure that it is the same artifact. This seems tricky enough that I'd lean towards expanding the index at decision task time, and specify the `taskId` and docker image sha in the task definition. Is there a reason why we're not doing that here?

tomprince: This is used in particular in the decision task itself, so there is no place to run code to do that.

escapewindow: Ah.

In gecko, we're hardcoding the decision image sha in the .taskcluster.yml, which allows us to bypass sha verification: you have to land a patch in-tree to change it. Also, it allows us to stop our verification chain at the decision task. If we:

- have a build B we need to verify
- it's created from a decision task D1
- it's run in a docker image from docker image task DI1
- that docker image task was generated from a decision task D2
- that decision task was built from a docker image from docker image DI2
- ...

Then we have the potential for a long chain, and we have to trust something at the beginning somehow.
Would landing the docker-image sha and url in-tree be doable? If not, we need to brainstorm a new way to end the chain at the decision task.

tomprince: We decided to allow this for tasks that don't require restricted scopes (so try and PRs) for the moment.

PR #347, opened 2019-06-14
add create_test_workdir entrypoint

### LINKED ISSUE #342: create a tool to populate a `work_dir` with `task.json` and `cot/`
Manually testing \*scripts involves either copying a `work_dir` while it's mid-task, or manually populating a `work_dir` with `task.json` and `cot/` to set up the `work_dir` before running the \*script manually.

Let's write a tool that does this for us, and ship it with scriptworker.

Currently I'm thinking

    create-test-workdir --taskid <TASKID> <PATH>

would

1. create an empty directory `PATH`
1. download the task definition of the scriptworker task `TASKID` and drop it in `PATH/task.json` (we can edit this later if wanted, but let's at least get the skeleton here)
1. download all the artifacts in the task's `upstreamArtifacts` into `PATH/cot/<upstreamTaskId>/<upstreamPath>`

and then let the user know that it's ready for someone to run `*script script_config.json`, once the `script_config.json` is created.

At some point we may figure out a better way for scriptworker to pass this info / these artifacts to each \*script, but until then, let's make it easier to test by creating this tool.

escapewindow:
fixes #342

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -70,6 +70,7 @@ from scriptworker.utils import (
     read_from_file,
     remove_empty_keys,
     rm,
+    write_to_file,
 )
 from taskcluster.exceptions import TaskclusterFailure
 from taskcluster.aio import Queue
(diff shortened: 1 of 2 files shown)


escapewindow:
Hm, not yet ready for prime time.

PR #349, opened 2019-06-21
add create_test_workdir entrypoint

### LINKED ISSUE #342: create a tool to populate a `work_dir` with `task.json` and `cot/`

escapewindow:
fixes #342

### DIFF
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -70,6 +70,7 @@ from scriptworker.utils import (
     read_from_file,
     remove_empty_keys,
     rm,
+    write_to_file,
 )
 from taskcluster.exceptions import TaskclusterFailure
 from taskcluster.aio import Queue
@@ -2225,3 +2226,80 @@ or in the CREDS_FILES http://bit.ly/2fVMu0A""")
             rm(tmp)
         else:
             log.info("Artifacts are in {}".format(tmp))
+
+
+# create_test_workdir {{{1
+async def _async_create_test_workdir(task_id, path, queue=None):
+    async with aiohttp.ClientSession() as session:
+        context = Context()
+        context.session = session
+        context.config = dict(deepcopy(DEFAULT_CONFIG))
+        context.credentials = read_worker_creds()
+        context.queue = queue or context.queue or Queue(
+            session=session,
+            options={
+                'rootUrl': os.environ.get('TASKCLUSTER_ROOT_URL', 'https://taskcluster.net'),
+            },
+        )
+        context.task = await context.queue.task(task_id)
+        work_dir = os.path.abspath(path)
+        context.config.update({
+            'work_dir': work_dir,
+            'artifact_dir': os.path.join(work_dir, 'artifacts'),
+            'task_log_dir': os.path.join(work_dir, 'artifacts', 'public', 'logs'),
+            'verify_cot_signature': False,
+        })
+        context.config = apply_product_config(context.config)
+        write_to_file(os.path.join(work_dir, "task.json"), context.task, file_type='json')
+        # we could add chain-of-trust.json and verify sha
+        for ua in context.task["payload"]["upstreamArtifacts"]:
+            task_id = ua["taskId"]
+            for path in ua["paths"]:
+                parent_dir = os.path.join(work_dir, "cot", task_id)
+                url = get_artifact_url(context, task_id, path)
+                loggable_url = get_loggable_url(url)
+                log.info("Downloading artifact:\n{}".format(loggable_url))
+                await download_artifacts(
+                    context, [url], parent_dir=parent_dir,
+                    valid_artifact_task_ids=[task_id]
+                )
+
+
+def create_test_workdir(args=None, event_loop=None):
+    """Create a test workdir, for manual testing purposes.
+
+    Args:
+        args (list, optional): the commandline args to parse.  If None, use
+            ``sys.argv[1:]`` .  Defaults to None.
+
+        event_loop (asyncio.events.AbstractEventLoop): the event loop to use.
+            If ``None``, use ``asyncio.get_event_loop()``. Defaults to ``None``.
+
+    """
+    args = args or sys.argv[1:]
+    parser = argparse.ArgumentParser(
+        description="""Populate a test `work_dir`.
+
+Given a scriptworker task's `task_id`, get its task definition, write it to
+`./work/task.json`, then download its `upstreamArtifacts` and put them in
+`./work/cot/TASK_ID/PATH`.
+
+This is helpful in manually testing a *script run.""")
+    parser.add_argument('--path', help='relative path to the work_dir', default='work')
+    parser.add_argument('--overwrite', help='overwrite an existing work_dir',
+                        action='store_true', default=False)
+    parser.add_argument('task_id', help='the task id to test')
+    opts = parser.parse_args(args)
+
+    log = logging.getLogger('scriptworker')
+    log.setLevel(logging.DEBUG)
+    logging.basicConfig()
+    if os.path.exists(opts.path):
+        if not opts.overwrite:
+            log.critical("Cowardly refusing to delete %s!", opts.path)
+            sys.exit(1)
+        rm(opts.path)
+    makedirs(opts.path)
+    event_loop = event_loop or asyncio.get_event_loop()
+    event_loop.run_until_complete(_async_create_test_workdir(opts.task_id, opts.path))
+    log.info("Done.")
--- a/scriptworker/cot/verify.py
+++ b/scriptworker/cot/verify.py
@@ -2252,9 +2252,9 @@ async def _async_create_test_workdir(task_id, path, queue=None):
         context.config = apply_product_config(context.config)
         write_to_file(os.path.join(work_dir, "task.json"), context.task, file_type='json')
         # we could add chain-of-trust.json and verify sha
-        for ua in context.task["payload"]["upstreamArtifacts"]:
-            task_id = ua["taskId"]
-            for path in ua["paths"]:
+        for upstream_artifact in context.task["payload"]["upstreamArtifacts"]:
+            task_id = upstream_artifact["taskId"]
+            for path in upstream_artifact["paths"]:
                 parent_dir = os.path.join(work_dir, "cot", task_id)
                 url = get_artifact_url(context, task_id, path)
                 loggable_url = get_loggable_url(url)
@@ -2287,7 +2287,7 @@ Given a scriptworker task's `task_id`, get its task definition, write it to
 This is helpful in manually testing a *script run.""")
     parser.add_argument('--path', help='relative path to the work_dir', default='work')
     parser.add_argument('--overwrite', help='overwrite an existing work_dir',
-                        action='store_true', default=False)
+                        action='store_true')
     parser.add_argument('task_id', help='the task id to test')
     opts = parser.parse_args(args)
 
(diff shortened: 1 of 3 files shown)


escapewindow:
@mitchhentges do you feel up to reviewing this? Non-urgent.

scriptworker/cot/verify.py
```diff
@@ -2225,3 +2226,80 @@ def verify_cot_cmdln(args=None, event_loop=None):
             rm(tmp)
         else:
             log.info("Artifacts are in {}".format(tmp))
+
+
+# create_test_workdir {{{1
```
mitchhentges: What does the `{{{1` represent? Does it format the pydoc? :thinking:

scriptworker/cot/verify.py
```diff
@@ -2225,3 +2226,80 @@ def verify_cot_cmdln(args=None, event_loop=None):
...
+        context.config.update({
+            'work_dir': work_dir,
+            'artifact_dir': os.path.join(work_dir, 'artifacts'),
+            'task_log_dir': os.path.join(work_dir, 'artifacts', 'public', 'logs'),
+            'verify_cot_signature': False,
+        })
+        context.config = apply_product_config(context.config)
+        write_to_file(os.path.join(work_dir, "task.json"), context.task, file_type='json')
+        # we could add chain-of-trust.json and verify sha
+        for ua in context.task["payload"]["upstreamArtifacts"]:
```
mitchhentges: Nit: can we rename `ua` so it's more clear what it represents? Perhaps `artifacts`?

scriptworker/cot/verify.py
```diff
@@ -2225,3 +2226,80 @@ def verify_cot_cmdln(args=None, event_loop=None):
...
+        description="""Populate a test `work_dir`.
+
+Given a scriptworker task's `task_id`, get its task definition, write it to
+`./work/task.json`, then download its `upstreamArtifacts` and put them in
+`./work/cot/TASK_ID/PATH`.
+
+This is helpful in manually testing a *script run.""")
+    parser.add_argument('--path', help='relative path to the work_dir', default='work')
+    parser.add_argument('--overwrite', help='overwrite an existing work_dir',
+                        action='store_true', default=False)
```
mitchhentges: I think that `store_true` values are `False` by default, by default :)

scriptworker/cot/verify.py
```diff
@@ -2225,3 +2226,80 @@ def verify_cot_cmdln(args=None, event_loop=None):
...
+    parser.add_argument('task_id', help='the task id to test')
+    opts = parser.parse_args(args)
+
+    log = logging.getLogger('scriptworker')
+    log.setLevel(logging.DEBUG)
+    logging.basicConfig()
+    if os.path.exists(opts.path):
+        if not opts.overwrite:
+            log.critical("Cowardly refusing to delete %s!", opts.path)
```
mitchhentges: Minor "no fun allowed" nit: this is ambiguous between "scriptworker is being cowardly and not deleting this, haha" and "the library called 'Cowardly' has refused to delete this for an unknown reason".
Threw me off for 0.1s, but just food for thought

scriptworker/test/test_cot_verify.py
```diff
@@ -2367,3 +2367,64 @@ def cot(*args, **kwargs):
...
+        args.append("--overwrite")
+
+    async def fake_create(*args):
+        assert args == ("taskId", work_dir)
+
+    mocker.patch.object(cotverify, "_async_create_test_workdir", new=fake_create)
+    if exists:
+        makedirs(work_dir)
+    if raises:
+        with pytest.raises(SystemExit):
```
mitchhentges: FWIW, To make it easier to read the flow of this test, you could break it into four simpler tests :thinking:

setup.py
```diff
@@ -78,6 +78,7 @@ def run_tests(self):
         "console_scripts": [
             "scriptworker = scriptworker.worker:main",
             "verify_cot = scriptworker.cot.verify:verify_cot_cmdln",
+            "create_test_workdir = scriptworker.cot.verify:create_test_workdir",
```
mitchhentges: :ok_hand:

scriptworker/cot/verify.py
escapewindow: It's a vim folding marker. Not everyone uses vim, but it's relatively harmless comment for other editors. This lets me group functions to fold, making the 2305 line file easier to navigate.

scriptworker/cot/verify.py
escapewindow: Done. I went ahead and did the non-lazy thing and named it `upstream_artifact`.

scriptworker/cot/verify.py
escapewindow: Makes sense. Removed the `default`.

scriptworker/cot/verify.py
escapewindow: Sure. Going with `Scriptworker will not delete %s without --overwrite!`

scriptworker/test/test_cot_verify.py
escapewindow: Sure. That would make it easier to read each test, by requiring more effort to create each test. I tend to rely on parameters to help test multiple code branches. The theory is if we make testing code branches easier to write, people will write more tests.

mitchhentges: I hear you, though of course sometimes it's not just the amount of code to write that contributes to the friction of writing tests. If I attempt to capture additional test cases, but existing parameter sets fail in unexpected ways (or if it's difficult to imagine how new tests would fit into the existing generic test structure), the costs might outweigh the theoretical benefit of fewer lines of test code needing to be written.

escapewindow: We have the pattern of the `raises` parameter all over the scriptworker tests, so changing that behavior here might not be a huge win overall. I could split this into two functions, one with a `makedirs` and one without. My first instinct would be to create a helper function that does everything `test_create_test_workdir` currently does, and create stub test functions that call the helper functions, but without parameterizing, but I don't think that helps readability at all.

Sounds like you consider this test change a blocker?

mitchhentges: Hmm, interesting :thinking: 
It is true that we have a convention of the `raises` parameter elsewhere, too. I'm definitely not worried about this, it _isn't_ a blocker. Feel free to merge! I think this is an interesting discussion either way :)

I agree that just moving the parameters into separate tests but still having a generic, separate `test_create_test_workdir` function wouldn't improve readability.
I'd probably lean on duplication and create four very straightforward tests:

```
async def fake_create_test_workdir(*args):
    assert args == ("taskId", work_dir)


def test_create_test_workdir(mocker, event_loop, tmpdir):
    work_dir = os.path.join(tmpdir, "work")
    mocker.patch.object(cotverify, "_async_create_test_workdir", new=fake_create_test_workdir)
    makedirs(work_dir)
    cotverify.create_test_workdir(args=["--path", work_dir, "taskId", "--overwrite"], event_loop=event_loop)


def test_create_test_workdir(mocker, event_loop, tmpdir):
    work_dir = os.path.join(tmpdir, "work")
    mocker.patch.object(cotverify, "_async_create_test_workdir", new=fake_create_test_workdir)
    cotverify.create_test_workdir(args=["--path", work_dir, "taskId"], event_loop=event_loop)

def test_create_test_workdir(mocker, event_loop, tmpdir):
    work_dir = os.path.join(tmpdir, "work")
    mocker.patch.object(cotverify, "_async_create_test_workdir", new=fake_create_test_workdir)
    cotverify.create_test_workdir(args=["--path", work_dir, "taskId", "--overwrite"], event_loop=event_loop)


def test_create_test_workdir(mocker, event_loop, tmpdir):
    work_dir = os.path.join(tmpdir, "work")
    mocker.patch.object(cotverify, "_async_create_test_workdir", new=fake_create_test_workdir)
    makedirs(work_dir)
    with pytest.raises(SystemExit):
        cotverify.create_test_workdir(args=["--path", work_dir, "taskId"], event_loop=event_loop)
```

escapewindow: I agree each individual test is easier to read in your example. We'd need to avoid duplicate test names, so we could add something like `_no_existing_work_dir_no_overwrite` to the test name.

Later on, if we refactored `create_test_workdir` significantly, it might mean we need to edit 4 test functions rather than 1. It might not be immediately clear which tests should stick around (which tests are still useful) and which are essentially duplicates of other tests. Also, let's say we change the behavior of the code that affects a subset of the tests: we'd need to figure out which test(s) to change. This could be 1, 4, or if we take this to more of an extreme, 10 or 100. Also, if we have enough of these functions with only minor changes between them, it may end up being a "where's waldo" or "can you find the 10 differences between these images" type game figuring out what's different between them, especially if the functions grow to more than 3-5 lines. Alternately if we have descriptive enough test names, the function names may approach the code length of the simpler tests ;-)

I don't think either approach is perfect for all cases. I still prefer the parameterization approach for now, so I'll merge. Thanks!

mitchhentges: > Later on, if we refactored create_test_workdir significantly, it might mean we need to edit 4 test functions rather than 1. It might not be immediately clear which tests should stick around (which tests are still useful) and which are essentially duplicates of other tests.

It's true that more test functions would need to be edited, though fortunately, it's work that's hard to do incorrectly, since the tests won't (shouldn't :stuck_out_tongue:) pass until they've all been updated.

However, and more importantly, is knowing which tests become obsolete. With a single generic parameterized test, wouldn't it be _more_ challenging to realize which parameter sets are obsolete?

> This could be 1, 4, or if we take this to more of an extreme, 10 or 100.

Of course, as the number of tests scale, when you find behaviour duplication within your tests, you should absolutely apply re-use accordingly. IMHO, the abstraction of re-use applied in this particular example is less valuable.

-----

Reading over your comment again, I think we're pretty aligned, but we draw the line of "at what point should code be re-used" at different parts of the scale. :+1:

escapewindow: > Reading over your comment again, I think we're pretty aligned, but we draw the line of "at what point should code be re-used" at different parts of the scale. 👍

I agree with this a lot. And I could be convinced that for super simple tests like this one, a different approach than parameterization might be better. We use parameterization across most (all?) the scriptworker tests, though, and I don't love having to rewrite simple tests to a parametrized test once the function passes a certain level of complexity. I think consistently using parameters across scriptworker tests is worth the readability costs of parameterized simpler tests.

mitchhentges: > I don't love having to rewrite [a set of multiple] simple tests to a [single] parameterized test once the function passes a certain level of complexity. I think consistently using parameters across scriptworker tests is worth the readability costs of parameterized simpler tests.

I agree that as we add more tests, our desire to add re-use between tests will definitely grow. I'm not sure that that re-use will always manifest by condensing multiple tests into a single, well-parameterized test, though I'm sure that it will still be sometimes applicable.

One other interesting observation is how the re-use is applied - I feel that it can be more straightforward to have "bottom-up" re-use in tests (share common grouped code like `fake_create_test_workdir`) rather than "top-down" (share a full test workflow, with different branches enabled/disabled with parameters)

PR #352, opened 2019-06-22
add TASK_ID to task env.

### LINKED ISSUE #350: add TASK_ID to os.environ when running task script
We may want to allow the script to reference TASK_ID, like the other worker implementations.

escapewindow:
fixes #350

### DIFF
--- a/scriptworker/task.py
+++ b/scriptworker/task.py
@@ -530,7 +530,7 @@ def prepare_to_run_task(context, claim_task):
     """
     current_task_info = {}
     context.claim_task = claim_task
-    current_task_info['taskId'] = get_task_id(claim_task)
+    current_task_info['taskId'] = context.task_id
     current_task_info['runId'] = get_run_id(claim_task)
     log.info("Going to run taskId {taskId} runId {runId}!".format(
         **current_task_info
@@ -556,12 +556,15 @@ async def run_task(context, to_cancellable_process):
         int: exit code
 
     """
+    env = deepcopy(os.environ)
+    env['TASK_ID'] = context.task_id or 'None'
     kwargs = {  # pragma: no branch
         'stdout': PIPE,
         'stderr': PIPE,
         'stdin': None,
         'close_fds': True,
         'preexec_fn': lambda: os.setsid(),
+        'env': env,
     }
 
     subprocess = await asyncio.create_subprocess_exec(*context.config['task_script'], **kwargs)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [23.1.0] - 2019-06-26
+### Added
+- Added `context.task_id`
+
+### Changed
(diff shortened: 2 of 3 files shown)


scriptworker/task.py
```diff
@@ -556,12 +556,15 @@ def prepare_to_run_task(context, claim_task):
         int: exit code
 
     """
+    env = deepcopy(os.environ)
+    env['TASK_ID'] = context.task_id or 'None'
```
mitchhentges: Is it possible to not have a `task_id`?

mitchhentges: As a more general question, is the environment the only way that we can communicate between `scriptworker` and the task script?

escapewindow: If we haven't claimed a task. It happens in unit tests, not so much in real life.

escapewindow: `env['TASK_ID']` is the standard way of communicating the taskId from the worker to the task across worker implementations. Scriptworkers *also* have the ability to read `task.json` and other metadata from disk, since scriptworker writes those artifacts in a predictable place the script can read. I think implementing the same convention that generic- and docker-worker have with the env var makes sense here.

mitchhentges: As a nit, should we assume that `context.task_id` is set (`env['TASK_ID'] = context.task_id`), and ensure that unit tests that run this logic are mocked accordingly?
Otherwise, it looks like it could be normal behaviour in real life for a task_id not to be set? :thinking:

escapewindow: Could do. Could also move the `or 'None'` logic into the `context.task_id` property. The main issue is that if we don't do this, we get `env['TASK_ID'] = None`, and environment values need to be strings. And we're *currently* getting `"None"` baked into our notarization id strings since we're templatizing it with `os.environ.get("TASK_ID")`.

I wasn't sure how much of a win either of those would be, and I know that people don't love having even more in-test setup and mocking to do before they can run a test, so this or the `context.task_id` returning the string `"None"` seemed like the best amount of improvement while not affecting ease of testing.

escapewindow: I'm going to merge this so we get TASK_ID in the notarization id, but if you feel strongly about this, I can address in a followup.

mitchhentges: > The main issue is that if we don't do this, we get env['TASK_ID'] = None, and environment values need to be strings

I feel like this is pushing an unnecessary extra case down into scripts, since they'll need to understand what it means if `env['TASK_ID']` is set to the magic string "None". If this case should be impossible, perhaps we should validate against it?

> people don't love having even more in-test setup and mocking to do before they can run a test

I agree that test ergonomics are important, but IIUC this solution is weakening the restrictions on state/variables to make tests easier. This will add complexity to the runtime, since there's more potential states to manage.

escapewindow: I only know of one scriptworker TASK_ID use case, and that's for use in notarization. In that case, a string of `None` would be no change, and adding a real taskId would be an improvement. If we need to guarantee that TASK_ID is set properly and we foresee issues here, I agree we should test around it.

Noting that this is a case where you're advocating more tests, and I'm pushing back a bit since there are no current real world needs: it's nice to switch roles sometimes :)

PR #356, opened 2019-06-27
Reports unexpected exceptions to Taskcluster, doesn't end process

### LINKED ISSUE #353: Logs aren't being uploaded upon exception being raised
For a `beetmover` job that I'm running that uses a new payload structure, the job is failing due to an exception being thrown - this is expected, because I haven't pinned the worker yet.
However, when the exception is raised, the worker crashes, and doesn't inform Taskcluster about the exception. So, [in the Taskcluster UI, the reason for the job not being completed is: "claim expired"](https://tools.taskcluster.net/groups/EKvNYTimRhGubuWMGQjWHg/tasks/JQNYYUTqQPWmGcQNCOGTGw/details).
Additionally, no logs are uploaded to Taskcluster.

SSH-ing into the worker, I can see the package versions as:
* `beetmoverscript`: `8.4.7`
* `scriptworker`: `23.0.10`

mitchhentges:
Fixes #353.
Rather than responding to unexpected errors within `do_run_task()` by bubbling it to the top and finishing the `scriptworker` process, this PR reports the unexpected error to Taskcluster, and lets the worker continue running tasks.

Since the process would've been restarted anyways, the net effect of this change is that unexpected-error logs are still given to Taskcluster.


escapewindow:
I don't think you mean #343 :)

scriptworker/test/test_worker.py
```diff
@@ -197,24 +198,31 @@ def _mocker_run_tasks_helper(mocker, exc, func_to_raise):
...
 
+@pytest.mark.asyncio
+async def test_unexpected_exception_catch_and_tell_taskcluster(context, successful_queue, mocker):
+    def fail():
+        raise KeyError()
+
+    async def run(coroutine):
+        await coroutine
+
+    mocker.patch.object(verify, "verify_chain_of_trust", new=fail)
```
escapewindow: You can actually `mocker.patch.object(worker, "verify_chain_of_trust", new=fail)` since `worker` has imported that function. Essentially the same thing, but you don't have to import `cot.verify` into the test module.

scriptworker/worker.py
```diff
@@ -62,7 +62,7 @@
         log.error("Hit ScriptWorkerException: {}".format(e))
     except Exception as e:
         log.exception("SCRIPTWORKER_UNEXPECTED_EXCEPTION task {}".format(e))
-        raise
+        status = STATUSES['internal-error']
```
escapewindow: It's not really an internal error we're talking about here. It's a malformed upstream artifact.
Is there a reason we're fixing this here, rather than adding `TypeError` to https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/cot/verify.py#L2146 , which would raise a CoTError?

From my viewpoint:

- we set up a graph that never could have worked, due to task-graph.json being malformed
- we have a long term solution to creating task-graph.json consistently: using taskgraph everywhere
- this may be a good change, so we can take this as well. However, we also have the information to make a more surgical change that would give us a better exit code for #353 specifically.

mitchhentges: Ah, yes, I'll gladly add `TypeError` to `verify_chain_of_trust` - providing a more-relevant exit code for taskcluster is a big :+1: from me :smile:, good catch! I assumed that a schema would help here, but seeing as we have a top-level `except` block already, I'll hook onto that!

mitchhentges:
No worries, I can take care of it!

PR #357, opened 2019-06-27
Add esr to pushsnap restricted scopes.

tomprince:


tomprince:
We already publish snap packages on esr60.

PR #361, opened 2019-07-02
Log workerId

rail:
In GCP we use slugid to generate workerId and it's harder to connect the random k8s replica name to this workerId.


rail:
> Would logging the hostname help at all? (For shutting down in k8s if it’s acting up, or ?)

Yeah, it's a good idea, let's log it too. Will push a commit in a bit.

PR #368, opened 2019-07-18
Enable stricter checks in `trace_back_to_tree`.

tomprince:


scriptworker/constants.py
```diff
@@ -202,30 +194,26 @@
                 "path_regexes": (
                     r"^(?P<path>/comm-central)(/|$)",
                     r"^(?P<path>/releases/comm-(beta|esr\d+))(/|$)",
-                    r"^(?P<path>/(try-comm-central))(/|$)",
+                ),
+                "unstrusted_path_regexes": (
```
escapewindow: sp. why do we have both path_regexes and untrusted_path_regexes?

scriptworker/task.py
```diff
@@ -293,7 +293,7 @@ def get_and_check_project(valid_vcs_rules, source_url):
         str: the project.
 
     """
-    project_path = match_url_regex(valid_vcs_rules, source_url, match_url_path_callback)
+    project_path = match_url_regex(valid_vcs_rules, source_url, match_url_path_callback, trusted=False)
```
escapewindow: I don't see us doing anything with `trusted`.

scriptworker/constants.py
tomprince: They aren't identical. For `comm`, there is `try-comm-central

scriptworker/task.py
tomprince: I could instead introduce a new config that is used here, instead of repurposing `valid_vcs_rules`. It was slightly more convenient to develop it this way (particularly before I realized this function wasn't needed for mobile

scriptworker/constants.py
escapewindow: What benefit do we have of returning `try-comm-central`, as opposed to `None`?

tomprince: Even for unrestricted scopes, we still verify tasks, so this is needed for the dep-signing worker, and all the staging workers, to be able to generate the correct json-e context.

escapewindow: Ok. Still, as noted in the first comment, `unstrusted` [sp] will make this not work. If we really want to open this up to all repos, we can probably do `r"^(?P<path>.*)/*$"` or the like.

scriptworker/task.py
tomprince: Done.

scriptworker/constants.py
tomprince: I ended up doing something like this. We do need to handle both
- `https://hg.mozilla.org/mozilla-central/file/<sha>/taskcluster/ci/build` -> `mozilla-central`
- `https://hg.mozilla.org/releases/mozilla-beta/file/<sha>/taskcluster/ci/build` -> `mozilla-beta`
so we do need at least a partial enumeration of things.

scriptworker/constants.py
```diff
@@ -202,34 +186,40 @@
...
+    # This contains a list of known projects to help extract it.
+    "valid_vcs_rules": (frozendict({
+        "schemes": ("https", "ssh", ),
+        "netlocs": ("hg.mozilla.org", ),
+        "path_regexes": (
+            r"^/(?P<project>mozilla-(central|unified))(/|$)",
+            r"^/(?P<project>comm-central)(/|$)",
+            r"^/integration/(?P<project>([A-Za-z0-9-]+))(/|$)",
+            r"^/releases/(?P<project>(comm|mozilla)-([A-Za-z0-9-]+))(/|$)",
+            r"^/projects/(?P<project>([A-Za-z0-9-]+))(/|$)",
```
escapewindow: Let's add `/ci/` in here if we want to trace tasks back to ci-* properly.

escapewindow:
Thanks! Looks good. Let's make sure Johan doesn't have any additional concerns.

scriptworker/constants.py
```diff
@@ -202,34 +186,41 @@
...
                     r"^(?P<path>/comm-central)(/|$)",
                     r"^(?P<path>/releases/comm-(beta|esr\d+))(/|$)",
-                    r"^(?P<path>/(try-comm-central))(/|$)",
                 ),
             }),),
             'mobile': (frozendict({
                 "schemes": ("https", "ssh", ),
                 "netlocs": ("github.com", ),
                 "path_regexes": tuple([
-                    # XXX We allow Github forks to run staging releases. Please make sure to
```
JohanLorenzo: Potential blocking issue: I'm not sure to follow, can we still run staging releases on our personal repos after these changes? If not, what would be the new way to do it?

scriptworker/test/test_cot_verify.py
```diff
@@ -1376,10 +1376,6 @@ def test_verify_link_in_task_graph_exception(chain, decision_link, build_link, i
         },
         'now': '2018-01-01T12:00:00.000Z',
         'ownTaskId': 'decision_task_id',
-        'repository': {
```
JohanLorenzo: :+1: Stripping `jsone_context['repository']` for Github events makes sense to me!

PR #369, opened 2019-07-19
Add some end-to-end mobile tests.

tomprince:


scriptworker/test/test_production.py
```diff
@@ -0,0 +1,187 @@
...
+from scriptworker.context import Context
+from scriptworker.cot.verify import ChainOfTrust, verify_chain_of_trust
+import scriptworker.log as swlog
+import scriptworker.utils as utils
+from taskcluster.aio import Index, Queue
+
+log = logging.getLogger(__name__)
+
+# constants helpers and fixtures {{{1
+SKIP_REASON = "NO_TESTS_OVER_WIRE: skipping integration test"
```
escapewindow: `s,integration,production` ?

scriptworker/test/test_production.py
```diff
@@ -0,0 +1,187 @@
...
+    "task_type": "signing",
+    "cot_product": "firefox",
+}, {
+    "name": "mozilla-esr68 win64 en-US repackage signing",
+    "index": "gecko.v2.mozilla-esr60.nightly.latest.firefox.win64-nightly-repackage-signing",
+    "task_type": "signing",
+    "cot_product": "firefox",
+}, {
+    "name": "fenix nightly",
+    "index": "project.mobile.fenix.v2.nightly.latest",
```
escapewindow: Awesome!
If we're able to verify a later task, e.g. a signing task or pushapk task, even better. That may be dependent on what's indexed.

scriptworker/test/test_production.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python
+# coding=utf-8
+"""Scriptworker integration tests.
```
escapewindow: production?

scriptworker/test/test_production.py
```diff
@@ -0,0 +1,187 @@
...
+        assert task_id, "{}: Can't get task_id from index {}!".format(
+            branch_context['name'], branch_context['index']
+        )
+        if branch_context.get('task_label_to_task_type'):
+            task_info = await get_completed_task_info_from_labels(
+                task_id, branch_context['task_label_to_task_type']
+            )
+            for task_id, task_type in task_info.items():
+                name = "{} {}".format(branch_context['name'], task_type)
+                await verify_cot(name, task_id, task_type)
```
escapewindow: Do these work for you? I get

`E               TypeError: verify_cot() missing 1 required positional argument: 'cot_product'`

Seems like a simple fix.

scriptworker/test/test_production.py
```diff
@@ -0,0 +1,187 @@
...
+        else:
+            await verify_cot(branch_context['name'], task_id, branch_context['task_type'])
```
escapewindow: Here too.

.travis.yml
```diff
@@ -5,12 +5,12 @@ matrix:
...
     - python: 3.6
       dist: trusty
-      env: TOXENV=py36 NO_TESTS_OVER_WIRE=1
+      env: TOXENV=py36 NO_CREDENTIALS_TESTS=1
       sudo: false
     - python: 3.7
       # from https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-401756442
       dist: xenial
-      env: TOXENV=py37 NO_TESTS_OVER_WIRE=1
+      env: TOXENV=py37 NO_CREDENTIALS_TESTS=1
```
escapewindow: We can see how these go, and tweak if this is flaky.

I was envisioning a py36 unittest run, a py37 unittest run, and an integration/network test run with live integration secrets, esp. once we switch our CI tests in this repo to taskcluster. That should help give us more information, while isolating the network tests' variability from the local tests.

scriptworker/github.py
```diff
@@ -119,13 +117,12 @@ def get_tag_hash(self, tag_name):
         url = '/'.join([repo.rstrip('/'), 'branch_commits', revision])
         from scriptworker.task import get_decision_task_id
         cache_key = '{}-{}'.format(get_decision_task_id(context.task), url)
-        async with _branch_commits_cache_lock:
```
escapewindow: I'd love to get @eoger and @JohanLorenzo 's take on this change, since they put it in originally.

scriptworker/test/test_integration.py
```diff
@@ -23,18 +22,19 @@
...
 from . import integration_create_task_payload
 
 log = logging.getLogger(__name__)
 
 # constants helpers and fixtures {{{1
-SKIP_REASON = "NO_TESTS_OVER_WIRE: skipping integration test"
+pytestmark = [
+    pytest.mark.skipif(os.environ.get("NO_TESTS_OVER_WIRE"), reason="NO_TESTS_OVER_WIRE: skipping integration test"),
+    pytest.mark.skipif(os.environ.get("NO_CREDENTIALS_TESTS"), reason="NO_CREDENTIALS_TESTS: skipping integration test")
+]
```
escapewindow: Nice.

.travis.yml
escapewindow: Hm, and it broke on this PR :) Maybe we should either enable in a 3rd non-blocking test, or keep them disabled in Travis til we have a better solution?

tomprince: I've done the former.

scriptworker/test/test_production.py
JohanLorenzo: :+1: Signing tasks are the ones indexed at `project.mobile.fenix.v2.nightly.latest` (for instance https://tools.taskcluster.net/groups/ahv5e-kfQUCtF-2J87nUHw/tasks/b3T1dbdJQMGqnaVpBBAS1Q/details)

IIRC, we haven't indexed pushapk tasks. Doing so shouldn't be a problem though.

scriptworker/github.py
tomprince: Running the test in `scriptworker/test/test_production.py`  I was getting consistent errors (both locally and in CI) without removing this.

JohanLorenzo: Okay! Would you have CI logs to share, @tomprince? I've thought about it a bit more and I believe I remember what happened: without the lock, 4 requests were fired at once which was unnecessary.

JohanLorenzo: Okay! Would you have CI logs to share, @tomprince? I've thought about it a bit more and I believe I remember what happened: without the lock, 4 requests were fired at once which was unnecessary.

tomprince: I can't find the logs in travis, but if you checkout f71081e805227aa3a66b346d5cd027e6fff20014 and run `py.test -k 'test_verify_production_cot[branch_context10]'` you should see the error:
```
scriptworker/task.py:462: in is_try_or_pull_request
    return await is_pull_request(context, task)
scriptworker/task.py:443: in is_pull_request
    conditions.append(not await github_repository.has_commit_landed_on_repository(context, revision))
scriptworker/github.py:122: in has_commit_landed_on_repository
    async with _branch_commits_cache_lock:
/usr/lib64/python3.6/asyncio/locks.py:79: in __aenter__
    yield from self.acquire()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asyncio.locks.Lock object at 0x7fd6a7bde128 [unlocked]>

    @coroutine
    def acquire(self):
        """Acquire a lock.
    
        This method blocks until the lock is unlocked, then sets it to
        locked and returns True.
        """
        if not self._locked and all(w.cancelled() for w in self._waiters):
            self._locked = True
            return True
    
        fut = self._loop.create_future()
        self._waiters.append(fut)
    
        # Finally block should be called before the CancelledError
        # handling as we don't want CancelledError to call
        # _wake_up_first() and attempt to wake up itself.
        try:
            try:
>               yield from fut
E               RuntimeError: Task <Task pending coro=<is_try_or_pull_request() running at /Depot/Mozilla/scriptworker/scriptworker/task.py:462> cb=[_wait.<locals>._on_completion() at /usr/lib64/python3.6/asyncio/tasks.py:380]> got Future <Future pending> attached to a different loop
```

JohanLorenzo: Status update: I think we need to bring this lock back. See https://bugzilla.mozilla.org/show_bug.cgi?id=1596439 for more details

PR #370, opened 2019-07-20
Improve production CoT checks.

tomprince:


scriptworker/test/test_production.py
```diff
@@ -118,6 +118,7 @@ def build_config(override, basedir):
     "index": "project.mobile.fenix.v2.branch.master.latest.raptor.aarch64",
     "task_type": "build",
     "cot_product": "mobile",
+    "check_task": False,
```
escapewindow: Is there a reason we're skipping these?

tomprince: These are dep scriptworkers, so the sig checks would fail.

PR #372, opened 2019-07-23
Bug 1563328 - Add CoT support dedicated workers for application-services

MihaiTabara:


MihaiTabara:
Staging release with these changes pinned to signing and beetmover workers are green in [here](https://tools.taskcluster.net/groups/aBIreYB0Q2i6-WkDGG88eQ). Merging this.

PR #373, opened 2019-07-24
Bug 1563328 - cleanup old workerType in application-services

MihaiTabara:


MihaiTabara:
Will merge and let it ride with whichever scriptworker release comes up next. This ain't burning and it doesn't need a release of itself.

PR #383, opened 2019-08-12
Use `verify_scriptworker_task` explicitly

rail:
The new treescript workers use `cot_job_type` set to `tree` instead of
generic `scriptworker` we use in puppet. This require a new entry in
`get_valid_task_types`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [24.0.2] - 2019-08-12
+### Changed
+- Use `verify_scriptworker_task` for workers indirectly using it
+
 ## [24.0.1] - 2019-08-08
 ### Added
 - Added new scriptworker names to CoT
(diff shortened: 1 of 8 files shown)


escapewindow:
+1, we should move to use `scriptworker` everywhere.

rail:
I'll change the pr to simplify things

version.json
```diff
@@ -2,7 +2,7 @@
     "version": [
         24,
         0,
-        1
+        2
     ],
-    "version_string": "24.0.1"
+    "version_string": "24.0.2"
```
escapewindow: We're removing a bunch of public functions, so this should probably be `25.0.0`.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [24.0.2] - 2019-08-12
```
escapewindow: 25.0.0; we should mention the functions removed.

rail: right, will fix now

version.json
rail: rgr

rail:
Come on, coveralls!

PR #386, opened 2019-08-16
GitHub taskgraph cron

tomprince:


escapewindow:
We should block on a green test_production.py run (as well as the other test fixes, but github will allow us to merge without test_production.py being green. we should verify it's green before merging). Still wrapping my head around this change.

escapewindow:
It looks like `test_cot_verify` may be hitting the live projects url. We should probably pre-populate the context.projects so we don't hit the network and aren't dependent on the state of an external file for our unittests.

scriptworker/test/test_task.py
```diff
@@ -196,25 +196,26 @@ def test_get_worker_type(task, result):
 @pytest.mark.parametrize("source_url, expected, raises, context_type", ((
     "https://hg.mozilla.org/mozilla-central", "mozilla-central", False, "firefox"
 ), (
-    "ssh://hg.mozilla.org/projects/foo", "foo", False, "firefox"
+    "https://hg.mozilla.org/projects/foo", "foo", True, "firefox"
 ), (
-    "ssh://hg.mozilla.org/releases/mozilla-esr60", "mozilla-esr60", False, "firefox"
+    "https://hg.mozilla.org/releases/mozilla-esr60", "mozilla-esr60", False, "firefox"
```
escapewindow: It looks like we no longer support `ssh` as an optional scheme, which is likely fine in the majority of cases.
If we ever go a shadow-central route (no public https access), is it plausible to put an `ssh://` url in `projects.yml` ?

tomprince:
> It looks like `test_cot_verify` may be hitting the live projects url. We should probably pre-populate the context.projects so we don't hit the network and aren't dependent on the state of an external file for our unittests.

Yeah. That would be a good follow-up.

scriptworker/test/test_task.py
tomprince: In principle, it should work, but I bet there are assumption all over the place that don't work with it. We do `url[8:]` in a number of places, to strip `https://` off, for example.

escapewindow: We should probably urlparse it rather than assume the scheme is a specific length.

tomprince: Yeah, although json-e (where this is often done) doesn't have that ability.

PR #389, opened 2019-08-28
reduce mohawk logging to logging.INFO

escapewindow:
To avoid logging like https://taskcluster-artifacts.net/RReanjm1TiKXVmIkEfFRAQ/0/public/logs/live_backing.log


escapewindow:
Thanks!

PR #391, opened 2019-09-05
Adds firefox-tv to CoT constants

mitchhentges:


scriptworker/constants.py
```diff
@@ -329,10 +329,14 @@
...
 
-                'project:mobile:reference-browser:releng:signing:cert:release-signing': 'reference-browser-repo',
                 'project:mobile:reference-browser:releng:googleplay:product:reference-browser': 'reference-browser-repo',
+                'project:mobile:reference-browser:releng:signing:cert:release-signing': 'reference-browser-repo',
+
+                'project:mobile:firefox-tv:releng:googleplay:product:firefox-tv': 'firefox-tv-repo',
+                'project:mobile:firefox-tv:releng:signing:cert:production-signing': 'firefox-tv-repo',
             }),
             'application-services': frozendict({
+                'project:releng:signing:cert:release-signing': 'application-services-repo',
```
escapewindow: oops.

scriptworker/constants.py
```diff
@@ -329,10 +329,14 @@
                 'project:mobile:focus:googleplay:product:focus': 'focus-repo',
                 'project:mobile:focus:releng:signing:cert:release-signing': 'focus-repo',
 
-                'project:mobile:reference-browser:releng:signing:cert:release-signing': 'reference-browser-repo',
                 'project:mobile:reference-browser:releng:googleplay:product:reference-browser': 'reference-browser-repo',
+                'project:mobile:reference-browser:releng:signing:cert:release-signing': 'reference-browser-repo',
+
+                'project:mobile:firefox-tv:releng:googleplay:product:firefox-tv': 'firefox-tv-repo',
+                'project:mobile:firefox-tv:releng:signing:cert:production-signing': 'firefox-tv-repo',
```
escapewindow: Any reason we're moving away from `release-signing` to `production-signing` here? Consistency would be good from a scriptworker viewpoint. Not sure if there are other concerns.

scriptworker/constants.py
mitchhentges: Oops, my stash-and-apply must've goof'd.
I should've checked this PR diff right after creating the PR, sorry about this!

scriptworker/constants.py
mitchhentges: We used to just have a single release key and a single dep key, but starting with `fenix`, we started to have multiple release keys (nightly, beta, production). See [this PR](https://github.com/mozilla-releng/scriptworker/pull/392) that updates CoT according to this change in expectations.

For this reason, since there might be multiple channels of `firefox-tv`, I'm using the same strategy of using the term "production" to refer to the "main" channel, and letting there be other release-like keys as well

PR #395, opened 2019-09-09
`arrow.get` no longer takes timestamp strings without a format string

### LINKED ISSUE #394: fix arrow 0.15.0
https://github.com/crsmithdev/arrow/blob/master/CHANGELOG.md#0150

specifically, `arrow.get("timestamp string")` will no longer work. These have to change to either be `arrow.get(int("timestamp string"))`, `arrow.get(float("timestamp string"))`, or `arrow.get("timestamp string", "X")`

escapewindow:
Fixes #394 .

### DIFF
--- a/scriptworker/utils.py
+++ b/scriptworker/utils.py
@@ -115,7 +115,7 @@ def datestring_to_timestamp(datestring):
         int: the corresponding timestamp.
 
     """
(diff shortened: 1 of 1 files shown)


escapewindow:
Hm, both of these are datestrings, which work as is. We might be good without this PR.

Issue #397, opened 2019-09-25
CoT version 7: drop non-hook action support


escapewindow:
The code change is fairly easy. We may want to create a new `scriptworker/test/data/cotv7` data directory and retire the `cotv2` and `cotv4` `test/data` directories, so the test changes may require a bit of work.

ahal:
I'll take a look into this to help familiarize myself with the repo.

PR #398, opened 2019-09-25
Cot cleanup

escapewindow:
While writing up the CoT technical details doc, we noticed some things we could clean up.

- remove the unused `extra_run_task_arguments` config
- clarify some path logic
- remove the obsolete `check_num_tasks` -- because of the logic in adding parent dirs per task, we're guaranteed at least one decision task in the graph

@tomprince do you feel up to reviewing this?

I filed #397 for dropping non-hook action support. I was going to bundle that here, but the test changes were large enough that it deserves its own PR.


scriptworker/artifacts.py
```diff
@@ -424,3 +429,22 @@ def get_optional_artifacts_per_task_id(upstream_artifacts):
             )
 
     return optional_artifacts_per_task_id
+
+
+def assert_is_parent(path, parent_dir, exception=ScriptWorkerTaskException):
```
JohanLorenzo: What's the use case for passing down `exception`? I don't see it used anywhere in this PR. I wonder if it's case of YAGNI.

scriptworker/test/test_artifacts.py
```diff
@@ -377,3 +378,31 @@ def test_get_single_upstream_artifact_full_path(context):
...
+@pytest.mark.parametrize("path, parent_path, raises", ((
+    "/foo/bar/baz", "/foo/bar", False
+), (
+    "/foo", "/foo/bar", True
+), (
+    "/foo/bar/..", "/foo/bar", True
+)))
+def test_assert_is_parent(path, parent_path, raises):
+    if raises:
+        with pytest.raises(ScriptWorkerTaskException):
```
JohanLorenzo: Nit: assert on `exception` if we want to support this use case.

scriptworker/context.py
```diff
@@ -82,13 +83,22 @@ def claim_task(self, claim_task):
...
             self.temp_credentials = None
             self.task = None
 
+    def verify_task(self):
+        """Run some task sanity checks on ``self.task``."""
+        for upstream_artifact in self.task.get("payload", {}).get("upstreamArtifacts", []):
+            task_id = upstream_artifact["taskId"]
+            for path in upstream_artifact["paths"]:
+                if os.path.isabs(path) or '..' in path:
+                    raise CoTError("upstreamArtifacts taskId {} has illegal path {}!".format(task_id, path))
```
JohanLorenzo: 👍

scriptworker/test/test_artifacts.py
```diff
@@ -377,3 +378,31 @@ def test_get_single_upstream_artifact_full_path(context):
...
+            swartifacts.assert_is_parent(path, parent_path)
+    else:
+        swartifacts.assert_is_parent(path, parent_path)
+
+
+def test_assert_is_parent_softlink(tmpdir):
```
JohanLorenzo: 👍  for thinking of soft links. That makes me think, should we also test hard links? https://docs.python.org/2/library/os.html#os.link

escapewindow:
> Looks great to me! Thanks for these cleanups.
Thanks!

> Do you think we should use `assert_is_parents` at the these places too?
> 
> https://github.com/mozilla-releng/scriptworker/blob/0d2053f600b5b3224db916a08c1349eb4db6f863/scriptworker/cot/verify.py#L2168

This is only for the create-test-workdir tool, which isn't used in production.

> https://github.com/mozilla-releng/scriptworker/blob/0d2053f600b5b3224db916a08c1349eb4db6f863/scriptworker/version.py#L75

This is only used to find the `version.json` file.

I don't think either of these are concerning, but I might be missing something.

scriptworker/test/test_artifacts.py
escapewindow: I don't think we can download a hardlink, while we might be able to download a softlink.

scriptworker/artifacts.py
escapewindow: Hm. I think I was angling for raising a `CoTError` if we hit this during `cot.verify`, and a `ScriptWorkerTaskException` elsewhere, but it looks like I don't use it.

PR #400, opened 2019-10-16
bug 1520281 - Support xpi signing

escapewindow:
@Callek , @tomprince , I think we've been working in the same areas of github cot verification. I'm currently working on a pinned puppet env, so I'm open to which order we land.

@JohanLorenzo , letting you know I'm changing how we're addressing email addresses on github. I noticed that the event email is my primary email, but my repo email is my secondary, and there's no way for me to verify if the task email is accurate. I think we just have to trust what comes through.


scriptworker/cot/verify.py
```diff
@@ -1217,13 +1217,19 @@ def verify_link_in_task_graph(chain, decision_link, task_link):
     ).get_commit(commit_hash)
 
     committer_login = commit_data['committer']['login']
-    committer_email = commit_data['commit']['committer']['email']
+    # committer_email = commit_data['commit']['committer']['email']
```
JohanLorenzo: Nit: let's remove these commented out lines in order to not confused future readers.

escapewindow:
> Looks good to me. Do you have an example that shows the primary email and the repo email? I'm curious to see this edge case in real life 🙂

If I use this branch, but go back to using `committer_email`, then with

`verify_cot --task-type build --cot-product xpi --cleanup -- YeSaMaavTsqrNAARnu-_oQ`

I get an `event` of

```
 'event': {'after': '0132dc5c1966cc2fcecb7ba0c946be8935d8593e',
           'pusher': {'email': 'aki@escapewindow.com'},
           'ref': 'refs/heads/master',
           'repository': {'full_name': 'escapewindow/xpi-manifest',
                          'html_url': 'https://github.com/escapewindow/xpi-manifest',
                          'name': 'xpi-manifest',
                          'pushed_at': None},
           'sender': {'login': 'escapewindow'}},
 'now': '2019-10-16T03:12:11.316Z',
 'ownTaskId': 'Oz2sARieTIWPt_CO7WnRrA',
 'taskId': None,
 'tasks_for': 'github-push'}
```

but in the [decision task](https://queue.taskcluster.net/v1/task/Oz2sARieTIWPt_CO7WnRrA) I get an `--owner='aki@mozilla.com'`. With the old `cot.verify` logic, this breaks since the emails don't match. I went into the github api docs to see how we could retrieve all the email addresses of a given login so we could verify the email used belongs to the login, but it looks like github prevents seeing all the email addresses of any user but yourself.

escapewindow:
Looks like I still need to add github `tasks_for == 'action'` support in `cot.verify`.

escapewindow:
Rebased against the new mpd001 work.

PR #403, opened 2019-10-29
Add support for running in alternative clusters.

tomprince:


scriptworker/cot/verify.py
```diff
@@ -2091,11 +2091,12 @@ def format(self, record):
         context = Context()
         context.session = session
         context.config = dict(deepcopy(DEFAULT_CONFIG))
+        context.config['taskcluster_root_url'] = os.environ.get('TASKCLUSTER_ROOT_URL', 'https://taskcluster.net')
```
escapewindow: We can do this `os.environ.get` in `scriptworker.constants.DEFAULT_CONFIG` for expediency.

escapewindow:
> Overall I'm good with this. We seem to have test bustage. Can you track that down?

Looks like we need to keep backwards compatibility until after the cutover.

PR #409, opened 2019-11-08
switch default taskcluster_root_url to firefox-ci

escapewindow:
We shouldn't land and release this til Saturday.


escapewindow:
We should address the integration and production test bustage, although it doesn't technically block merging.

Issue #419, opened 2020-01-08
CoT intermittently fails to connect to Github


JohanLorenzo:
It's a recurrent issue that occurs in most android-component releases. The latest occurrence is:

```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/lib/python3.7/site-packages/scriptworker/worker.py", line 55, in do_run_task
    await run_cancellable(verify_chain_of_trust(chain))
  File "/app/lib/python3.7/site-packages/scriptworker/worker.py", line 158, in _run_cancellable
    result = await self.future
  File "/app/lib/python3.7/site-packages/scriptworker/cot/verify.py", line 2078, in verify_chain_of_trust
    await trace_back_to_tree(chain)
  File "/app/lib/python3.7/site-packages/scriptworker/cot/verify.py", line 2024, in trace_back_to_tree
    if restricted_privs and await chain.is_try_or_pull_request():
  File "/app/lib/python3.7/site-packages/scriptworker/cot/verify.py", line 143, in is_try_or_pull_request
    conditions = await raise_future_exceptions(tasks)
  File "/app/lib/python3.7/site-packages/scriptworker/utils.py", line 354, in raise_future_exceptions
    succeeded_results, _ = await _process_future_exceptions(tasks, raise_at_first_error=True)
  File "/app/lib/python3.7/site-packages/scriptworker/utils.py", line 388, in _process_future_exceptions
    raise exc
  File "/app/lib/python3.7/site-packages/scriptworker/cot/verify.py", line 268, in is_try_or_pull_request
    return await is_try_or_pull_request(self.context, self.task)
  File "/app/lib/python3.7/site-packages/scriptworker/task.py", line 494, in is_try_or_pull_request
    return await is_pull_request(context, task)
  File "/app/lib/python3.7/site-packages/scriptworker/task.py", line 474, in is_pull_request
    github_repository = GitHubRepository(repo_owner, repo_name, context.config['github_oauth_token'])
  File "/app/lib/python3.7/site-packages/scriptworker/github.py", line 38, in __init__
    self._github_repository = GitHub(token=token).repository(owner, repo_name)
  File "/app/lib/python3.7/site-packages/github3/github.py", line 1981, in repository
    json = self._json(self._get(url), 200)
  File "/app/lib/python3.7/site-packages/github3/models.py", line 211, in _get
    return self._request("get", url, **kwargs)
[...]

PR #420, opened 2020-01-08
Make scriptworker retry more whenever CoT fails to connect to Github

JohanLorenzo:
Fixes https://github.com/mozilla-releng/scriptworker/issues/419. What do you think @escapewindow?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,19 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+[29.1.0] - 2019-01-09
+---------------------
+
+Added
+~~~~~
+
+- utils.retry_sync() to enable retries on functions that cannot be asyncronous.
+
+Fixed
+~~~~~
+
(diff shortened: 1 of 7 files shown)


JohanLorenzo:
The CI bustage is going to be fixed in #421

JohanLorenzo:
Makes sense. I wasn't sure of the `retry_sync` approach because the library already retries requests for us. So we may end up hammering github with 5 * 5 requests in a single task run. That sounds quite big to me. What do you think @escapewindow? I'm fine changing this patch to `retry_sync` it.

escapewindow:
Setting the task to intermittent-task means we may retry immediately in the next task run, so we may get 5*5 attempts, with minimal sleep time in between. With a `retry_sync`, we have the ability to change the sleep time logic for a longer sleep time (30s? `n` minutes?), since we already tried 5x in the GitHub object.

Having said that, I'm not 100% sure if this will be an issue, so I'm ok with holding off on `retry_sync` til we know we need it.

JohanLorenzo:
You're right, using `retry_sync` would give a better control on the waiting. I gave it a shot. Please let me know what you think of it 🙂 @escapewindow

PR #422, opened 2020-01-10
add new docker-worker-gcp CoT public key

milescrabill:
This adds a new/separate Chain of Trust public key that will be used for workers in GCP

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -79,8 +79,9 @@ DEFAULT_CONFIG = frozendict(
         "ed25519_private_key_path": "...",
         "ed25519_public_keys": frozendict(
             {
-                "docker-worker": tuple(["J+PAKmq3jkS2uCpBk5WU2ycrnTFPwZujJT4OHAxm38I="]),
-                "docker-worker-gcp": tuple(["tk/SjxY3mREARba6ODw7qReUoVWj0RgEIxBURkwcM4I="]),
+                # the key beginning in J+PAK... is the longstanding docker-worker CoT key, to be phased out
+                # the key beginning in tk/Sj... is a new key added to differentiate GCP workers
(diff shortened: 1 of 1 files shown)


src/scriptworker/constants.py
```diff
@@ -80,6 +80,7 @@
         "ed25519_public_keys": frozendict(
             {
                 "docker-worker": tuple(["J+PAKmq3jkS2uCpBk5WU2ycrnTFPwZujJT4OHAxm38I="]),
+                "docker-worker-gcp": tuple(["tk/SjxY3mREARba6ODw7qReUoVWj0RgEIxBURkwcM4I="]),
```
escapewindow: These are [keyed off of `worker_impl`](https://github.com/mozilla-releng/scriptworker/blob/eeff110efe803c64303252cb1b947cbeda08acb4/src/scriptworker/cot/verify.py#L777). We currently don't have a way of differentiating between gcp docker-worker and non-gcp docker-worker, and it's not clear we should. This means we should add this to the `docker-worker` tuple. Essentially,

```
"docker-worker": tuple(["J+PAKmq3jkS2uCpBk5WU2ycrnTFPwZujJT4OHAxm38I=", "tk/SjxY3mREARba6ODw7qReUoVWj0RgEIxBURkwcM4I="]),
```

Alternately,

```
"docker-worker": ("J+PAKmq3jkS2uCpBk5WU2ycrnTFPwZujJT4OHAxm38I=", "tk/SjxY3mREARba6ODw7qReUoVWj0RgEIxBURkwcM4I="),
```

We probably want to add a comment or file an issue to nuke the `J+PAK...` pubkey when it's no longer used.

milescrabill: That makes sense, thanks for the feedback!

milescrabill:
Next week would be great! I'm going to be out, so please sync up with @walac for deployment timing.

PR #425, opened 2020-01-28
Mypy

Callek:
First pass at starting to add mypy to scriptworker

### DIFF
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -112,17 +112,17 @@ class Context(object):
         if self._credentials:
             return dict(deepcopy(self._credentials))
 
-    @property
-    def task_id(self):
-        """string: The running task's taskId."""
-        if self.claim_task:
-            return self.claim_task["status"]["taskId"]
-
     @credentials.setter
     def credentials(self, creds):
         self._credentials = creds
         self.queue = self.create_queue(self.credentials)
         self.credentials_timestamp = arrow.utcnow().timestamp
+
+    @property
+    def task_id(self):
+        """string: The running task's taskId."""
+        if self.claim_task:
+            return self.claim_task["status"]["taskId"]
 
     def create_queue(self, credentials):
         """Create a taskcluster queue.
--- a/src/scriptworker/exceptions.py
+++ b/src/scriptworker/exceptions.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 """scriptworker exceptions."""
 
+from typing import Any
+
 from scriptworker.constants import STATUSES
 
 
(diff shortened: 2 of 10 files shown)


src/scriptworker/context.py
```diff
@@ -112,18 +112,18 @@ def credentials(self):
         if self._credentials:
             return dict(deepcopy(self._credentials))
 
-    @property
-    def task_id(self):
-        """string: The running task's taskId."""
-        if self.claim_task:
-            return self.claim_task["status"]["taskId"]
-
     @credentials.setter
```
Callek: this is due to a mypy bug, which flags @credentials.setter as wrong unless it is right next to the @property getter.

src/scriptworker/exceptions.py
```diff
@@ -46,18 +48,17 @@ class ScriptWorkerTaskException(ScriptWorkerException):
...
 
-    def __init__(self, *args, exit_code=1, **kwargs):
+    def __init__(self, *args: Any, exit_code: int = 1):
         """Initialize ScriptWorkerTaskException.
 
         Args:
             *args: These are passed on via super().
             exit_code (int, optional): The exit_code we should exit with when
                 this exception is raised.  Defaults to 1 (failure).
-            **kwargs: These are passed on via super().
```
Callek: kwargs is invalid on `BaseException`

src/scriptworker/version.py
```diff
@@ -40,9 +44,9 @@ def get_version_string(version):
     """
     version_len = len(version)
     if version_len == 3:
-        version_string = "%d.%d.%d" % version
+        version_string = "%d.%d.%d" % cast(ShortVerType, version)
     elif version_len == 4:
-        version_string = "%d.%d.%d%s" % version
+        version_string = "%d.%d.%d%s" % cast(LongVerType, version)
```
Callek: this dance was needed because mypy didn't identify the Tuple Length variants between the len== piece here, and flagged both branches of the `Union`, specifically too-few for the ==3 side, and too-many for the ==4

src/scriptworker/context.py
escapewindow: That's probably preferred sorting anyway, no worries :)

src/scriptworker/utils.py
```diff
@@ -217,7 +218,15 @@ def calculate_sleep_time(attempt, delay_factor=5.0, randomization_factor=0.5, ma
...
-async def retry_async(func, attempts=5, sleeptime_callback=calculate_sleep_time, retry_exceptions=Exception, args=(), kwargs=None, sleeptime_kwargs=None):
+async def retry_async(
+    func: Callable[..., Awaitable[Any]],
+    attempts: int = 5,
+    sleeptime_callback: Callable[..., Any] = calculate_sleep_time,
+    retry_exceptions: Union[Type[BaseException], Tuple[Type[BaseException], ...]] = Exception,
+    args: Sequence[Any] = (),
+    kwargs: Optional[Dict[str, Any]] = None,
+    sleeptime_kwargs: Optional[Dict[str, Any]] = None,
+) -> Any:
```
escapewindow: Cool, good to see how we can encapsulate complex args and kwargs in typing.
It looks like you're intentionally only covering specific functions? Is there any logic to which?

Callek: I'm tackling one whole file at a time; and then the functions I'm in related files that I am touching, are ones that are causing errors in the files I am touching -- e.g. `error: Call to untyped function "retry_async" in typed context`

Callek:
> Looks good, and thank you! Wondering if there was any impetus to this work? Just scratching an itch?

Yep just scratching an itch, I wanted to work with mypy more and get to know its use cases better. This could help us catch actual bugs in the code, and make our changes in the future happen with more confidence.

Callek:
For future reference, when landed mypy was `mypy==0.761`

PR #432, opened 2020-02-14
Py38

### LINKED ISSUE #415: add python 3.8 support to scriptworker
- add the tests
- get them green
- run them in automation
- block merges on non-green py38 tests

escapewindow:
Fixes #415 . cc @rail

### DIFF
--- a/tests/test_artifacts.py
+++ b/tests/test_artifacts.py
@@ -25,7 +25,7 @@ from scriptworker.artifacts import (
 )
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 
-from . import create_finished_future, create_rejected_future, touch
+from . import touch
 
 
 @pytest.yield_fixture(scope="function")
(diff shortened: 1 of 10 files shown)


escapewindow:
Locally, the only test that fails is `test_run_tasks_cancel_right_before_cot`.

In Taskcluster, it looks like

- `test_run_tasks_no_cancel`
- `test_run_tasks_cancel_right_before_cot`
- `test_run_tasks_cancel_run_tasks`
- `test_run_tasks_cancel_cot`
- `test_run_tasks_cancel_right_before_proc_created`

I believe these errors are all around our signal handling. We probably want to fix this. An alternative might be to tear out the signal handling logic, and deal with hard shutdowns in the pool.

escapewindow:
Let's deprecate `create_finished_future` , and use [`AsyncMock`](https://docs.python.org/3/library/unittest.mock-examples.html#mocking-asynchronous-iterators) for py38.

escapewindow:
We could add a `py38-cot` target.

tests/test_artifacts.py
```diff
@@ -84,15 +84,18 @@ def test_expiration_arrow(context):
...
-
-    generator = mock_create_artifact()
-    with mock.patch("scriptworker.artifacts.create_artifact", new=lambda *args, **kwargs: next(generator)):
-        with pytest.raises(ArithmeticError):
-            await upload_artifacts(context, ["one", "public/two"])
+async def test_upload_artifacts_throws(context, mocker):
+    exceptions = [None, ArithmeticError]
+
+    async def mock_create_artifact(*args, **kwargs):
+        exc = exceptions.pop()
```
escapewindow: hm, this might want to be `pop(0)`

PR #433, opened 2020-02-21
Bug 1617128 - Enable fennec-profile-manager project

JohanLorenzo:


src/scriptworker/constants.py
```diff
@@ -361,16 +361,27 @@
                         {
                             "project:mobile:android-components:releng:beetmover:bucket:maven-production": "android-components-repo",
                             "project:mobile:android-components:releng:beetmover:bucket:maven-snapshot-production": "android-components-repo",
+
                             "project:mobile:fenix:releng:googleplay:product:fenix": "fenix-repo",
                             "project:mobile:fenix:releng:signing:cert:nightly-signing": "fenix-repo",
                             "project:mobile:fenix:releng:signing:cert:beta-signing": "fenix-repo",
                             "project:mobile:fenix:releng:signing:cert:production-signing": "fenix-repo",
+                            "project:mobile:fenix:releng:signing:cert:fennec-nightly-signing": "fenix-repo",
```
JohanLorenzo: I just realized these `fenix` scopes were not guarded.

src/scriptworker/constants.py
```diff
@@ -453,11 +464,12 @@
                     ),
                     "mobile": immutabledict(
                         {
+                            "android-components-repo": ("/mozilla-mobile/android-components",),
```
JohanLorenzo: I just alphabetically ordered these lists.

src/scriptworker/constants.py
```diff
@@ -453,11 +470,12 @@
...
                             "fenix-repo": ("/mozilla-mobile/fenix",),
+                            "fennec-profile-manager": ("/mozilla-mobile/fennec-profile-manager",),
```
MihaiTabara: question: did we want a trailing `-repo` to be consistent with others or should we not bother given that this is going away once adhoc signing is up?

src/scriptworker/constants.py
```diff
@@ -365,12 +376,18 @@
                             "project:mobile:fenix:releng:signing:cert:nightly-signing": "fenix-repo",
                             "project:mobile:fenix:releng:signing:cert:beta-signing": "fenix-repo",
                             "project:mobile:fenix:releng:signing:cert:production-signing": "fenix-repo",
+                            "project:mobile:fenix:releng:signing:cert:fennec-nightly-signing": "fenix-repo",
+                            # Fennec Beta and Fennec Release share the same key
+                            "project:mobile:fenix:releng:signing:cert:fennec-production-signing": "fenix-repo",
+                            "project:mobile:fennec-profile-manager:releng:signing:cert:fennec-nightly-signing": "fenix-repo",
```
MihaiTabara: `s/fenix-repo/fennec-profile-manager`for this line and the next in their values? I'm assuming we want to be the value we defined at https://github.com/mozilla-releng/scriptworker/pull/433/files#diff-4635e904df4496f16271f8684ac489f5R475

src/scriptworker/constants.py
JohanLorenzo: Oops, you're right! Fixed!

src/scriptworker/constants.py
JohanLorenzo: 👍 Fixed

PR #434, opened 2020-02-21
Temp queue download

### LINKED ISSUE #426: download artifacts with temp_queue if possible
We currently download artifacts with `queue`, which uses the worker creds. This means for private artifacts, the worker clientId needs to have access to those private artifacts.

The `temp_queue` will be given access to private artifacts for those tasks that need them. So if we download artifacts with that, we no longer need to add private artifact support to the worker clientId. Let's do that.

escapewindow:
Fixes #426 .

### DIFF
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -234,12 +234,6 @@ class Context(object):
     def temp_queue(self, queue):
         self._temp_queue = queue
 
-    @credentials.setter
-    def credentials(self, creds):
-        self._credentials = creds
-        self.queue = self.create_queue(self.credentials)
-        self.credentials_timestamp = arrow.utcnow().timestamp
-
     async def populate_projects(self, force=False):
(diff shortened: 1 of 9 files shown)


src/scriptworker/context.py
```diff
@@ -169,7 +169,10 @@ def temp_credentials(self):
     @temp_credentials.setter
     def temp_credentials(self, credentials):
         self._temp_credentials = credentials
-        self.temp_queue = self.create_queue(self.temp_credentials)
+        if credentials is not None:
+            self.temp_queue = self.create_queue(self.temp_credentials)
+        else:
+            self.temp_queue = None
```
nthomas-mozilla: Out of curiosity, when would we reach this `else` case ?

escapewindow: Here? https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/context.py#L91

PR #435, opened 2020-02-24
Bug 1617612 - add CoT support for flatpaks

MihaiTabara:


escapewindow:
Oops, we'll need 32.0.1

PR #438, opened 2020-02-28
Bug 1618940: Add `action_perm` to the context for actions.

tomprince:


tomprince:
This is definitely low priority, I'll probably revisit it in a week or two (and would have no problem with someone picking it up in the mean time).

PR #440, opened 2020-03-05
Add new repo url

Callek:


Callek:
This is the new guardian repo, noteworthy its not actually a private repo....

Callek:
Reflagged for review - I decided to drop the secret-repo magic, due to the fact that we have the old repo with old scriptworker pinned, and I'm less confident on how this will behave with both private and non-private repos in the configuration for same trust_domain.

escapewindow:
Looks like we have some private repo tests that broke? We may want to keep them, but use specific test cases rather than rely on the contents of constants.py ?

PR #443, opened 2020-03-13
scriptworker cot_product + cleanup

escapewindow:
This patch adds the scriptworker-scripts cot-product, and cleans up some old configs.

- there is no more aws-provisioner-v1 in https://firefox-ci-tc.services.mozilla.com/provisioners/ , so I removed all workers with that provisioner
- I populated the adhoc-3 workers, though they don't exist yet. Optional.
- I added scriptworker-scripts as a cot product. I wasn't sure which scopes would be restricted, and I'm open about the names.
- birch and jamun appear to be idle for the past year. esr60 is EOL. We don't appear to reference `all-staging-branches` anywhere.

I can split this up if preferred.

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -159,7 +159,7 @@ DEFAULT_CONFIG = immutabledict(
                                     r"^(?P<path>/mozilla-(central|unified))(/|$)",
                                     r"^(?P<path>/integration/(autoland|fx-team|mozilla-inbound))(/|$)",
                                     r"^(?P<path>/releases/mozilla-(beta|release|esr\d+))(/|$)",
-                                    r"^(?P<path>/projects/(birch|jamun|maple|oak))(/|$)",
+                                    r"^(?P<path>/projects/(maple|oak))(/|$)",
                                 ),
                             }
                         ),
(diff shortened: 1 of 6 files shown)


src/scriptworker/constants.py
```diff
@@ -116,43 +116,14 @@
...
+                        "firefox": ("gecko-1/decision", "gecko-2/decision", "gecko-3/decision",),
+                        "thunderbird": ("comm-1/decision", "comm-2/decision", "comm-3/decision",),
+                        "mobile": ("mobile-1/decision", "mobile-3/decision",),
+                        "mpd001": ("mpd001-1/decision", "mpd001-3/decision"),
+                        "application-services": ("app-services-1/decision", "app-services-3/decision",),
                         "xpi": ("xpi-1/decision", "xpi-3/decision"),
-                        # TODO: Add adhoc-3 whenever we're ready to go to prod
-                        "adhoc": ("adhoc-1/decision",),
+                        "adhoc": ("adhoc-1/decision", "adhoc-3/decision"),
+                        "scriptworker-scripts": ("swscripts-1/decision", "swscripts-3/decision",),
```
tomprince: I think I would use `scriptworker` both for the CoT product, and the trust domain (which is the worker prefix).

Although, thinking ahead, I wonder if we can use one CoT product + trust domain for all the releng projects which are deployed by cloudoups? If we go this route, I'd be inclined to use `releng`, and migrate the few projects using that that we don't deploy to another trust domain.

src/scriptworker/constants.py
```diff
@@ -116,43 +116,14 @@
...
-                            "aws-provisioner-v1/gecko-2-decision",
-                            "aws-provisioner-v1/gecko-3-decision",
-                            "comm-1/decision",
-                            "comm-2/decision",
-                            "comm-3/decision",
-                        ),
-                        "mobile": (
-                            # gecko-focus was for mozilla-mobile releases (bug 1455290) for more details.
-                            # TODO: Remove it once not used anymore
-                            "aws-provisioner-v1/gecko-focus",
```
JohanLorenzo: I'm glad to see it go away!

src/scriptworker/constants.py
```diff
@@ -162,38 +133,14 @@
...
-                        ),
+                        "firefox": ("gecko-1/images", "gecko-2/images", "gecko-3/images",),
+                        "thunderbird": ("comm-1/images", "comm-2/images", "comm-3/images",),
+                        "mobile": ("mobile-1/images", "mobile-3/images",),
+                        "mpd001": ("mpd001-1/images", "mpd001-3/images"),
+                        "application-services": ("app-services-1/images", "app-services-3/images",),
                         "xpi": ("xpi-1/images", "xpi-3/images"),
-                        # TODO: Add adhoc-3 whenever we're ready to go to prod
-                        "adhoc": ("adhoc-1/images",),
+                        "adhoc": ("adhoc-1/images", "adhoc-3/images"),
```
JohanLorenzo: 👍 to get adhoc-3 in, even though workers aren't there yet.

src/scriptworker/constants.py
```diff
@@ -212,7 +159,7 @@
                                     r"^(?P<path>/mozilla-(central|unified))(/|$)",
                                     r"^(?P<path>/integration/(autoland|fx-team|mozilla-inbound))(/|$)",
                                     r"^(?P<path>/releases/mozilla-(beta|release|esr\d+))(/|$)",
-                                    r"^(?P<path>/projects/(birch|jamun|maple|oak))(/|$)",
```
JohanLorenzo: I'm glad we now have try to run most staging releases \o/

src/scriptworker/constants.py
```diff
@@ -414,60 +373,38 @@
...
                         {
                             # Which repos can perform release actions?
-                            # XXX remove /projects/maple and birch when taskcluster relpro
-                            #     migration is tier1 and landed on mozilla-central
-                            # XXX remove /projects/jamun when we no longer run staging releases
-                            #     from it
-                            "all-release-branches": (
-                                "/releases/mozilla-beta",
-                                "/releases/mozilla-release",
-                                "/releases/mozilla-esr60",
```
JohanLorenzo: Bye-bye esr60 👋 ! 👍

src/scriptworker/constants.py
```diff
@@ -414,60 +373,38 @@
...
+                            "esr": ("/releases/mozilla-esr68",),
                             "esr68": ("/releases/mozilla-esr68",),
                             "nightly": ("/mozilla-central",),
                             # Which repos can do nightly signing?
-                            # XXX remove /projects/maple and birch when taskcluster relpro
-                            #     migration is tier1 and landed on mozilla-central
-                            # XXX remove /projects/jamun when we no longer run staging releases
-                            #     from it
+                            # XXX remove /projects/maple when we have a
+                            #     different prod signing testing solution
```
JohanLorenzo: Thanks for the comment 👍

src/scriptworker/constants.py
```diff
@@ -493,11 +431,13 @@
                 {
                     "firefox": ("decision", "action", "docker-image"),
                     "thunderbird": ("decision", "action", "docker-image"),
+                    # XXX now that we're on taskgraph, we should limit these.
```
JohanLorenzo: Agreed!

src/scriptworker/constants.py
escapewindow: Hm. I think we may have technical issues -- solvable, but not necessarily without some work -- that keep us from using `releng` if, say, the trusted repos are in separate Github organizations. I do think we should be able to reuse the same pools across both the `scriptworker` and `scriptworker-scripts` repos, though, so I think going with `scriptworker` makes a lot of sense.

escapewindow: Done. I also added the `project:scriptworker:pypi:production` restricted scope, in case we decide to do automatic pypi pushes from `scriptworker{,-scripts}`. These may change in the future.

PR #446, opened 2020-03-25
bug 1624794 - catch TimeoutError on upload

escapewindow:


src/scriptworker/worker.py
```diff
@@ -90,7 +90,7 @@
     except ScriptWorkerException as e:
         status = worst_level(status, e.exit_code)
         log.error("Hit ScriptWorkerException: {}".format(e))
-    except aiohttp.ClientError as e:
+    except (aiohttp.ClientError, asyncio.TimeoutError) as e:
         status = worst_level(status, STATUSES["intermittent-task"])
         log.error("Hit aiohttp error: {}".format(e))
```
nthomas-mozilla: Might be worth tweaking the error message here - something something network ?

src/scriptworker/worker.py
```diff
@@ -90,7 +90,7 @@
     except ScriptWorkerException as e:
         status = worst_level(status, e.exit_code)
         log.error("Hit ScriptWorkerException: {}".format(e))
-    except aiohttp.ClientError as e:
+    except (aiohttp.ClientError, asyncio.TimeoutError) as e:
```
nthomas-mozilla: Does this change also need to go into https://github.com/mozilla-releng/scriptworker/blob/3f71b1a0713fc2b946e13530d0fc4035a3fd1f90/src/scriptworker/artifacts.py#L132 to ensure we retry ?

escapewindow:
Thanks for catching this, and for the review!

PR #450, opened 2020-04-06
Bug 1618731 - retry getting task definition

escapewindow:
This also detects if the task definition we receive is an empty `{}` (currently just by checking the existence of `payload`). In theory that should be a taskcluster error.

### DIFF
--- a/src/scriptworker/cot/verify.py
+++ b/src/scriptworker/cot/verify.py
@@ -69,7 +69,6 @@ from scriptworker.utils import (
 )
 from scriptworker.version import __version_string__
 from taskcluster.aio import Queue
-from taskcluster.exceptions import TaskclusterFailure
 
 log = logging.getLogger(__name__)
 
@@ -571,17 +570,14 @@ async def build_link(chain, task_name, task_id):
     """
     link = LinkOfTrust(chain.context, task_name, task_id)
     json_path = link.get_artifact_full_path("task.json")
-    try:
-        task_defn = await retry_get_task_definition(chain.context.queue, task_id)
-        link.task = task_defn
-        chain.links.append(link)
-        # write task json to disk
-        makedirs(os.path.dirname(json_path))
-        with open(json_path, "w") as fh:
-            fh.write(format_json(task_defn))
-        await build_task_dependencies(chain, task_defn, task_name, task_id)
-    except TaskclusterFailure as exc:
-        raise CoTError(str(exc))
+    task_defn = await retry_get_task_definition(chain.context.queue, task_id, exception=CoTError)
+    link.task = task_defn
+    chain.links.append(link)
+    # write task json to disk
+    makedirs(os.path.dirname(json_path))
+    with open(json_path, "w") as fh:
+        fh.write(format_json(task_defn))
+    await build_task_dependencies(chain, task_defn, task_name, task_id)
 
 
 async def build_task_dependencies(chain, task, name, my_task_id):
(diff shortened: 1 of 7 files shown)


src/scriptworker/task.py
```diff
@@ -51,6 +52,39 @@ def worst_level(level1, level2):
...
+# get_task_definition {{{1
+async def get_task_definition(queue, task_id, exception=TaskclusterFailure):
+    """Get the task definition from the queue.
+
+    Detect whether the task definition is empty, per bug 1618731.
+
+    Args:
+        queue (taskcluster.aio.Queue): the taskcluster Queue object
+        task_id (str): the taskId of the task
+        exception (Exception, optional): the exception to raise if unsuccessful.
```
JohanLorenzo, line 64: Question: What is the pro of making `exception` configurable?

src/scriptworker/task.py
```diff
@@ -51,6 +52,39 @@ def worst_level(level1, level2):
...
+            Defaults to ``TaskclusterFailure``.
+
+    """
+    task_defn = await queue.task(task_id)
+    if "payload" not in task_defn:
+        raise exception("Task definition for {} is empty!\n {}".format(task_id, task_defn))
```
JohanLorenzo, line 70: 👍 Good idea on erroring out early!

src/scriptworker/task.py
```diff
@@ -51,6 +52,39 @@ def worst_level(level1, level2):
...
+    """Retry ``get_task_definition``.
+
+    Args:
+        queue (taskcluster.aio.Queue): the taskcluster Queue object
+        task_id (str): the taskId of the task
+        exception (Exception, optional): the exception to raise if unsuccessful.
+            Defaults to ``TaskclusterFailure``.
+
+    """
+    kwargs.setdefault("retry_exceptions", tuple(set([TaskclusterFailure, exception])))
```
JohanLorenzo, line 84: 👍

src/scriptworker/cot/verify.py
```diff
@@ -570,17 +570,14 @@ def _sort_dependencies_by_name_then_task_id(dependencies):
...
-    try:
-        task_defn = await chain.context.queue.task(task_id)
-        link.task = task_defn
-        chain.links.append(link)
-        # write task json to disk
-        makedirs(os.path.dirname(json_path))
-        with open(json_path, "w") as fh:
-            fh.write(format_json(task_defn))
-        await build_task_dependencies(chain, task_defn, task_name, task_id)
-    except TaskclusterFailure as exc:
```
JohanLorenzo, line 582: 👍 for moving this try/catch outside of this function.

tests/test_cot_verify.py
```diff
@@ -654,23 +653,20 @@ def test_find_sorted_task_dependencies(task, expected, task_type):
 @pytest.mark.asyncio
 async def test_build_task_dependencies(chain, mocker):
     async def fake_task(task_id):
-        if task_id == "die":
-            raise TaskclusterFailure("dying")
```
JohanLorenzo, line 658: Nice cleanup! I see `build_task_dependencies` doesn't raise `TaskclusterFailure` (anymore?)

src/scriptworker/task.py
escapewindow: I wanted to use `CoTError` in `cot.verify`, but that exception doesn't necessarily make sense if we call this function elsewhere in the future.

tests/test_cot_verify.py
escapewindow: I think this was to get coverage in the try/except that I removed above. Now that we're throwing `CoTError` on a bad `queue.task()` result instead of `TaskclusterFailure`, we don't have to catch `TaskclusterFailure` and make sure we're throwing `CoTError`.

escapewindow:
Hm, I've lost coverage... fixing

tests/test_cot_verify.py
escapewindow: Looks like I needed to keep this test code, but change the exception to `CoTError`, for full coverage.

PR #451, opened 2020-04-09
catch TimeoutError during claimWork

escapewindow:


escapewindow:
Thanks! I also added it to notarization_poller with my uuid work.

PR #452, opened 2020-04-16
check generic-worker interactive tasks. raise on docker-worker interactive tasks.

escapewindow:
The generic-worker rdp info is [here](https://github.com/taskcluster/taskcluster/blob/05c5897fafc9d96f5c7f8f8e9dd0ecb08831c358/workers/generic-worker/schemas/multiuser_windows.yml#L296-L325).

This patch:
1. changes `verify_docker_worker_task` to run `check_interactive_docker_worker`, even if the target task is the `chain` task (i.e., the specific task we want to verify the chain against). Since `check_interactive_docker_worker` doesn't check `link.cot`, we don't require that task to have finished yet.
2. raises `CoTError` on an interactive docker-worker task. Previously we returned a list of errors, which we then ignored. I'm not thrilled with this, but we're partially protected by ci-admin checks and the fact that the interactive flag would need to be set in-tree to affect a release graph.
3. adds a `check_interactive_generic_worker`, which we call in `verify_generic_worker_task`.

@JohanLorenzo @tomprince what do you think?

### DIFF
--- a/src/scriptworker/cot/verify.py
+++ b/src/scriptworker/cot/verify.py
@@ -365,7 +365,6 @@ def get_valid_worker_impls():
             validation functions.
 
     """
-    # TODO support taskcluster worker
     return immutabledict({"docker-worker": verify_docker_worker_task, "generic-worker": verify_generic_worker_task, "scriptworker": verify_scriptworker_task})
 
 
@@ -436,8 +435,8 @@ def check_interactive_docker_worker(link):
     Args:
         link (LinkOfTrust): the task link we're checking.
 
-    Returns:
-        list: the list of error errors.  Success is an empty list.
+    Raises:
+        CoTError: on interactive.
 
     """
     errors = []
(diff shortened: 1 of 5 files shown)


src/scriptworker/cot/verify.py
```diff
@@ -449,7 +448,34 @@ def check_interactive_docker_worker(link):
...
-    return errors
+    raise_on_errors(errors)
+
+
+# check_interactive_generic_worker {{{1
+def check_interactive_generic_worker(link):
+    """Given a task, make sure the task was not defined as interactive.
+
+    * ``task.payload.rdpInfo`` must be absent or False.
+    * ``task.payload.env.TASKCLUSTER_INTERACTIVE`` must be absent or False.
```
JohanLorenzo, line 459: Nit: This condition isn't tested. Let's change it for the scope check 🙂

escapewindow:
Thanks!

PR #449, opened 2020-04-01
Bug 1613800 - part 2: add old docker-worker cot key back

JohanLorenzo:
This reverts https://github.com/mozilla-releng/scriptworker/pull/447. It broke mobile releases:

```
2020-04-01T09:34:52    DEBUG -     Verifying the scriptworker:build:docker-image aP4nedQnTAWQl2fKawVGwg docker-worker ed25519 chain of trust signature
2020-04-01T09:34:52 CRITICAL - Chain of Trust verification error!
Traceback (most recent call last):
  File "/app/lib/python3.8/site-packages/scriptworker/cot/verify.py", line 1905, in verify_chain_of_trust
    verify_cot_signatures(chain)
  File "/app/lib/python3.8/site-packages/scriptworker/cot/verify.py", line 836, in verify_cot_signatures
    verify_link_ed25519_cot_signature(chain, link, unsigned_path, ed25519_signature_path)
  File "/app/lib/python3.8/site-packages/scriptworker/cot/verify.py", line 815, in verify_link_ed25519_cot_signature
    raise CoTError(message)
scriptworker.exceptions.CoTError: '"scriptworker:build:docker-image aP4nedQnTAWQl2fKawVGwg: docker-worker ed25519 cot signature doesn\'t verify against tk/SjxY3mREARba6ODw7qReUoVWj0RgEIxBURkwcM4I=: "'
```

https://firefox-ci-tc.services.mozilla.com/tasks/BBN7xM39Tzi4EZXojWXj5A/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FBBN7xM39Tzi4EZXojWXj5A%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Fchain_of_trust.log#L54

Full task group: https://firefox-ci-tc.services.mozilla.com/tasks/groups/LeSantlsRcKDxmUhGC9o9w

I guess there's a better solution, like maybe rebuilding old docker images. That said, I found some new gecko workers were created (for instance https://firefox-ci-tc.services.mozilla.com/provisioners/gecko-3/worker-types/b-linux-gcp), but mobile ones don't have them (yet?) https://firefox-ci-tc.services.mozilla.com/provisioners/mobile-3.

This issue is blocking the Fennec to Fenix beta migration which is currently ongoing. Thus, I prefer to re-enable this key and deprecate it later. @escapewindow, please let me know if you need more context 🙂

PR #453, opened 2020-04-17
Bug 1613800 - part 3 to 5: Add tests for missing mobile products and remove old key

JohanLorenzo:
Follows https://github.com/mozilla-releng/scriptworker/pull/449 up. Depends on https://github.com/mozilla-mobile/android-components/pull/6698.


JohanLorenzo:
I think this PR is ready to reviewed. The tests are failing because I hammered the Github API with many pushes. I'll rerun them in a few hours. I don't foresee any failures.

JohanLorenzo:
Ah crap, it turns out we need a Github token if we want to run all tests.

escapewindow:
> Ah crap, it turns out we need a Github token if we want to run all tests.

We're on TC now. We may want to put a token in tc secrets.

JohanLorenzo:
Depends on https://phabricator.services.mozilla.com/D77797. I manually added the scope to get these tasks green, for now. 

I'd like to get this reviewed once more to make sure I didn't mess up anything secret-wise. r? @escapewindow

tests/test_production.py
```diff
@@ -53,11 +54,22 @@ def build_config(override, basedir):
     return config
 
 
+async def _get_github_token():
```
escapewindow, line 57: We may end up tweaking this to make testing locally easier, but I think this works for now.

.taskcluster.yml
```diff
@@ -37,8 +37,12 @@ tasks:
...
            # See https://github.com/mozilla-releng/scriptworker/issues/415
            ['py38', 'python:3.8', {NO_TESTS_OVER_WIRE: '1'}, []],
-           ['py37-cot', 'python:3.7', {NO_CREDENTIALS_TESTS: '1'}, []],
-           ['py38-cot', 'python:3.8', {NO_CREDENTIALS_TESTS: '1'}, []],
+           ['py37-cot', 'python:3.7', {
+                NO_CREDENTIALS_TESTS: '1'
+            }, ['secrets:get:repo:github.com/mozilla-releng/scriptworker:github']],
+           ['py38-cot', 'python:3.8', {
+                NO_CREDENTIALS_TESTS: '1',
+            }, ['secrets:get:repo:github.com/mozilla-releng/scriptworker:github']],
```
escapewindow, line 45: Hm. Now that we have secrets, we could add integration tests.
This is a comment for future enhancements (for anyone to do, not necessarily you), and definitely not a blocker for this PR :)

PR #454, opened 2020-05-01
bug 1594621 - add cot_product_type map

escapewindow:
In [this adhoc task](https://firefox-ci-tc.services.mozilla.com/tasks/HNPN1sluQ8mTmLNaG3KB0w), we got a bad `tasks_for` error, because we have a large `if` statement that whitelists a bunch of `cot_product`s.

Let's create a `cot_product_type` map that we can check, so we don't have to update this `if` block every time we add a new github product.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,24 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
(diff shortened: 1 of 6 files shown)


escapewindow:
https://firefox-ci-tc.services.mozilla.com/tasks/HNPN1sluQ8mTmLNaG3KB0w run 0 failed CoT without this patch. Run 1 passed CoT with this patch, but failed later :)

PR #455, opened 2020-05-22
fix xpi private cot verification

escapewindow:
Previously, we assumed that we would list any and all private github repos in `constants.py`. However, with the xpi project, we'll get many private repos that we will want to verify without needing to land a change in scriptworker.

Let's allow for `ssh://github.com` source urls to imply a private repository.

As ridealongs, I added `--verbose` and `--no-check-task` to `verify-cot`. I also cleaned up some extraneous output at the end of `verify_cot`.

Also, the second patch gets us to stop creating `...` directories while running `test_production.py` tests.

The `base_*_dir` and `task_log_dir_template` config entries were from a historic, unmerged PR. Let's set the actual config paths inside our tempdir. Bonus: we won't have weird bustage related to hardcoded directory paths if and when we run two concurrent sets of tests.

I tested this patchset against the tasks in https://firefox-ci-tc.services.mozilla.com/tasks/groups/NFYpj38mQdSJGRxqG44_Iw , after setting `SCRIPTWORKER_GITHUB_OAUTH_TOKEN` to a token with read access to the private repo. I can invite you to my `test-xpi-private` repo if you want to give it a spin.

@Callek , @JohanLorenzo do you have cycles to look at this?

### DIFF
--- a/src/scriptworker/cot/verify.py
+++ b/src/scriptworker/cot/verify.py
@@ -1303,7 +1303,9 @@ async def get_in_tree_template(link):
         raise CoTError("{} source url {} doesn't end in .yml or .yaml!".format(link.name, source_url))
 
     auth = None
-    if (any(vcs_rule.get("require_secret") for vcs_rule in context.config["trusted_vcs_rules"])) and "github.com" in source_url:
+    if (
+        source_url.startswith("ssh://") or any(vcs_rule.get("require_secret") for vcs_rule in context.config["trusted_vcs_rules"])
+    ) and "github.com" in source_url:
         newurl = re.sub(
             r"^(?:ssh://|https?://)(?:[^@/\:]*(?:\:[^@/\:]*)?@)?github.com(?:\:\d*)?/(?P<repopath>.*)/raw/(?P<sha>[a-zA-Z0-9]*)/(?P<filepath>.*)$",
             r"https://raw.githubusercontent.com/\g<repopath>/\g<sha>/\g<filepath>",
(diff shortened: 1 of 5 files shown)


tests/test_production.py
```diff
@@ -45,6 +46,10 @@ def build_config(override, basedir):
     with open(os.path.join(basedir, "config.json"), "w") as fh:
         json.dump(config, fh, indent=2, sort_keys=True)
     config = apply_product_config(config)
+    # Avoid creating a `...` directory
+    for k,v in config.items():
+        if v == '...':
+            raise Exception(f"Let's not keep any '...' config values. {k} is {v}!")
```
JohanLorenzo, line 52: Nice 👍

src/scriptworker/cot/verify.py
```diff
@@ -1303,7 +1303,9 @@ async def get_in_tree_template(link):
         raise CoTError("{} source url {} doesn't end in .yml or .yaml!".format(link.name, source_url))
 
     auth = None
-    if (any(vcs_rule.get("require_secret") for vcs_rule in context.config["trusted_vcs_rules"])) and "github.com" in source_url:
+    if (
+        source_url.startswith("ssh://") or any(vcs_rule.get("require_secret") for vcs_rule in context.config["trusted_vcs_rules"])
+    ) and "github.com" in source_url:
```
Callek, line 1308: I'd be ok if you wanted to drop `require_secret` entirely -- guardian no longer uses it... but its fine to leave.

src/scriptworker/cot/verify.py
```diff
@@ -2008,11 +2007,14 @@ def verify_cot_cmdln(args=None, event_loop=None):
...
+    parser.add_argument("--verbose", "-v", help="enable debug logging", action="store_true", default=False)
+    parser.add_argument("--no-check-task", help="skip verifying the taskId's cot status", action="store_true", default=False)
     opts = parser.parse_args(args)
     tmp = tempfile.mkdtemp()
     log = logging.getLogger("scriptworker")
-    log.setLevel(logging.DEBUG)
-    logging.basicConfig()
+    level = logging.DEBUG if opts.verbose else logging.INFO
+    log.setLevel(level)
+    logging.basicConfig(level=level)
```
Callek, line 2017: :+1:

src/scriptworker/cot/verify.py
escapewindow: Hm. XPI doesn't have the requirement of releasing off these private source repos (we release off the manifest repo, which is public and known). We still have a check in `is_pull_request` that checks for `has_commit_landed_on_repository`. That check is essentially skipped if we have `require_secret` set. I don't know if we still need that behavior. I'll leave it for now, but we may want to revisit.

Callek: we can't test `has_commit_landed_on_repository` for a private repository. I don't have a need for this skippage anymore for guardian at all, but its worth knowing that there is no other way to get this information [that I could find] which would let us use a token.

It's essentially github-web-secret-api

escapewindow: https://github.com/mozilla-releng/scriptworker/issues/456 for followup.
I may need actual private repo releases for adhoc, so hopefully I'll have a better understanding in a bit.

escapewindow:
Thank you both!

PR #460, opened 2020-06-09
build_taskcluster_yml_url

escapewindow:


src/scriptworker/cot/verify.py
```diff
@@ -1281,6 +1281,29 @@ async def populate_jsone_context(chain, parent_link, decision_link, tasks_for):
...
+    repo_url = get_repo(link.task, source_env_prefix)
+    repo_url = repo_url.replace("git@github.com:", "ssh://github.com/", 1)
+    revision = get_revision(link.task, source_env_prefix)
+    repo_parts = urlparse(repo_url)
+    if repo_parts.netloc == "github.com":
+        user, repo_name = extract_github_repo_owner_and_name(repo_url)
+        path_prefix = f"/{user}/{repo_name}/raw"
+    else:
+        path_prefix = f"{repo_parts.path}/raw-file"
+    return f"{repo_parts.scheme}://{repo_parts.netloc}{path_prefix}/{revision}/.taskcluster.yml"
```
tomprince, line 1304: I think this would be better if the two branches didn't try to share code. While both hg.mozilla.org and github happen to have similiar URL structures[1], I don't think trying to reduce duplication makes things clearer. If one of these needed to change for any reason[2], it is unlikely that the other branch will need to change, which will make maintenance harder rather than easier.

---

[1] probably because it is a fairly natural URL strucutre
[2] perhaps so that github always uses raw.githubusercontent.com

src/scriptworker/cot/verify.py
```diff
@@ -1281,6 +1281,29 @@ async def populate_jsone_context(chain, parent_link, decision_link, tasks_for):
...
+    """
+    source_env_prefix = link.context.config["source_env_prefix"]
+    repo_url = get_repo(link.task, source_env_prefix)
+    repo_url = repo_url.replace("git@github.com:", "ssh://github.com/", 1)
+    revision = get_revision(link.task, source_env_prefix)
+    repo_parts = urlparse(repo_url)
+    if repo_parts.netloc == "github.com":
+        user, repo_name = extract_github_repo_owner_and_name(repo_url)
+        path_prefix = f"/{user}/{repo_name}/raw"
+    else:
```
tomprince, line 1302: I would be inclined to check explicitly for `hg.mozilla.org` here and fail otherwise

escapewindow:
Ok. I think I've addressed all review comments.
I think this PR addresses the issue of specifying the `.taskcluster.yml` url for `get_in_tree_template`. I noticed the following additional issues while testing locally:

- we still check the `metadata.source` url for `scriptworker.task.is_pull_request`. I started tackling that, but that will definitely make this PR much larger, as it changes some assumptions in `scriptworker.github`. My current wip makes the repo and revision optional per task. Since we look at all tasks in the chain to determine if the chain `is_try_or_pull_request`, and the decision task(s) must specify repo and revision, we can have the other tasks have a non-positive `is_pull_request` and still prevent the usage of restricted scopes for the chain.
- `test_build_task_dependencies` is slow... ~180s slow. I definitely want to address that soonish.

I'm open to either continuing to pull at these threads, or landing this set of changes and leaving the others for later PRs.

escapewindow:
#463 for the slow test fix. Looking at coverage ...

src/scriptworker/cot/verify.py
```diff
@@ -1281,6 +1280,32 @@ async def populate_jsone_context(chain, parent_link, decision_link, tasks_for):
...
+    source_env_prefix = link.context.config["source_env_prefix"]
+    repo_url = get_repo(link.task, source_env_prefix)
+    repo_url = repo_url.replace("git@github.com:", "ssh://github.com/", 1)
+    revision = get_revision(link.task, source_env_prefix)
+    repo_parts = urlparse(repo_url)
+    if repo_parts.netloc == "github.com":
+        user, repo_name = extract_github_repo_owner_and_name(repo_url)
+        url = f"https://raw.githubusercontent.com/{user}/{repo_name}/{revision}/.taskcluster.yml"
+    elif repo_parts.netloc == "hg.mozilla.org":
+        url = f"{repo_parts.scheme}://{repo_parts.netloc}{repo_parts.path}/raw-file/{revision}/.taskcluster.yml"
```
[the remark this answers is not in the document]
escapewindow: Once upon a time we did have private hg.m.o repos, which might use `ssh://`. I could also see us adding another hg server/hostname and change the check to `elif repo_parts.netloc in (...):`, as well. But yeah, I don't have strong opinions here. We could change it easily if our needs change.

escapewindow:
Thanks!

PR #461, opened 2020-06-09
Bug 1563169 - add different CoT product for app-services and add glea…

MihaiTabara:
While porting to taskgraph the https://github.com/mozilla/application-services/ we noticed we have a lot of inconsistency around naming (both the GCP workers, the clients, scopes, etc). Because of that reason, we're porting over from `application-services` to `app-services` everywhere. Once the migration is completed, we will retire the former.

While doing this, I'm tagging along the `glean` part too since we need to add that logic for https://github.com/mozilla/glean/ too in https://bugzilla.mozilla.org/show_bug.cgi?id=1644441

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -240,9 +240,7 @@ DEFAULT_CONFIG = immutabledict(
                         ),
                     ),
                     "glean": (
-                        immutabledict(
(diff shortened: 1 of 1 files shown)


MihaiTabara:
Once this is reviewed and landed I'll bump scriptworker in a separate PR like https://github.com/mozilla-releng/scriptworker/pull/445/files

PR #462, opened 2020-06-09
35.0.0

MihaiTabara:


HISTORY.rst
```diff
@@ -4,6 +4,14 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+[34.4.0] - 2020-06-09
```
escapewindow, line 7: Can we make this 35.0.0 and note that:

- the old docker-worker key is no longer valid
- production tests now require `TASKCLUSTER_PROXY_URL` to be set

in the changelog? (I want to fix the latter to allow for local production tests, but that'll come later.)
Thanks!

PR #463, opened 2020-06-09
fix slow test_build_task_dependencies

escapewindow:
This was taking ~180 seconds plus. Now it takes <1s.


escapewindow:
More closed event loop =\

PR #468, opened 2020-08-05
Upload a live.log redirect that points at the log or chain-of-trust log.

tomprince:


escapewindow:
Interesting. I think that live logging is likely not on the table, so it's a bit of a misnomer, but so is `live_backing.log`. This could resolve #179, if treeherder looked at `live.log` instead of `live_backing.log`.

tomprince:
Yeah, this was motivated by thinking that treeherder should look at `live.log` instead of `live_backing.log`. The fact that it resolves #179 is a nice bonus.

PR #474, opened 2020-11-13
Semaphore

escapewindow:
These are changes to make the chemspill fixes more permanent.
I'm not 100% sure the aiohttp max connections were working as designed. We use semaphores elsewhere (e.g. scriptworker-scripts) and they work very well.

### DIFF
--- a/src/scriptworker/utils.py
+++ b/src/scriptworker/utils.py
@@ -228,6 +228,7 @@ async def retry_async(
     args: Sequence[Any] = (),
     kwargs: Optional[Dict[str, Any]] = None,
     sleeptime_kwargs: Optional[Dict[str, Any]] = None,
+    log_exceptions: Optional[bool] = False,
 ) -> Any:
     """Retry ``func``, where ``func`` is an awaitable.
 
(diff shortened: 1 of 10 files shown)


src/scriptworker/utils.py
```diff
@@ -258,7 +259,9 @@ async def retry_async(
     while True:
         try:
             return await func(*args, **kwargs)
-        except retry_exceptions:
+        except retry_exceptions as exc:
+            if log_exceptions:
+                log.info(f"retry_async exception:\n{type(exc)} {exc}")
```
JohanLorenzo, line 264: Nit: Should we `log.warning()` instead? I get the feeling an exception is worth being called out as a warning.

tests/test_context.py
```diff
@@ -149,3 +149,12 @@ def test_bad_verify_task(claim_task, bad_path):
...
     with pytest.raises(CoTError):
         context.verify_task()
+
+
+@pytest.mark.asyncio
+async def test_download_semaphore():
+    context = swcontext.Context()
+    sem = context.download_semaphore
+    assert type(sem) == asyncio.BoundedSemaphore
+    assert sem._value == swcontext.DEFAULT_MAX_CONCURRENT_DOWNLOADS
```
JohanLorenzo, line 159: 👍

JohanLorenzo (approved):
LGTM!

src/scriptworker/utils.py
escapewindow: Hm. Today, we retry on exception, and we don't log anything if we're retrying; we silently eat the error and sleep, so just logging anything is an improvement.

I don't have a strong objection to warning though, so I can try that.

PR #476, opened 2020-12-11
restrict github scopes

escapewindow:


JohanLorenzo (approved):
Good catch! 

I think we'll have to rename these 2 products: 

 * https://github.com/mozilla-releng/scriptworker-scripts/blob/21615ebb86153487490edb93ac5646f7c4f27b45/githubscript/docker.d/worker.yml#L34 
 * https://github.com/mozilla-releng/scriptworker-scripts/blob/21615ebb86153487490edb93ac5646f7c4f27b45/githubscript/docker.d/worker.yml#L23

Otherwise, CoT will bail on us.

escapewindow:
That's https://github.com/mozilla-releng/scriptworker-scripts/pull/310 :)

PR #481, opened 2021-02-05
add taskQueueId to the ignore list

escapewindow:
Should address https://github.com/taskcluster/taskcluster/pull/4277 .
Once taskQueueId is everywhere on all tasks that we want to verify, we can remove it from the ignore list.


jmaher (approved):
just reading the code briefly and what taskQueueId is, this seems unrelated to really validating a copy.

PR #484, opened 2021-02-11
[RELENG-42] add new scriptworker, docker-worker, and generic-worker cot keys

escapewindow:
The docker-worker keypair isn't 100% guaranteed to stay the same, but let's roll this out in case it is. At least we'll mark the scriptworker+generic-worker keys valid.


jmaher (approved):
sort of a rubber stamp here, not sure how to verify these keys.

escapewindow:
Thanks!

PR #486, opened 2021-02-19
[RELENG-348] remove esr68, add pine as level 3

escapewindow:


src/scriptworker/constants.py
```diff
@@ -457,39 +451,35 @@
...
                             # Limit things like pushapk to just these branches
                             "release": ("/releases/mozilla-release",),
                             "beta": ("/releases/mozilla-beta",),
                             "beta-or-release": ("/releases/mozilla-beta", "/releases/mozilla-release"),
-                            "esr": (
-                                "/releases/mozilla-esr68",
-                                "/releases/mozilla-esr78",
-                            ),
-                            "esr68": ("/releases/mozilla-esr68",),
+                            "esr": ("/releases/mozilla-esr78",),
```
jmaher, line 461: do we also need "esr78" ?

jmaher (approved):
looks good, maybe 1 nit?

src/scriptworker/constants.py
escapewindow: I think we only needed esr68 for fennec-specific things.
Good eye, though.

PR #490, opened 2021-03-22
Add typing to context.py

Callek:
Took a stab at adding some more typing to scriptworker.

### DIFF
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -232,7 +232,7 @@ class Context(object):
     def event_loop(self, event_loop: asyncio.AbstractEventLoop) -> None:
         self._event_loop = event_loop
 
-    async def populate_projects(self, force: bool=False) -> None:
+    async def populate_projects(self, force: bool = False) -> None:
         """Download the ``projects.yml`` file and populate ``self.projects``.
 
         This only sets it once, unless ``force`` is set.
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -22,9 +22,9 @@ import aiohttp
 import arrow
 from taskcluster.aio import Queue
 
+from scriptworker import task_process
 from scriptworker.exceptions import CoTError
 from scriptworker.utils import load_json_or_yaml_from_url, makedirs
-from scriptworker import task_process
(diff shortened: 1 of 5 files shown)


src/scriptworker/utils.py
```diff
@@ -25,6 +25,7 @@
 import yaml
 from taskcluster.client import createTemporaryCredentials
 
+from scriptworker.context import Context
```
escapewindow, line 28: This causes a circular import.

Callek: @escapewindow circular import is now fixed, I opted for a "Don't refactor" solution

src/scriptworker/context.py
```diff
@@ -231,12 +242,14 @@ async def populate_projects(self, force=False):
...
         """
+        assert self.config
         if force or not self.projects:
             with tempfile.TemporaryDirectory() as tmpdirname:
                 self.projects = await load_json_or_yaml_from_url(self, self.config["project_configuration_url"], os.path.join(tmpdirname, "projects.yml"))
 
     @property
-    def download_semaphore(self):
+    def download_semaphore(self) -> asyncio.BoundedSemaphore:
+        assert self.config
```
escapewindow, line 252: This causes the following test bustage:

```
________________________________ test_download_semaphore ________________________________

    @pytest.mark.asyncio
    async def test_download_semaphore():
        context = swcontext.Context()
>       sem = context.download_semaphore

tests/test_context.py:157:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <scriptworker.context.Context object at 0x110227460>

    @property
    def download_semaphore(self) -> asyncio.BoundedSemaphore:
>       assert self.config
E       AssertionError

src/scriptworker/context.py:252: AssertionError
```

Fairly easy fix of setting `context.config` to a non-empty dict in that test.

escapewindow (changes requested):

escapewindow (approved):
Thank you!

PR #491, opened 2021-04-14
Docker worker cot

escapewindow:
We never used the previous-new docker-worker CoT key. We're rolling out a new docker-worker CoT key; let's allowlist the pubkey.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,13 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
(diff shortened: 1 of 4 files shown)


bhearsum (approved):

escapewindow:
Mypy is broken. @Callek I'm guessing this is due to a new mypy version? Tempted to turn off mypy tests for now.

escapewindow:
No, same mypy, different immutabledict. I'll force squash and merge, and look to disable mypy tests if this continues being a problem.

Callek:
I have a quick fix here I can give rather than disable mypy

PR #502, opened 2021-05-19
Adjust xpi-manifest cot-scopes

jmaher:


src/scriptworker/cot/verify.py
```diff
@@ -1907,7 +1923,7 @@ async def trace_back_to_tree(chain):
     # check for restricted scopes.
     my_repo = repos[chain]
     for scope in chain.task["scopes"]:
-        if scope in rules["scopes"]:
+        if chain.is_scope_in_restricted_scopes(scope, rules["scopes"]):
             log.info("Found privileged scope {}".format(scope))
             level = rules["scopes"][scope]
```
escapewindow, line 1928: Hm. I think this line is going to break, since `scope` may not be the key; it may be a glob.
We may need to change `is_scope_in_restricted_scopes` to return the matching rule, which means we need to update the test as well.

escapewindow (changes requested):
Looks good, thank you!
Looks like the code requires a slightly different behavior than we initially thought. Hoping that's fairly straightforward to fix.

src/scriptworker/cot/verify.py
```diff
@@ -164,14 +164,32 @@ def is_decision(self):
...
+           if one of the restricted scopes ends with '*', find a partial match.
+
+        Returns:
+            String: string of matching restricted_scope, if no match ""
+        """
+        match = [""]
+        for restricted in restricted_scopes:
+            if restricted.endswith("*"):
+                if scope.startswith(restricted.rstrip("*")):
+                    match.append(restricted)
```
escapewindow, line 178: I could see us just returning `restricted` in each of these, rather than building a list. I don't have a strong opinion here; in theory, we shouldn't have overlapping restricted scopes.

Alternately, matching multiple restricted scopes may be a reason to throw an error.

Do you have thoughts here?

src/scriptworker/cot/verify.py
```diff
@@ -1907,9 +1925,10 @@ async def trace_back_to_tree(chain):
     # check for restricted scopes.
     my_repo = repos[chain]
     for scope in chain.task["scopes"]:
-        if scope in rules["scopes"]:
+        matched_scope = chain.is_scope_in_restricted_scopes(scope, rules["scopes"])
+        if matched_scope:
             log.info("Found privileged scope {}".format(scope))
-            level = rules["scopes"][scope]
+            level = rules["scopes"][matched_scope]
```
escapewindow, line 1931: 👍

tests/test_cot_verify.py
```diff
@@ -2193,3 +2193,31 @@ async def test_get_scm_level(rw_context, project, level, raises):
...
+
+
+# tests for matching scopes with a partial match, implemented for xpi
+@pytest.mark.parametrize(
+    "scope, restricted_scopes, expected",
+    (
+        (
+            "project:xpi:signing:cert:release-signing",
+            ["project:xpi:ship-it:production", "project:xpi:signing:cert:release-signing"],
+            "project:xpi:signing:cert:release-signing",
```
escapewindow, line 2205: I like using the real scope strings, ty.

src/scriptworker/cot/verify.py
```diff
@@ -164,14 +164,35 @@ def is_decision(self):
...
+        """
+        match = [""]
+        for restricted in restricted_scopes:
+            if restricted.endswith("*"):
+                if scope.startswith(restricted.rstrip("*")):
+                    match.append(restricted)
+            elif scope == restricted:
+                match.append(restricted)
+
+        if len(match) > 2:
```
escapewindow, line 182: `> 1` or `>= 2` ?

src/scriptworker/version.py
```diff
@@ -54,7 +54,7 @@ def get_version_string(version: Union[ShortVerType, LongVerType]) -> str:
 
 # 1}}}
 # Semantic versioning 2.0.0  http://semver.org/
-__version__ = (38, 0, 0)
+__version__ = (38, 1, 0)
```
escapewindow, line 57: Can you also update HISTORY.rst with the change?

escapewindow (approved):
Thank you!

Issue #509, opened 2021-08-04
unit tests busted on new pip dependency resolver


escapewindow:
https://github.com/pypa/pip/issues/9168
https://discuss.python.org/t/announcement-pip-20-3-release/5948
https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020

escapewindow:
Fixed by #508.

PR #512, opened 2021-08-23
Replace mpd001 trust domain with mozillavpn

bhearsum:
AFAICT there's no reason to continue supporting the `mpd001` trust domain.

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -116,7 +116,7 @@ DEFAULT_CONFIG: immutabledict[str, Any] = immutabledict(
                         "firefox": "hg",
                         "thunderbird": "hg",
                         "mobile": "github",
-                        "mpd001": "github",
+                        "mozillavpn": "github",
                         "app-services": "github",
                         "glean": "github",
                         "xpi": "github",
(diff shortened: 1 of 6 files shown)


escapewindow:
Timeout bustage appears to be #509 ; merging master in or rebasing the PR should resolve.

src/scriptworker/constants.py
```diff
@@ -512,7 +512,7 @@
                     "firefox": "GECKO",
                     "thunderbird": "COMM",
                     "mobile": "MOBILE",
-                    "mpd001": "MPD001",
+                    "mozillavpn": "MOZILLAVPN",
```
escapewindow, line 515: This means we're going to look for `MOZILLAVPN_HEAD_REV` etc, which will be relevant in your `.tc.yml`

tests/conftest.py
```diff
@@ -154,10 +154,10 @@ async def mobile_rw_context(event_loop):
 
 @pytest.mark.asyncio
 @pytest.yield_fixture(scope="function")
-async def mpd_private_rw_context(event_loop):
+async def vpn_private_rw_context(event_loop):
```
escapewindow, line 157: This is specifically for testing private repos, but I think it may be useful for later private repo usage. We can keep it, esp since your test data file below starts with `git@github` rather than `https://github`.

escapewindow (approved):
Thanks, looks good!

PR #514, opened 2021-08-24
Scriptworker 40.0.0

bhearsum:
In reality, bumping any part of the version should be fine, but this seems most correct?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,13 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+[40.0.0] - 2021-08-24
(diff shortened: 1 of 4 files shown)


escapewindow (approved):
Thanks!
Once we push this to pypi, we roll out via 1) scriptworker-scripts pin.sh + associated `production*` branches, and 2) ronin_puppet [scriptworker_revision](https://github.com/mozilla-platform-ops/ronin_puppet/blob/7c3c12e7d4e554ca0286c2e9d3da26d60926ea6c/data/common.yaml#L124-L141) and `production-mac-signing` branch.

I'll want to roll out to all scriptworkers to remove the old CoT keys. I'm happy to help with rollout, especially if you only need a subset of scriptworkers updated.

PR #516, opened 2021-09-01
Drop non-hook action support in CoT

ahal:
This bumps CoT to version 7.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,16 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+[41.0.0] - 2021-09-02
+---------------------
+Removed
+~~~~~~~
+- Support for non-hook actions in CoT
+
+Changed
+~~~~~~~
+- CoT bumped to version 7
+
 [40.0.0] - 2021-08-24
 ---------------------
 Changed
--- a/docs/cot_overview.rst
+++ b/docs/cot_overview.rst
@@ -29,13 +29,10 @@ In conjunction with other best practices, like `separation of roles <https://en.
 Chain of Trust Versions
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-1. Initial Chain of Trust implementation with GPG signatures: Initial `1.0.0b1 on 2016-11-14 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#100b1---2016-11-14>`_
-2. CoT v2: rebuild task definitions via json-e. `7.0.0 on 2018-01-18 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#700---2018-01-18>`_
-3. Generic action hook support. `12.0.0 on 2018-05-29 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#1200---2018-05-29>`_
-4. Release promotion action hook support. `17.1.0 on 2018-12-28 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#1710---2018-12-28>`_
-5. ed25519 support; deprecate GPG support. `22.0.0 on 2019-03-07 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#2200---2019-03-07>`_
-6. drop support for gpg `23.0.0 on 2019-03-27 <https://github.com/mozilla-releng/scriptworker/blob/master/CHANGELOG.md#2300---2019-03-27>`_
-
-Planned future versions:
-
-* drop support for non-hook actions
+1. Initial Chain of Trust implementation with GPG signatures: Initial `1.0.0b1 on 2016-11-14 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#100b1---2016-11-14>`_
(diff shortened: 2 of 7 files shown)


HISTORY.rst
```diff
@@ -4,6 +4,16 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+[41.0.0] - 2021-09-01
```
escapewindow, line 7: We can update this date to when we actually create the release. (I'm not sure how accurate this needs to be, but I've tended to try to get it accurate.)

docs/cot_overview.rst
```diff
@@ -29,13 +29,10 @@ In conjunction with other best practices, like `separation of roles <https://en.
...
-Planned future versions:
-
-* drop support for non-hook actions
+1. Initial Chain of Trust implementation with GPG signatures: Initial `1.0.0b1 on 2016-11-14 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#100b1---2016-11-14>`_
+2. CoT v2: rebuild task definitions via json-e. `7.0.0 on 2018-01-18 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#700---2018-01-18>`_
+3. Generic action hook support. `12.0.0 on 2018-05-29 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#1200---2018-05-29>`_
+4. Release promotion action hook support. `17.1.0 on 2018-12-28 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#1710---2018-12-28>`_
+5. ed25519 support; deprecate GPG support. `22.0.0 on 2019-03-07 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#2200---2019-03-07>`_
+6. drop support for gpg `23.0.0 on 2019-03-27 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#2300---2019-03-27>`_
+7. drop support for non-hook actions `41.0.0 on 2021-09-01 <https://github.com/mozilla-releng/scriptworker/blob/master/HISTORY.rst#2300---2019-03-27>`_
```
escapewindow, line 38: Thanks for updating these links!

src/scriptworker/cot/verify.py
```diff
@@ -1375,14 +1374,11 @@ async def get_in_tree_template(link):
...
 def _get_action_from_actions_json(all_actions, callback_name):
     for defn in all_actions:
-        if defn.get("kind") == "hook":
-            if defn.get("hookPayload", {}).get("decision", {}).get("action", {}).get("cb_name") == callback_name:
-                return defn
-        elif defn.get("kind") == "task":
-            if defn.get("task", {}).get("$let", {}).get("action", {}).get("cb_name") == callback_name:
-                return defn
-        else:
+        if defn.get("kind") != "hook":
```
escapewindow, line 1377: I wonder if we should move this under the `if` below.

As currently written, it'll bail if we find a non-hook action at all, even if it doesn't match the callback name. This may not be problematic now, but if we want to introduce a new action kind on an experimental basis, it may potentially break all action CoT verification until we update scriptworker. (We had this issue before you touched this code, though, so up to you if you want to make that fix now.)

tests/test_cot_verify.py
```diff
@@ -292,18 +293,18 @@ def get_cot(task_defn, task_id="task_id"):
...
     if path.endswith("taskcluster.yml"):
         return load_json_or_yaml(os.path.join(parent_path, ".taskcluster.yml"), is_path=True, file_type="yaml")
     elif path.endswith("projects.yml"):
         return load_json_or_yaml(os.path.join(parent_path, "projects.yml"), is_path=True, file_type="yaml")
 
 
-async def cotv4_load_url(context, url, path, **kwargs):
-    return await cotv2_load_url(context, url, path, parent_path=COTV4_DIR, **kwargs)
+cotv2_load_url = partial(cot_load_url, parent_path=COTV2_DIR)
+cotv4_load_url = partial(cot_load_url, parent_path=COTV4_DIR)
```
escapewindow, line 304: Smart.

escapewindow (approved):
I don't see anything blocking here. We can update the date if wanted, and the if block if wanted, not required.

LMK if you want to handle the [release](https://scriptworker.readthedocs.io/en/latest/releases.html) or if you want me to.

escapewindow:
(Thanks for handling this! Do you feel like you have a better grasp of CoT? Do you want more, or is that enough for now?)

ahal:
> (Thanks for handling this! Do you feel like you have a better grasp of CoT? Do you want more, or is that enough for now?)

I definitely have a better grasp on it! Digging into the tests was a good way to see how the various artifacts in dependencies along the chain get used to validate everything. I think I'm ok for the immediate short term (and I have some fission stuff I should probably tackle first anyway), but I'll create a new CoT cross-training ticket to update the test data with modern task definitions and we can prioritize that (or not) through the normal process.

If we wanted to ramp up someone else on CoT, I could even attempt mentoring it (which might result in the two of us digging into code together)

ahal:
I'll fix the issues, and for some reason a test started failing for me locally which didn't before (and which doesn't on this PR).. strange.

src/scriptworker/cot/verify.py
ahal: Fixed

ahal:
> for some reason a test started failing for me locally which didn't before (and which doesn't on this PR).. strange.

nvm, this was just caused by me fixing one of the review suggestions.. solved now.

PR #522, opened 2021-11-24
Update mobile cot restricted scopes for focus-android

sarah-clements:


sarah-clements:
I think I fixed the test failure Julien but I'm still seeing the taskcluster error :/

README.rst
```diff
@@ -58,20 +56,15 @@ With integration tests, first create a client with the scopes::
...
     }
 
 
-(``certificate`` is only specified if using temp creds)
-
-
-then
-
-``tox``
+then to run all tests: ``tox``
```
escapewindow, line 70: Thanks for this cleanup!

src/scriptworker/constants.py
```diff
@@ -369,10 +369,13 @@
...
                             "project:mobile:firefox-tv:releng:signing:cert:production-signing": "firefox-tv-repo",
-                            "project:mobile:focus:googleplay:product:focus": "focus-repo",
-                            "project:mobile:focus:releng:signing:cert:release-signing": "focus-repo",
                             "project:mobile:reference-browser:releng:googleplay:product:reference-browser": "reference-browser-repo",
                             "project:mobile:reference-browser:releng:signing:cert:release-signing": "reference-browser-repo",
+                            "project:mobile:focus-android:releng:github:project:focus-android": "focus-repo",
+                            "project:mobile:focus-android:releng:googleplay:product:focus-android": "focus-repo",
+                            "project:mobile:focus-android:releng:signing:format:autograph_focus": "focus-repo",
+                            # beta and nightly are signed with same key as production
+                            "project:mobile:focus-android:releng:signing:cert:production-signing": "focus-repo",
```
escapewindow, line 378: It looks like we also have the `release-signing` scope enabled in signingscript [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/d60deec5a61255f6bd21282b2a694e26b350f3ae/signingscript/docker.d/passwords.yml#L270-L289). So while the `production-signing` scope will be restricted after this rolls out, someone could still sign with a valid key with the unrestricted `release-signing` scope.

Let's:
- determine which old branches we still support, and which signing cert scopes we still support,
- remove the cert scopes we aren't using from ci-config and signingscript, and
- make sure that any remaining non-dep cert scopes we still support are listed here.

I suspect that means we can remove `focus-android:...:release-signing` from signingscript, and that `focus:...:release-signing` is removable now or soonish, and the above restricted scope is sufficient for signing cert scopes.

I'm hoping all of my comments here are clear; let me know if you have questions?

escapewindow (changes requested):
Thank you for opening this PR! I think you've learned about how scriptworker tests and restricted scopes work. There are some changes needed. If we get this PR fixed up and ready to ship before I'm back from Thanksgiving, someone with pypi access can follow [these steps](https://scriptworker.readthedocs.io/en/latest/releases.html) after merging the PR.

I suspect the error is something to do with this PR coming from an in-repo branch as opposed to a fork, but I'm not 100% sure. I'm going to poke at it.

escapewindow:
> I think I fixed the test failure but I'm still seeing the taskcluster error :/

You're a collaborator of this repo, as part of the mozilla-releng/releng team.

I suspect this is a combination of:
- add-focus being a branch off the main repo, rather than a fork,
- the scopes being granted to either `pull-request` or `branch:master` (not `branch:add-focus`), and
- this repo [not yet being ported to taskgraph](https://github.com/mozilla-releng/scriptworker/blob/master/.taskcluster.yml). the `tasks_for` logic is pretty weak.

Still guessing: the revision is pushed to the main repo (different branch), we don't have the scopes to trigger the task non a non-pr non-master revision, and then opening a PR from that branch isn't retriggering an event since it's the same repo.

Potential short-term fixes: allow those secrets to be readable from `branch:*`, or submit the PR from a fork. Long-term fix: let's port scriptworker to taskgraph when we have the cycles. I suspect that won't be in H1, though, unless someone wants the practice.

src/scriptworker/constants.py
```diff
@@ -369,10 +369,13 @@
                             "project:mobile:fennec-profile-manager:releng:signing:cert:fennec-production-signing": "fennec-profile-manager-repo",
                             "project:mobile:firefox-tv:releng:googleplay:product:firefox-tv": "firefox-tv-repo",
                             "project:mobile:firefox-tv:releng:signing:cert:production-signing": "firefox-tv-repo",
-                            "project:mobile:focus:googleplay:product:focus": "focus-repo",
-                            "project:mobile:focus:releng:signing:cert:release-signing": "focus-repo",
```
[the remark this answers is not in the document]
sarah-clements: I can see in the focus-android code base that only this is used: `'project:mobile:focus-android:releng:signing:cert:production-signing` (or :dep-signing). Same goes for the `focus:googleplay` scope. I'll clean those up in the files.

src/scriptworker/constants.py
sarah-clements: As I understand it, anything pre-taskgraph used the `focus` scopes, not the new `focus-android` scopes so I'm confident those are safe to remove.

PR #525, opened 2021-11-30
Add scope constants for microsoftstore / pushmsixscript.

gbrownmozilla:
Follow-up on a review comment from the pushmsixscript PR.

These scopes match those in pushmsixscript. Now that I see them in this list, I wonder, should they contain ":firefox"?

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -337,6 +337,8 @@ DEFAULT_CONFIG: immutabledict[str, Any] = immutabledict(
                             "project:releng:flathub:firefox:stable": "release",
                             "project:releng:ship-it:production": "all-production-branches",
                             "project:releng:treescript:action:push": "all-production-branches",
+                            "project:releng:microsoftstore:beta": "beta",
+                            "project:releng:microsoftstore:release": "release",
                         }
                     ),
                     "thunderbird": immutabledict(


escapewindow (approved):
> These scopes match those in pushmsixscript. Now that I see them in this list, I wonder, should they contain ":firefox"?

I give you a solid "maybe".

It makes sense if we're planning on submitting other things to the microsoft store, and we want to limit what can be pushed there.

If we're pushing, say, thunderbird or mobile or something to the store, then that will be a separate trust domain and a separate set of pools, so splitting out `:firefox` may or may not help, especially if we have a different set of store creds per pool. We could add the trust domain here, e.g. `:gecko`, if that's what we're guarding against.

If we're planning on pushing multiple products from the same trust domain, and we're releasing, say, devedition from beta, but on the release track, then absolutely this makes sense.

Then again, if we don't know of any such requirement, we can:

a) roll out as-is
b) find out that there is another product we want to ship off a different set of branches in the same trust domain,
c) deprecate the old scopes,
d) add a new set of scopes that is more granular, add support for the new scopes in the pools, restrict the appropriate scopes, and roll out, then
e) when the new scopes have ridden the trains or have been uplifted, remove the deprecated scopes.

So if we know of other uses of this worker coming shortly, we may want to adjust now. If we may someday need more uses, we can adjust the scopes as needed when we know more about the requirements. Long story short: I think we're probably fine for now unless there's something coming I'm not aware of.

gbrownmozilla:
I appreciate the explanation -- helps to confirm my understanding.

afaik, no other uses of the worker are planned, so let's go ahead like this.

PR #528, opened 2022-02-09
Bug 1613283 - Add log rotation option for taskcluster.yaml

hneiva:


src/scriptworker/log.py
```diff
@@ -57,6 +57,12 @@ def update_logging_config(context: Any, log_name: Optional[str] = None, file_nam
         # If we rotate the log file via logrotate.d, let's watch the file
         # so we can automatically close/reopen on move.
         handler = logging.handlers.WatchedFileHandler(path)  # type: ignore
+    elif context.config.get("log_max_bytes", 0) or context.config.get("log_max_backups", 0):
```
escapewindow, line 60: maybe `and` instead of `or` ?

src/scriptworker/log.py
```diff
@@ -57,6 +57,12 @@ def update_logging_config(context: Any, log_name: Optional[str] = None, file_nam
...
+        handler = logging.handlers.RotatingFileHandler(  # type: ignore
+            filename=path,
+            maxBytes=context.config.get("log_max_bytes", 0),
+            backupCount=context.config.get("log_max_backups", 0),
```
escapewindow, line 64: at which point we don't need the `get`s here.

escapewindow (approved):
Thank you!
I do prefer if we make the changes suggested.

src/scriptworker/log.py
hneiva: Mmmm I think we should go by `log_max_bytes` only then.
Wouldn't make sense to have unlimited max_bytes but max_backups > 0

src/scriptworker/log.py
hneiva: Right :)

src/scriptworker/log.py
escapewindow: Does this mean you want to allow for max_backups == 0 but we blow away the log every log_max_bytes?

escapewindow: If so, I'd lean towards the default of max_backups to be, say, 10 in constants.py.

PR #530, opened 2022-03-11
RELENG-847 add mozilla vpn apk release scope to restricted scopes

gabrielBusta:
RELENG-847 add mozilla vpn apk release scope to restricted scopes


escapewindow:
The failure is probably because we're opening a PR from the base repo rather than a fork.

escapewindow (approved):

PR #532, opened 2022-04-06
unpin pytest-asyncio

escapewindow:
I pinned pytest-asyncio to deal with https://github.com/pytest-dev/pytest-asyncio/issues/209; unpinning to see if we're still broken with recent releases.


escapewindow:
Still getting `event loop is closed`, same as https://github.com/pytest-dev/pytest-asyncio/issues/209#issuecomment-822901644.

I'm guessing we're going to have to fix the tests on our side to work with the new pytest-asyncio.

PR #535, opened 2022-04-20
Added changes for VPN Mac signing

hneiva:


src/scriptworker/constants.py
```diff
@@ -504,9 +505,9 @@
                 {
                     "firefox": ("decision", "action", "docker-image"),
                     "thunderbird": ("decision", "action", "docker-image"),
+                    "mozillavpn": ("decision", "action", "docker-image"),
```
hneiva, line 508: @escapewindow BTW, I'm not 100% sure this is what we want. Could you please confirm.

escapewindow: This is what we want. We're already pinned to a dockerhub decision image, so we should be good.
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/64d6b0147c698c8baba189b107fc09fc2ee011df/.taskcluster.yml#L241

src/scriptworker/constants.py
```diff
@@ -67,6 +67,7 @@
         "task_log_dir": "...",  # set this to ARTIFACT_DIR/public/logs
         "artifact_upload_timeout": 60 * 20,
         "max_concurrent_downloads": 5,
+        "provisionprofile_dir": "...",  # For apple signing
```
escapewindow, line 70: This looks like an iscript script_config.yml setting, not a scriptworker worker daemon config setting, no?

escapewindow (changes requested):

ahal:
Looks fine to me, but I'm not familiar with this code at all, so going to defer to Aki here.

hneiva:
Changed how `provisionprofile_dir` is loaded. Not needed anymore.

PR #537, opened 2022-04-20
add codecov

escapewindow:


escapewindow:
Looks like we need a token, same as coveralls. Need to figure that out -- maybe a helper script, or have a tox target that hits the tc proxy?

ahal:
In taskgraph I went with a helper script:
https://github.com/taskcluster/taskgraph/blob/main/taskcluster/scripts/codecov-upload.py

PR #539, opened 2022-05-09
Bug 1766928 - Add mozilla-esr102 to CoT

gabrielBusta:
Should we remove esr78?


escapewindow (approved):
> Should we remove esr78?

Yes please!

PR #543, opened 2022-06-13
CoT verify: Enhance error message to show the root cause of download …

JohanLorenzo:
…errors may be in the upstream task

That's a quick change while https://github.com/taskcluster/taskgraph/issues/47 gets implemented and rolled out. The current error message created confusion again in https://bugzilla.mozilla.org/show_bug.cgi?id=1773274.


hneiva (approved):

src/scriptworker/cot/verify.py
```diff
@@ -713,7 +713,13 @@ async def download_cot_artifact(chain, task_id, path):
...
 
     if path not in link.cot["artifacts"]:
-        raise CoTError("path {} not in {} {} chain of trust artifacts!".format(path, link.name, link.task_id))
+        raise CoTError(
+            "path {path} not in {link_name} {task_id} chain of trust artifacts! This is likely "
+            "a bug in {task_id}: docker-worker does not fail if a declared artifact was never "
+            "generated by the task. Please download and make sure the artifact {path} is the "
+            "one you expect. If it is, then please reach out to the Release Engineering "
+            "team. For more information: https://github.com/taskcluster/taskgraph/issues/47".format(path=path, link_name=link.name, task_id=link.task_id)
+        )
```
escapewindow, line 722: I think this is generally the case, and this would also point them at the upstream artifact if we haven't configured a branch correctly for win32 stub installers. 👍 

We probably want to update this message if and when we either fix the docker-worker bug, the run-task bug, or migrate all docker-worker tasks to generic-worker.

JohanLorenzo: I agree, this message won't be needed anymore once the root cause is fixed.

PR #544, opened 2022-06-21
clean up old project branches

escapewindow:


src/scriptworker/constants.py
```diff
@@ -193,7 +193,7 @@
                                     r"^(?P<path>/mozilla-(central|unified))(/|$)",
                                     r"^(?P<path>/integration/(autoland|fx-team|mozilla-inbound))(/|$)",
                                     r"^(?P<path>/releases/mozilla-(beta|release|esr\d+))(/|$)",
-                                    r"^(?P<path>/projects/(maple|oak|cedar|pine))(/|$)",
```
JohanLorenzo, line 196: Question: I see `cedar` was last updated 3 months ago[1]. Although `oak` was last modified 9 months ago.  


Screenshot 2022-06-22 at 10 10 05

Do we want to get rid of `oak` as well?


[1] https://hg.mozilla.org/projects?sort=lastchange

JohanLorenzo (approved):
LGTM modulo a quick question!

src/scriptworker/constants.py
jcristau: https://wiki.mozilla.org/ReleaseEngineering/DisposableProjectRepositories#BOOKING_SCHEDULE has oak booked by the install-update team, might be worth checking with them.

JohanLorenzo: Oh TIL about this page 🙂 Yeah, we may want to check with the rest of the engineering org if can close a few of these branches, then.

PR #547, opened 2022-07-27
RELENG-830 - Add 2022 docker-worker trusted CoT public key

ahal:


ahal:
Moving this to https://github.com/mozilla-releng/scriptworker/pull/549

PR #548, opened 2022-07-27
Add beetmover:bucket:release to mozillavpn cot_restricted_scopes

gabrielBusta:
Add beetmover:bucket:release to mozillavpn cot_restricted_scopes


hneiva (approved):
I wonder if the emoji in the commit message will cause issues one day...

PR #550, opened 2022-07-27
fix cot test

escapewindow:
I'm guessing the old one hasn't been released in over a year.
I went to shipit to find recent xpi releases and chose the most recent.
Adhoc I looked at the index list and chose something 1 month old I think.

We could potentially mitigate this in the future by adding a generic "latest-release-signing-task-of-any-variety" index to both xpi and adhoc, and pointing at that.


ahal (approved):
Thanks!

PR #551, opened 2022-08-12
Replace 2022 docker-worker CoT public key

ahal:


ahal:
Explanation in commit message:
https://github.com/mozilla-releng/scriptworker/pull/551/commits/82e8d7f4e39de06d522f8dd6cafa92643d999eec

escapewindow (approved):
The longer key format is likely the private and public keys concatenated. Perhaps the node library has to have it in that format?

ahal:
> The longer key format is likely the private and public keys concatenated. Perhaps the node library has to have it in that format?

Oh good call, that looks like what it is!

I'm going to close this then as Dave can just append the public key to the one I had previously generated manually.

PR #552, opened 2022-08-15
Allow autoland branch to use treescript push

gabrielBusta:
This patch allows autoland branch to push to trees


escapewindow (changes requested):
Looks good, *but* I neglected to mention we need to add integration/autoland to the [`path_regexes`](https://github.com/mozilla-releng/scriptworker/blob/37f35ab39d84bffa162108a272a4ec8f449dc863/src/scriptworker/constants.py#L194-L207)

escapewindow:
> Looks good, _but_ I neglected to mention we need to add integration/autoland to the [`path_regexes`](https://github.com/mozilla-releng/scriptworker/blob/37f35ab39d84bffa162108a272a4ec8f449dc863/src/scriptworker/constants.py#L194-L207)

Oh wait, it's already there?

escapewindow (approved):
I missed that we already check for integration/autoland :)

PR #554, opened 2022-08-19
Bug 1784232 - Let CoT support BASE_REF and BASE_REV

JohanLorenzo:
Depends on https://phabricator.services.mozilla.com/D155083

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -110,7 +110,7 @@ DEFAULT_CONFIG: immutabledict[str, Any] = immutabledict(
             }
         ),
         "project_configuration_url": "https://hg.mozilla.org/ci/ci-configuration/raw-file/default/projects.yml",
-        "pushlog_url": "{repo}/json-pushes?changeset={revision}&tipsonly=1&version=2&full=1",
+        "pushlog_url": "{repo}/json-pushes?changeset={revision}&version=2&full=1",
         "chain_of_trust_hash_algorithm": "sha256",
         "cot_schema_path": os.path.join(os.path.dirname(__file__), "data", "cot_v1_schema.json"),
         # for download url validation.  The regexes need to define a 'filepath'.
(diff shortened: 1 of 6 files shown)


escapewindow (approved):
I was concerned with backwards compatibility, but all the cot tests passed. Thanks!

src/scriptworker/constants.py
```diff
@@ -110,7 +110,7 @@
             }
         ),
         "project_configuration_url": "https://hg.mozilla.org/ci/ci-configuration/raw-file/default/projects.yml",
-        "pushlog_url": "{repo}/json-pushes?changeset={revision}&tipsonly=1&version=2&full=1",
+        "pushlog_url": "{repo}/json-pushes?changeset={revision}&version=2&full=1",
```
jcristau, line 113: I'm curious about this, I couldn't find where we need the other revisions.

src/scriptworker/cot/verify.py
```diff
@@ -1037,8 +1039,16 @@ async def _get_additional_hg_action_jsone_context(parent_link, decision_link):
 async def _get_additional_hgpush_info(decision_link):
     pushlog_info = await get_pushlog_info(decision_link)
     pushlog_id = list(pushlog_info["pushes"].keys())[0]
-    push_comment = pushlog_info["pushes"][pushlog_id]["changesets"][0]["desc"]
-    return {"id": pushlog_id, "comment": push_comment, "date": pushlog_info["pushes"][pushlog_id]["date"], "user": pushlog_info["pushes"][pushlog_id]["user"]}
+    first_revision = pushlog_info["pushes"][pushlog_id]["changesets"][0]
+    push_comment = first_revision["desc"]
```
jcristau, line 1043: Shouldn't this be the tip commit's description, to preserve the previous behavior?

JohanLorenzo: Oh yeah, definitely! Thanks for pointing that out. Fixed!

PR #556, opened 2022-08-24
Bug 1786838 - Allow comm-central to use production bouncer scope.

jfx2006:
The scope is used by the "bouncer-locations" on comm-central to
keep the Thunderbird Daily bouncer products pointing to the
latest version.


jfx2006:
@escapewindow: review request

escapewindow (approved):

PR #559, opened 2022-09-14
test: pass 'SCRIPTWORKER_GITHUB_TOKEN' env from system to tox

ahal:
This is apparently needed to run the `test_production.py` tests locally.


tox.ini
```diff
@@ -11,6 +11,7 @@ passenv =
     HOME
     NO_TESTS_OVER_WIRE
     NO_CREDENTIALS_TESTS
+    SCRIPTWORKER_GITHUB_TOKEN
```
JohanLorenzo, line 14: I confirm we do need this env var passed down per:

https://github.com/mozilla-releng/scriptworker/blob/98b1fe9ffea85bb42a3cd29b0f1a1e20511914f5/tests/test_production.py#L59-L60

JohanLorenzo (approved):
LGTM!

PR #568, opened 2022-11-03
Update android-components index paths in `test_verify_production_cot`

ahal:


ahal:
Oops didn't mean to include the patch to remove old keys.

ahal:
We'll need to do this for fenix as well when it merges in.

JohanLorenzo (approved):
LGTM! I agree. That makes me realize we don't have a test for focus-android. I'll add one once we merge that one in too.

PR #578, opened 2022-12-14
Fix unused imports for flake8 check

ahal:
I have no idea why this started failing, it doesn't appear to be caused by any change in the code.. but at least the fix is simple.


ahal:
Oh, looks like the Python 3.7 version is using a type checking format involving comments that actually uses the imports. So if I remove them it fails in the py37 version. I guess we'll have to just # noqa them.

gbrownmozilla (approved):

PR #581, opened 2022-12-21
Bug 1799696 - Update CoT production tests with new indices

JohanLorenzo:
~~Some indices don't exist yet so let's wait until we ship a few more Android releases before merging this PR.~~ Edit: It turns they already do exist!


ahal (approved):
Thanks for also adding focus!

PR #582, opened 2023-01-04
Bug 1808492 - add app-services index to production CoT test

jcristau:
Ideally we'd look at a build or signing task instead, but those aren't indexed currently so use the docker-image task for now instead.


jcristau:
In local testing this currently passes, while replacing the index with `project.application-services.cache.level-3.docker-images.v2.linux.pushdate.2022.08.17.20220817182624` (pointing at the previous task, using the old CoT key) properly fails.

ahal (approved):
Thanks!

PR #588, opened 2023-02-10
tests: mark adhoc-signing CoT checks as expected failure, and update fenix index paths

jcristau:
We've rotated the scriptworker CoT key but haven't re-signed everything since.  A lot of these will be fixed next week with the next beta build.


JohanLorenzo (approved):
Thanks! I agree this is the right fix until we get adhoc signing fixed.

jcristau:
It's not so much getting adhoc signing fixed as having a new task to point to; we haven't needed to sign anything new there last week.

PR #592, opened 2023-02-20
Mark firefox-translations CoT test as passing

jcristau:
Today's 1.3.2 release means this test should now pass.


JohanLorenzo (approved):
🎉

PR #594, opened 2023-02-24
fix: Update adhoc-signing index and remove xfail

hneiva:


tests/test_production.py
```diff
@@ -172,16 +172,13 @@ async def get_context(config_override=None):
...
-        },
-        marks=pytest.mark.xfail,
-    ),
+    {
+        "name": "adhoc-signing",
+        "taskcluster_root_url": "https://firefox-ci-tc.services.mozilla.com/",
+        "index": "adhoc.v2.adhoc-signing.latest.taskgraph.decision",
+        "task_type": "decision",
+        "cot_product": "adhoc",
+    },
```
ahal, line 181: Could you add this test in addition to the old one? The old one is going to test the full chain from signing -> docker-image -> decision, while the latter will only test decision. So the old test is still providing coverage that the new one is not.

ahal (changes requested):
See comment

tests/test_production.py
hneiva: Unfortunately we won't be able to use an index. Should we skip testing, or point to a specific manifest?

ahal: Why can't we use the index? Once we adhoc sign something we can just update the index to point to whatever manifest was adhoc signed and remove the `xfail`. It doesn't have to be for this manifest specifically, it can be for any of them.

jcristau (changes requested):
agree with ahal we should keep this pointing at a release-signing task

hneiva:
Ok, after some invalidations on the adhoc repo, should be good

ahal (approved):
Thank you!

PR #596, opened 2023-03-02
Bug 1819863 - Remove references to fennec-profile-manager

gbrownmozilla:


src/scriptworker/constants.py
```diff
@@ -229,16 +229,7 @@
...
-                                        r"^(?P<path>/mozilla-mobile/"
-                                        "(?:"
-                                        "fennec-profile-manager"
-                                        "|firefox-android"
-                                        "|firefox-tv"
-                                        "|reference-browser"
-                                        "))"
-                                        "(/|.git|$)"
-                                    ]
+                                    [r"^(?P<path>/mozilla-mobile/" "(?:" "firefox-android" "|firefox-tv" "|reference-browser" "))" "(/|.git|$)"]
```
gbrownmozilla, line 232: I'm sad about this, but it makes black happy. 

I see that this could be one simple string, but thought I'd leave it like this in case we add back another case, which I presume black would allow to be formatted sensibly.

JohanLorenzo: Yeah, I don't always agree with what `black` outputs but I love the value this tool brings. I don't think about code style anymore when I write code - it's so great! 

Regarding this specific case, I think we can tweak the line a little bit while keeping `black` happy. It should be a matter of concatenating the substrings together: 

```suggestion
                                    [r"^(?P<path>/mozilla-mobile/(?:firefox-android|firefox-tv|reference-browser))(/|.git|$)"]
```

Speaking for `firefox-tv`, I just realized we don't need it anymore either. I'll file a bug about it.

JohanLorenzo (approved):
Sweet! ✅ 

I think we can make the output from `black` a little bit better. More details below.

src/scriptworker/constants.py
JohanLorenzo: I filed [bug 1820469](https://bugzilla.mozilla.org/show_bug.cgi?id=1820469)

PR #598, opened 2023-03-13
Some minor cleanup to 'guess_worker_impl'

ahal:


bhearsum (approved):
Just to be clear: this wouldn't fix the issue we just had with langpacks (that resulted in https://phabricator.services.mozilla.com/D172475 as a workaround) - right? (This is purely just cleanup, that should have no impacts.)

ahal:
> This is purely just cleanup, that should have no impacts

Correct, well other than all the tests I broke ><

ahal:
Sigh, I guess that explicit `scriptworker-prov-v1` check was left in to keep the tests happy. Ideally we'd update the tests to use a dummy configuration, but I'm not sure it's worth it.. may just end up abandoning this PR.

PR #600, opened 2023-04-25
docs: add some info about CoT verification expectations w.r.t decisio…

jcristau:
…n and action tasks


jcristau:
@ahal this is a stab at what I mentioned yesterday, let me know what you think :)

ahal:
Looks great, though it occurs to me that we definitely make security relevant decisions based on `tasks_for`.. It might have been nice if all the necessary context was put into a special `cot` subkey, but probably not worth refactoring everything at this point (maybe once everything is using a single shared template).

jcristau:
> Looks great, though it occurs to me that we definitely make security relevant decisions based on `tasks_for`..
> 
Good call I should use more precise wording there... I'll ponder this some more.

PR #607, opened 2023-06-19
fix readthedocs doc builds

gbrownmozilla:
Doc builds have been busted on readthedocs for a while: I think that's because `conda` was yanked from pypi.

I found local doc builds failed to create a proper TOC; adding myst-parser seemed the fastest fix.


jcristau (approved):
Seems reasonable.

PR #612, opened 2023-07-11
Bug 1817658 - Add Thunderbird "microsoftstore" scopes to scriptworker.

jfx2006:

PR #613, opened 2023-07-27
Bug 1817658 - Add Thunderbird "microsoftstore" scopes to scriptworker.

bhearsum:
This is the same as #612; except it should run tests and be mergable.


bhearsum:
Looks like the original PR will work - closing this.

PR #601, opened 2023-05-04
Add birch as release branch to enable CoT for signing

hneiva:

PR #616, opened 2023-07-28
bug 1845368: add pine to all-nightly-branches

bhearsum:
This will be a permanent project branch that is set-up for the express purpose of testing nightlies and updates.


JohanLorenzo (approved):
LGTM modulo a nit. Per https://github.com/mozilla-releng/scriptworker/pull/601, we also have to update the project regex on `trusted_vcs_rules`:

https://github.com/mozilla-releng/scriptworker/blob/43b53b4d864c3fcee4b0f63c9e56f82609636118/src/scriptworker/constants.py#L199-L211

PR #624, opened 2023-10-26
Bug 1860092 - Add comm-release support to scriptworker.

jfx2006:


jcristau (approved):
LGTM, just needs reformatting to appease black.

PR #626, opened 2023-11-13
Bug 1864413: don't report success for tasks killed by a signal

jcristau:
`run_task`, and thus `do_run_task`, were using the (negative) exit code from the task process as status.  That status was then combined with the 0 from `do_upload` in `RunTasks.invoke` via `worst_level`, and ended up getting swallowed.  Replace that with a failure code.

### DIFF
--- a/tests/data/killed.py
+++ b/tests/data/killed.py
@@ -1,0 +1,6 @@
+#!/usr/bin/python3
+
+import os
+import signal
+
+os.kill(0, signal.SIGKILL)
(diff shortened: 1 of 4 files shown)


tests/data/killed.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/python3
+
+import os
+import signal
+
+os.kill(0, signal.SIGKILL)
```
hneiva, line 6: I'm curious, wouldn't this kill the process on import? How does this work?

hneiva (approved):
LGTM

PR #631, opened 2023-12-09
Bug 1869123 - Allow comm-esr115 to use the "balrog:server:release" restricted scope.

jfx2006:
Fixes a regression from PR #624.

This is a temporary situation while Thunderbird monthly releases are in development. Until comm-esr115 is migrated to use a (new) "esr" update channel, it needs to use the "release" channel.

Details why comm-esr115 was able to use "release" prior to now are in the bug.


jcristau (approved):

jcristau:
@jfx2006 we might need to also update `project:comm:thunderbird:releng:flathub:stable`?

jfx2006:
I need to push a fixup to that. IDE autoformatting got overzealous.

PR #634, opened 2024-01-05
unpin pytest-asyncio

hneiva:
Closes #532

The event loop closed issue seems to stem from `aiomemoizettl.py:6` where it uses the event loop to invalidate the cache. Once the invalidation happens, the event loop is already closed, raising an exception.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ deps =
     mock
     pytest
     # https://github.com/pytest-dev/pytest-asyncio/issues/209
-    pytest-asyncio<0.15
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -24,7 +24,7 @@ deps =
     mock
     pytest
     # https://github.com/pytest-dev/pytest-asyncio/issues/209
```
ahal, line 26: Should we remove this comment too?

ahal (approved):
Nice, thanks for doing this!

PR #644, opened 2024-04-04
feat: add support for deferring upstream artifact selection until runtime

bhearsum:
This patch allows the deferral of upstream artifact selection to runtime by supporting `*` in upstreamArtifacts `paths`.

This is primarily built to support the work being done to add beetmover support for translations in https://github.com/mozilla/firefox-translations-training/issues/466, but there's no reason it wouldn't work more generally. I've tested it a bunch in that context, and the scriptworke parts (download artifacts and verifying CoT) are working well with this patch. See https://firefox-ci-tc.services.mozilla.com/tasks/YThKSz44QQGIF-b_L3958Q/runs/0 and other beetmover tasks in that graph for examples. (Note that while some of the tasks are failing, all of CoT stuff works - which is the part that largely exercises this code.)

### DIFF
--- a/src/scriptworker/artifacts.py
+++ b/src/scriptworker/artifacts.py
@@ -6,6 +6,7 @@ in S3.
 """
 
 import asyncio
+import fnmatch
 import gzip
 import logging
 import mimetypes
@@ -15,6 +16,7 @@ from pathlib import Path
 import aiohttp
 import arrow
 import async_timeout
+from taskcluster.exceptions import TaskclusterFailure
 
 from scriptworker.client import validate_artifact_url
 from scriptworker.exceptions import DownloadError, ScriptWorkerRetryException, ScriptWorkerTaskException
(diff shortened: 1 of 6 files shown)


src/scriptworker/artifacts.py
```diff
@@ -221,6 +223,16 @@ def get_artifact_url(context, task_id, path):
...
 
 
+# list_latest_artifacts {{{1
+async def list_latest_artifacts(queue, task_id, exception=TaskclusterFailure):
+    return await queue.listLatestArtifacts(task_id)
+
+
+async def retry_list_latest_artifacts(queue, task_id, exception=TaskclusterFailure, **kwargs):
+    kwargs.setdefault("retry_exceptions", tuple(set([TaskclusterFailure, exception])))
+    return await retry_async(list_latest_artifacts, args=(queue, task_id), kwargs={"exception": exception}, **kwargs)
```
jcristau, line 233: Do we actually need this?  https://github.com/taskcluster/taskcluster/blob/b09f8e56045981af8e7a163ea3e00b40f074108a/clients/client-py/taskcluster/client.py#L456 seems to have retry logic already.

src/scriptworker/context.py
```diff
@@ -108,7 +108,9 @@ def verify_task(self) -> None:
             task_id: str = upstream_artifact["taskId"]
             for path in upstream_artifact["paths"]:
                 if os.path.isabs(path) or ".." in path:
-                    raise CoTError("upstreamArtifacts taskId {} has illegal path {}!".format(task_id, path))
+                    raise CoTError(f"upstreamArtifacts taskId {task_id} has illegal path {path}!")
+                if "*" in path and not upstream_artifact.get("optional", False):
+                    raise CoTError(f"upstreamArtifacts taskId {task_id} has globbed path {path} as a non-optional artifact!")
```
jcristau, line 113: Curious why this is an error?

jcristau (approved):

src/scriptworker/artifacts.py
bhearsum: Ah, indeed. I'll remove this.

Issue #645, opened 2024-04-26
verify_cot fails to validate pull requests on repos that are forks


ahal:
While testing some bitrisescript changes on `staging-firefox-ios`, @isabelrios discovered a CoT failure:
https://firefox-ci-tc.services.mozilla.com/tasks/C4i1LZAQS1iLhBbATtJBBw/runs/0/logs/public/logs/chain_of_trust.log

This is a json-e rendering error:

> jsone.shared.InterpreterError: InterpreterError at template.tasks[0].repoUrl: infix: . expects objects

It happens because the `.taskcluster.yml` evaluates [event.pull_request.head.repo.html_url](https://github.com/mozilla-releng/staging-firefox-ios/blob/995926262571e9627eee4070b4f82414107aed2f/.taskcluster.yml#L40) (which is standard in all our `.taskcluster.yml` files), but the Github context that `verfiy_cot` resolves has:
```
pull_request:
    head:
        repo: None
```

I believe this is happening because `staging-firefox-ios` is a fork of `firefox-ios`, so it means we hit this condition:
https://github.com/mozilla-releng/scriptworker/blob/cd0aac0510f69386b28a375656b5cc070dc2fb7f/src/scriptworker/cot/verify.py#L1179

This causes us to use `mozilla-mobile/firefox-ios` as the head repository instead, and my theory is this causes the `pull_request.head.repo` to be None as the pull request in question doesn't exist over there.

ahal:
My intuition is that the `if repo_definition["fork"]` logic is wrong and should be deleted. Whether a repo is a fork or not shouldn't matter, we should only care about where the task is running.

ahal:
I can confirm that removing the `if repo_definition["fork"]` block fixes Isabel's failure.

bhearsum:
Did #646 resolve this?

PR #646, opened 2024-04-26
fix(cot): don't use parent repo when evaluating pull request json-e c…

ahal:
…ontext on forks

Issue: #645

### DIFF
--- a/src/scriptworker/constants.py
+++ b/src/scriptworker/constants.py
@@ -171,7 +171,7 @@ DEFAULT_CONFIG: immutabledict[str, Any] = immutabledict(
                         "glean": ("glean-1/decision", "glean-3/decision", "glean-1/decision-gcp", "glean-3/decision-gcp"),
                         "xpi": ("xpi-1/decision", "xpi-3/decision", "xpi-1/decision-gcp", "xpi-3/decision-gcp"),
                         "adhoc": ("adhoc-1/decision", "adhoc-3/decision", "adhoc-1/decision-gcp", "adhoc-3/decision-gcp"),
-                        "scriptworker": ("scriptworker-1/decision", "scriptworker-3/decision" "scriptworker-1/decision-gcp", "scriptworker-3/decision-gcp"),
+                        "scriptworker": ("scriptworker-1/decision", "scriptworker-3/decision", "scriptworker-1/decision-gcp", "scriptworker-3/decision-gcp"),
                     }
                 )
             }
--- a/src/scriptworker/cot/verify.py
+++ b/src/scriptworker/cot/verify.py
@@ -1167,6 +1167,8 @@ async def _get_additional_github_pull_request_jsone_context(decision_link):
     context = decision_link.context
     source_env_prefix = context.config["source_env_prefix"]
     task = decision_link.task
+    base_repo_url = get_repo(task, source_env_prefix, repo_type="base")
+    base_repo_url = base_repo_url.replace("git@github.com:", "ssh://github.com/", 1)
     repo_url = get_repo(task, source_env_prefix)
     repo_url = repo_url.replace("git@github.com:", "ssh://github.com/", 1)
     repo_owner, repo_name = extract_github_repo_owner_and_name(repo_url)
@@ -1176,7 +1178,11 @@ async def _get_additional_github_pull_request_jsone_context(decision_link):
     github_repo = GitHubRepository(repo_owner, repo_name, token)
     repo_definition = github_repo.definition
 
-    if repo_definition["fork"]:
+    # We need to query the repository where the pull request was made to extract
+    # pull request data. The pull request could be created on the same repo as
+    # the commit, or an upstream repo. We can compare the base and head repo URLs
+    # to infer where the pull request lives.
+    if repo_definition["fork"] and base_repo_url != repo_url:
         github_repo = GitHubRepository(owner=repo_definition["parent"]["owner"]["login"], repo_name=repo_definition["parent"]["name"], token=token)
 
     pull_request_data = await github_repo.get_pull_request(pull_request_number)
--- a/src/scriptworker/task.py
+++ b/src/scriptworker/task.py
@@ -178,7 +178,7 @@ def get_parent_task_id(task):
 
 
 # get_repo {{{1
-def get_repo(task, source_env_prefix):
+def get_repo(task, source_env_prefix, repo_type="head"):
     """Get the repo for a task.
 
     Args:
@@ -191,7 +191,7 @@ def get_repo(task, source_env_prefix):
         None: if not defined for this task.
 
     """
-    repo = _extract_from_env_in_payload(task, source_env_prefix + "_HEAD_REPOSITORY")
+    repo = _extract_from_env_in_payload(task, f"{source_env_prefix}_{repo_type.upper()}_REPOSITORY")
     if repo is not None:
         repo = repo.rstrip("/")
     return repo
--- a/tests/test_cot_verify.py
+++ b/tests/test_cot_verify.py
@@ -199,6 +199,7 @@ def mobile_github_pull_request_link(mobile_chain):
     )
     decision_link.task["payload"]["env"] = {
         "MOBILE_BASE_REF": "main",
+        "MOBILE_BASE_REPOSITORY": "https://github.com/mozilla-mobile/firefox-android",
         "MOBILE_BASE_REV": "baserev",
         "MOBILE_HEAD_REF": "some-branch",
         "MOBILE_HEAD_REPOSITORY": "https://github.com/JohanLorenzo/reference-browser",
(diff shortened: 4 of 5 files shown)


ahal:
I'm looking for some extra scrutiny on this PR. I think this is correct, but I don't feel very confident. There's a deeper explanation of what this is fixing in issue #645

bhearsum:
I don't have any prior knowledge of whether this is important or not :(. It looks like this was added by @JohanLorenzo in https://github.com/mozilla-releng/scriptworker/pull/307/commits/93f061f68ed1661471550286453980deb8c743da. I don't see any discussion of this specific point there - but maybe @JohanLorenzo will have some memory of this.

I do think that this specific change is _very_ low risk, as literally the only thing we pull from it is the pull request data. There is certainly no point in trying to pull that data from the wrong repository. If there's something wrong or insecure about pulling it from a fork, we should simply desupport this behaviour altogether.

I will note that many of our staging repos are _not_ created as forks. For example https://github.com/mozilla-releng/staging-adhoc-signing and https://github.com/mozilla-releng/staging-xpi-manifest. This was done at least in part to ensure that they are fully disconnected from production, and as close as possible to a production scenario. I don't know whether this code is scriptworker was a reason we created them like this, though. It is, at least, another way to deal with the accute issue in the staging-firefox-ios repository. The fact that this workaround exists also seems to suggest that this fix is a good one - nothing about the security properties of the repository changes with it being a fork vs. not being a fork. Scopes remain the same, access control on the repo remains the same, etc.

ahal:
> I will note that many of our staging repos are not created as forks.

Ah interesting, I recall a conversation sclements drove where we agreed that staging repos *should* be forks. IIRC, she even went and made a bunch of them forks. This might have happened while you were off the team.

ahal:
> nothing about the security properties of the repository changes with it being a fork vs. not being a fork. Scopes remain the same, access control on the repo remains the same, etc.

Yes, this was my thinking too. Also we don't typically run tasks on user forks anyway so I don't think we would ever hit this condition outside of forked `staging-*` repos.

JohanLorenzo:
> It looks like this was added by @JohanLorenzo in https://github.com/mozilla-releng/scriptworker/commit/93f061f68ed1661471550286453980deb8c743da. I don't see any discussion of this specific point there - but maybe @JohanLorenzo will have some memory of this.

For an unknown reason, the commit should point to https://github.com/mozilla-releng/scriptworker/commit/d08d306e50cd8011fe7cba37ab6300560a58e850, which in turns point to https://github.com/mozilla-releng/scriptworker/pull/307. Tom, Aki, and I had discussion there. Let me read them again to see if they refresh my memory

JohanLorenzo:
> Ah interesting, I recall a conversation sclements drove where we agreed that staging repos should be forks.

Oh, I thought they should be forks too. E.g.: https://github.com/mozilla-releng/staging-firefox-android. I didn't know we had some that weren't. 

> I think this is correct, but I don't feel very confident.

Here are some cases I tested when I introduced the behavior back then: https://github.com/mozilla-releng/scriptworker/pull/307#issuecomment-461106511. My main worry here is that we break PRs made against the main repo. If that works, then we covered the main case, in my opinion. 

I don't remember why I added this code in the first place. I'm fine removing it as long as we don't regress anything. Unlike then, testing CoT today against several known scenarios should be easier. We don't have to schedule graphs anymore, we can just CoT locally.

ahal:
Good call @JohanLorenzo! This does break regular pull requests. Looking at the error it makes sense.. we do need to get pull request data from the parent repo (in my mind I was thinking that the tasks run on the base repo so it would never be in the context of the fork.. but looks like I was wrong).

I guess the issue is that whether the current repo is a fork or not, is not the right check to use here. I'll poke around a bit.. but this patch introduces a much worse regression than it solves, so if I don't find anything I'll WONTFIX.

ahal:
This latest iteration works both with regular pull requests and Isabel's case. I'm feeling more confident about this iteration.

ahal:
Oops, missed some tests, will fix tomorrow.

src/scriptworker/constants.py
```diff
@@ -171,7 +171,7 @@
                         "glean": ("glean-1/decision", "glean-3/decision", "glean-1/decision-gcp", "glean-3/decision-gcp"),
                         "xpi": ("xpi-1/decision", "xpi-3/decision", "xpi-1/decision-gcp", "xpi-3/decision-gcp"),
                         "adhoc": ("adhoc-1/decision", "adhoc-3/decision", "adhoc-1/decision-gcp", "adhoc-3/decision-gcp"),
-                        "scriptworker": ("scriptworker-1/decision", "scriptworker-3/decision" "scriptworker-1/decision-gcp", "scriptworker-3/decision-gcp"),
```
bhearsum, line 174: Obviously this is broken:
```
>>> ("scriptworker-1/decision", "scriptworker-3/decision" "scriptworker-1/decision-gcp", "scriptworker-3/decision-gcp")
('scriptworker-1/decision', 'scriptworker-3/decisionscriptworker-1/decision-gcp', 'scriptworker-3/decision-gcp')
```

Is it part of this fix too, or just a ridealong? (I don't care either way, just wondering, and also wondering if this is actively causing issues somewhere...)

src/scriptworker/cot/verify.py
```diff
@@ -1176,7 +1179,7 @@ async def _get_additional_github_pull_request_jsone_context(decision_link):
     github_repo = GitHubRepository(repo_owner, repo_name, token)
     repo_definition = github_repo.definition
 
-    if repo_definition["fork"]:
+    if repo_definition["fork"] and base_repo_url != repo_url:
```
bhearsum, line 1182: I _think_ this is fine. It looks like the original `github_repo` comes from `head_repository` (via `get_repo`). From there we have a few cases:
* Head `github_repo` is not a fork, and we keep it. Eg: opening a PR to `firefox-ios` from a branch in that repo. Obviously we need the original `github_repo` in this case (and redoing it would end up with the same data anyways...)
* Head `github_repo` is a fork, and the base repo is different. Eg: opening a PR to `firefox-ios` from `staging-firefox-ios`, or even `staging-firefox-ios` from some user's repository. In this case we must redefine `github_repo` again, because we can't trust whatever data comes out of the head `github_repo`. I guess we can trust whatever `repo_definition["parent"]` provides us because there's no way it could be anything _except_ the repo that the event is firing from?
* Head `github_repo` is a fork and the base repo is the same. Eg: opening a PR to `staging-firefox-ios` from another branch in that repository. We can use the existing head `github_repo` in this case because the only people that could open this PR would be those that can modify the one on `main` anyways? (I guess that's mostly true...depending on branch protections we could theoretically have a case where someone could push to a branch, open a PR, but be unable to modify things on `main` directly. I'm not sure if that's worth dealing with, and it's certainly not something to deal with here if so.)

Two suggestions:
* Add a comment that explains this stuff a bit - it's a little bit head spinning IMO (doesn't need to be my textwall :laughing:)
* Consider defaulting to the base `github_repo`. Obviously this doesn't change anything now, but it might protect against accidents in a future refactor.

bhearsum (changes requested):
Marking as request changes due to the test issues. I assume you'll add some variants for the various base/head fork/not-fork cases in addition to fixing the existing tests?

src/scriptworker/constants.py
ahal: I hit this bug while trying to run `verify_cot` manually in order to test this PR. So it is causing active issues, but I guess we don't run into it in day to day. It's not related to this in a technical capacity.

src/scriptworker/cot/verify.py
ahal: Comment added!

It occurred to me that we could have the case that `github_repo` is not a fork, but it still opens a pull request on an upstream repo. E.g, we open a pull request from `staging-xpi-manifest` to `xpi-manifest`. But in this case I don't think the parent repo information would be populated in the Github API return value.. so I'm not sure there's anything we can do about it unless we trust what's in the `BASE_REPOSITORY` env

ahal: > Consider defaulting to the base github_repo. Obviously this doesn't change anything now, but it might protect against accidents in a future refactor.

Sorry, I'm confused by what you mean here

bhearsum: > > Consider defaulting to the base github_repo. Obviously this doesn't change anything now, but it might protect against accidents in a future refactor.
> 
> Sorry, I'm confused by what you mean here

I guess what I'm trying to guard against is some future refactor that causes a bug where we don't set `github_repo` a second time, and we wrongly end up with the original `github_repo` definition. In my head when I wrote the earlier comment that might be a security issue...but looking at it again the worst that can happen is a failure to verify CoT....so I don't really care now!

jcristau (approved):
I wish we just used the base repo for everything all the time.. This looks like a reasonable fix though.

jcristau:
Side note: the test changes will conflict with #648.

bhearsum:
> I wish we just used the base repo for everything all the time.. This looks like a reasonable fix though.

How would this work with PRs? (I totally agree with regard to any other event type...)

jcristau:
> > I wish we just used the base repo for everything all the time.. This looks like a reasonable fix though.
> 
> How would this work with PRs? (I totally agree with regard to any other event type...)

I don't think CoT really cares?  and run-task could fetch `refs/pull/ID/head` from the base repo instead of pulling from the head repo.

bhearsum:
> > > I wish we just used the base repo for everything all the time.. This looks like a reasonable fix though.
> > 
> > 
> > How would this work with PRs? (I totally agree with regard to any other event type...)
> 
> I don't think CoT really cares? and run-task could fetch `refs/pull/ID/head` from the base repo instead of pulling from the head repo.

Maybe I'm missing something, but given that decision tasks are built based on pull request versions of `.taskcluster.yml` (See [this PR](https://github.com/mozilla-releng/balrog/pull/3115) and [decision task](https://firefox-ci-tc.services.mozilla.com/tasks/AIPWF4wwS22dG8_exTcZFQ)), CoT presumably needs to find that version of the `.taskcluster.yml`?

jcristau:
It should be enough to have the base repo and commit hash for that, e.g. https://github.com/mozilla-releng/balrog/raw/10dc4169f8828751e93b084c8e2bc550665a4031/.taskcluster.yml?

bhearsum:
> It should be enough to have the base repo and commit hash for that, e.g. https://github.com/mozilla-releng/balrog/raw/10dc4169f8828751e93b084c8e2bc550665a4031/.taskcluster.yml?

Ah right, I forgot that you could access fork content like that.

bhearsum (approved):
We may want a follow-up issue or two given some of the discussion here, but I agree this is a fine fix for now, modulo the needed rebase.

PR #648, opened 2024-05-18
Bug 1896964 - Cleanup firefox-android repo references

gbrownmozilla:


tests/test_github.py
```diff
@@ -353,20 +353,20 @@ def test_extract_github_repo_ssh_url(repo_url, expected, raises):
     "repo_url, expected_user, expected_repo_name, raises",
     (
         (
-            "https://github.com/mozilla-releng/staging-firefox-android/raw/0123456789abcdef0123456789abcdef01234567/.taskcluster.yml",
-            "https://github.com/mozilla-releng/staging-firefox-android",
+            "https://github.com/mozilla-releng/staging-reference-browser/raw/0123456789abcdef0123456789abcdef01234567/.taskcluster.yml",
+            "https://github.com/mozilla-releng/staging-reference-browser",
```
jcristau, line 357: I was confused for a bit here, because the parameter names don't actually match what they are (copy-pasted from test_extract_github_repo_owner_and_name, by the looks of it).

jcristau (approved):

tests/test_github.py
gbrownmozilla: Agreed - that's strange! Fixed.

PR #654, opened 2024-06-14
fix: adjust mobile CoT verify tests after migration into Gecko repo

bhearsum:
There's no longer indexes for the a-c tasks, so I'm just removing them (I can't see any value in going to the trouble of re-adding indexes). The focus/fenix ones are simple adjustments.


gbrownmozilla:
Sorry I missed that. Thanks - this looks great.

PR #657, opened 2024-06-24
Update 60.1.0 changelog to include deferring upstream artifact selection until runtime

bhearsum:
#644 merged after the version bump, but before we tagged for 60.1.0; may as well include it.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,7 @@ This project adheres to `Semantic Versioning <http://semver.org/>`__.
 Added
 ~~~~~
 - Values for Firefox Translations training repository scriptworker constants
+- Support for deferring upstream artifact selection until runtime
 
 Fixed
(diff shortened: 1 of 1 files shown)


HISTORY.rst
```diff
@@ -10,6 +10,7 @@ This project adheres to `Semantic Versioning <http://semver.org/>`__.
 Added
 ~~~~~
 - Values for Firefox Translations training repository scriptworker constants
+- Support for deferring upstream artifact selection until runtime
```
jcristau, line 13: I should have mentioned in the other PR, but this might warrant an update to the docs, at least cot_verify.rst, to mention globs.

jcristau (approved):

HISTORY.rst
bhearsum: Good call; I will follow up with this.

PR #661, opened 2024-06-27
feat: set User-Agent in our http requests

### LINKED ISSUE #660: scriptworker should set specific user agent for its http requests
http requests for e.g. artifact download and CoT verification should have a scriptworker-specific user-agent instead of something generic like "Python/3.11 aiohttp/3.9.5".

jcristau:
Add scriptworker.utils.scriptworker_session to create our aiohttp ClientSession objects in a single place.

Fixes #660


bhearsum (approved):
Good idea; thank you!

PR #664, opened 2024-07-12
ci: for pushes, only run taskcluster on main

bhearsum:
Anything else should route through a PR. If we need other long lived branches in the future we can add them, and also add them to https://github.com/mozilla-releng/fxci-config/blob/main/projects.yml.


bhearsum:
Needs rebasing after https://github.com/mozilla-releng/scriptworker/pull/665 is merged.

jcristau (approved):

PR #676, opened 2025-01-08
fix: update translations repo in constants

bhearsum:
This repository was renamed, and the source_env_prefix we use in `.taskcluster.yml` is derived from the repo name. We need to update this to match, as well as update the repo name itself.


jcristau (approved):

bhearsum:
I've updated this patch to fix the translations repo name as well. The test failures are unrelated. An indexed task in adhoc-signing that they depend on expired.

PR #681, opened 2025-01-30
bug 1944603: fix: add maple to all-nightly-branches

bhearsum:
...so that we can test L3 signing there without complaints.

This probably should've been done with https://github.com/mozilla-releng/scriptworker/commit/2296f2509ed011e18ec25f3a31f80b0e4106eeba

PR #682, opened 2025-01-30
chore: reformat with black

bhearsum:


jcristau:
Should we have a CI task for this?

jcristau:
Nevermind, I see #681.  Odd though, do we not pin the version of black? Or how did this pass before?

jcristau (approved):

bhearsum:
> Nevermind, I see #681. Odd though, do we not pin the version of black? Or how did this pass before?

I presume last time it ran it picked up an older version of black.

PR #684, opened 2025-02-25
fix: properly support globbed upstream artifacts in subdirectories

bhearsum:
The original code for this looks like it supports them, and the tests pass, but the tests weren't set-up correctly (they specified the pattern as `*` instead of `dir/*`).

This update fixes this and adds tests to ensure that artifacts in subdirectories and not in subdirectories continue to work.


src/scriptworker/artifacts.py
```diff
@@ -440,6 +440,7 @@ def get_artifacts_matching_glob(context, task_id, pattern):
     matching = []
     for root, _, files in os.walk(parent_dir):
         for f in files:
-            if fnmatch.fnmatch(f, pattern):
+            relpath = os.path.relpath(os.path.join(root, f), parent_dir)
+            if fnmatch.fnmatch(relpath, pattern):
```
hneiva, line 444: if the "globbed" entry doesn't match anything, should we error out?

hneiva (approved):
one question, but not a block for landing this.
LGTM

src/scriptworker/artifacts.py
bhearsum: I could see an argument for that being optional behaviour, but that can be implemented if the use case ever comes up.

PR #685, opened 2025-02-25
slight clarification in flow of control documentation

bhearsum:
I ended up down a rabbithole again wondering why `landoscript` was being run when there were no references in the codebase to it. (It's found in the `task_script` variable, whose value is only known at runtime, making this virtually ungrepable.)


hneiva (approved):
👍

PR #693, opened 2025-04-16
feat: Add Firefox iOS CoT scopes

hneiva:


hneiva:
@jcristau would you know if we need the staging repo here as well?

jcristau (approved):

PR #696, opened 2025-05-22
feat: upload all artifacts even when SIGTERM is received

bhearsum:
SIGTERM handling was originally added in https://github.com/mozilla-releng/scriptworker/pull/310. There is a [conversation thread](https://github.com/mozilla-releng/scriptworker/pull/310#discussion_r253707819) there about not uploading artifacts other than logs, but doesn't have any strong justification for this behaviour:
* "I think that the potentially-incomplete/corrupt artifacts won't have much value"
* "it [uploading files before being killed] becomes tougher"

I don't think either of these hold true at this point. I have a very strong use case for uploading non-log files in https://bugzilla.mozilla.org/show_bug.cgi?id=1967787 (making it trivial to find lando job status in a rerun). We've also proven out in other contexts that we can uploading 10s of gigabytes from a GCE instance to GCS in less than 30 seconds. In scriptworkers, we have a 2 minute grace period before we're killed, and we have no scriptworker tasks that I'm aware of that that upload more than 1-2 gigabytes.

### DIFF
--- a/src/scriptworker/worker.py
+++ b/src/scriptworker/worker.py
@@ -7,7 +7,6 @@ Attributes:
 """
 import asyncio
 import logging
-import os
 import signal
 import socket
 import sys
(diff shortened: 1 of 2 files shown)


jcristau (approved):
Could we update e.g. `test_run_tasks_cancel_run_tasks` to assert this?

bhearsum:
> Could we update e.g. `test_run_tasks_cancel_run_tasks` to assert this?

Yeah, good call.

PR #704, opened 2025-06-04
Repin aiohttp to >= 3 and bump version to 60.10.2

Eijebong:
3.12.9 contains the fix for
https://github.com/aio-libs/aiohttp/issues/11138 which caused our OoM issues on artifacts upload.

Unfortunately we cannot repin to >=3.12.9  because it doesn't support python 3.8, that should be fine as it should always take the latest version but it's something to keep in mind if someone's doing partial dependency updates


Eijebong:
Urgh... Fine... I'll unpin it completely then

PR #707, opened 2025-07-07
Switch project to uv

Eijebong:


Eijebong:
Checked that the github workflow was working/doing the right thing: https://github.com/Eijebong/scriptworker/actions/runs/16117890279/job/45475969968 (ls & ls dist/), looks correct.
I don't know how to check readthedocs but I copied it from taskgraph and just updated the uv version

src/scriptworker/__init__.py
```diff
@@ -1 +1,5 @@
 """Scriptworker."""
+
+import importlib.metadata
+
+__version__ = importlib.metadata.version("scriptworker")
```
bhearsum, line 5: Yay, this a small, but fantastic improvement.

pyproject.toml
```diff
@@ -1,3 +1,96 @@
+[project]
```
bhearsum, line 1: If I understand correctly, we'll also need a block like this:
```
[tool.hatch.build.targets.wheel]
packages = ["src/scriptworker"]
```

bhearsum (changes requested):
Does this have any implications in the scriptworker-scripts repo when we first release this? It seems like it shouldn't, but it's worth checking if possible (even if it's immediately after this is merged and released).

pyproject.toml
bhearsum: Weird. I tried a similar pyproject.toml without a `packages` block, and it failed to even make a package for me. (We also have this block in ship it: https://github.com/mozilla-releng/shipit/blob/0e50df631305a2002e934d9db81dfb762a55c99a/api/pyproject.toml#L73-L74.)

Presumably there's some magic under the hood here automatically adding `src`?

.github/workflows/pypi-publish.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           cache: 'pip'
       - name: Build package distributions
         run: |
-          pip install wheel
-          python setup.py sdist bdist_wheel
+          pip install hatch
```
hneiva, line 21: TIL hatch

PR #708, opened 2025-07-07
feat: set a max age for projects.yml caching

bhearsum:


src/scriptworker/context.py
```diff
@@ -237,17 +240,22 @@ def event_loop(self, event_loop: asyncio.AbstractEventLoop) -> None:
...
         Args:
             force (bool, optional): Re-run the download, even if ``self.projects``
                 is already defined. Defaults to False.
 
         """
         assert self.config
-        if force or not self.projects:
+        now = time.time()
+        last_fetched = now - self._projects_timestamp
+        if force or not self.projects or last_fetched > 3600:
```
hneiva, line 255: nit: I would make the `3600` here a const up top so it's easier to spot/adjust if necessary.

src/scriptworker/context.py
```diff
@@ -15,6 +15,7 @@
 import logging
 import os
 import tempfile
+import time
```
hneiva, line 18: at first glance I was going to suggest just using datetime and timedelta for better readability, but after a second pass on this, I'm ok with using numeric calculations with simpler code.

hneiva (approved):

src/scriptworker/context.py
```diff
@@ -237,17 +241,22 @@ def event_loop(self, event_loop: asyncio.AbstractEventLoop) -> None:
     async def populate_projects(self, force: bool = False) -> None:
         """Download the ``projects.yml`` file and populate ``self.projects``.
 
-        This only sets it once, unless ``force`` is set.
+        Unless ``force`` is set or it was last fetched over an hour ago this
```
jcristau, line 244: "over an hour ago" is now a lie.  Should be something like "over PROJECTS_YML_MAX_AGE_SECONDS seconds ago".

tests/test_context.py
```diff
@@ -88,16 +88,25 @@ async def fake_load(*args):
...
     assert fake_projects["count"] == 1
 
     await rw_context.populate_projects(force=True)
     assert rw_context.projects == fake_projects
+    # forced, must be fetched
     assert fake_projects["count"] == 2
 
     await rw_context.populate_projects()
     assert rw_context.projects == fake_projects
+    # already exists, and age is less than an hour, noop
```
jcristau, line 101: ditto.

PR #724, opened 2025-11-13
Bug 1991990: Add Scope for Thunderbird ESR Channel

stephenking314:


stephenking314:
> Presumably this needs a companion fxci-config patch to actually grant that scope?

Good catch!  I'm still learning how this all comes together, so I appreciating you thinking it through!

Here's the PR for fxci-config: https://github.com/mozilla-releng/fxci-config/pull/633

PR #791, opened 2026-04-23
feat(cot): Bug 2034520 improve json download and parsing error handling

hneiva:
On `load_json_or_yaml`:

    load_json_or_yaml's default exception message now names the file when
    is_path=True ("Failed to load json from /path/to/x.json: ..." instead
    of just "Failed to load json: ..."). load_json_or_yaml_from_url passes
    a custom message that also includes the (loggable) source URL alongside
    the cache path.
    
    Motivation: a chain-of-trust verification failure surfaced only the
    JSONDecodeError offset, forcing a manual trace through several
    concurrent downloads to identify the bad file. The extra context pins
    the culprit immediately.

On `load_json_or_yaml_from_url`:

    Combine download and parse into a single retry unit in
    load_json_or_yaml_from_url: if the downloaded file fails to parse as
    JSON/YAML, the cache is invalidated and retry_async triggers a fresh
    download (up to the default 5 attempts). This turns transient artifact-
    server glitches (e.g. GCS gunzip-transcoding returning 'Bad Request' with
    200 OK, HTML error pages from hg, truncated responses) from hard failures
    into self-healing retries.
    
    download_file gains an optional expected_content_type kwarg that rejects
    'text/html' responses when the caller asked for something else, failing
    fast with a clearer error before writing a byte. DownloadError is reused
    so retry_async picks it up if the HTML turns out to be transient.
    
    The misleading 'overwrite' parameter semantics are preserved
    (overwrite=True keeps any existing cached file; overwrite=False always
    redownloads) and flagged with a comment for future cleanup.
    
    Tests added for HTML rejection and the parse-failure-then-retry-succeeds
    flow.

### DIFF
--- a/src/scriptworker/utils.py
+++ b/src/scriptworker/utils.py
@@ -523,7 +523,7 @@ def format_json(data):
 # is omitted we don't actually ever return None (because on failure we raise an Exception)
 @overload
 def load_json_or_yaml(
-    string: str, is_path: Optional[bool] = ..., file_type: Optional[str] = ..., exception: Type[BaseException] = ..., message: str = ...
+    string: str, is_path: Optional[bool] = ..., file_type: Optional[str] = ..., exception: Type[BaseException] = ..., message: Optional[str] = ...
 ) -> Dict[str, Any]:  # pragma: no cover
     ...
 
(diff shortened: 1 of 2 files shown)


ahal (approved):
Thanks, lgtm!

PR #792, opened 2026-04-27
Download to temp file and rename atomically in download_file

Eijebong:
Instead of downloading to the final file, download to a temporary file then move the file into place. This prevents two different coroutines rom writing chunks one after another into the same file. This feels a bit like a bandaid for something that shouldn't happen in the first place (we shouldn't be trying to download the same file twice) but it's a safe patch for a problem that is very real anyway as a failure would leave a corrupted file on the system right now and any retry would just see that and fail later on.

I intend on fixing the fact that we are downloading the same file multiple times in a follow up by caching the coroutine instead but this is a worthwhile fix anyway

### DIFF
--- a/src/scriptworker/utils.py
+++ b/src/scriptworker/utils.py
@@ -16,6 +16,7 @@ import random
 import re
 import shutil
 import time
+import uuid
 from copy import deepcopy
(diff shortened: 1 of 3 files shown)


bhearsum (approved):
wfm as a short term fix

PR #796, opened 2026-05-15
feat(logging): Bug 2039779 Copy chain_of_trust.log to live_backing.log on CoT verification failure

hneiva:
When verify_chain_of_trust fails, run_task never runs and live_backing.log is never created. Treeherder parses live_backing.log as the task's failure log, so users could not see why CoT verification failed.

Copy chain_of_trust.log to live_backing.log on verification failure so the verification output is visible from TH.

### DIFF
--- a/src/scriptworker/worker.py
+++ b/src/scriptworker/worker.py
@@ -8,6 +8,7 @@ Attributes:
 
 import asyncio
 import logging
+import os
 import signal
 import socket
 import sys
@@ -17,12 +18,13 @@ from typing import Any
 import aiohttp
 import arrow
 
-from scriptworker.artifacts import upload_artifacts
+from scriptworker.artifacts import create_link_artifact, upload_artifacts
 from scriptworker.config import get_context_from_cmdln
 from scriptworker.constants import STATUSES
 from scriptworker.cot.generate import generate_cot
 from scriptworker.cot.verify import ChainOfTrust, verify_chain_of_trust
 from scriptworker.exceptions import ScriptWorkerException, WorkerShutdownDuringTask
+from scriptworker.log import get_chain_of_trust_log_filename
 from scriptworker.task import claim_work, complete_task, prepare_to_run_task, reclaim_task, run_task, worst_level
 from scriptworker.task_process import TaskProcess
 from scriptworker.utils import cleanup, filepaths_in_dir, scriptworker_session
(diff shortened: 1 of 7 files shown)


src/scriptworker/worker.py
```diff
@@ -53,7 +56,14 @@ async def do_run_task(context, run_cancellable, to_cancellable_process):
...
         if context.config["verify_chain_of_trust"]:
             chain = ChainOfTrust(context, context.config["cot_job_type"])
-            await run_cancellable(verify_chain_of_trust(chain))
+            try:
+                await run_cancellable(verify_chain_of_trust(chain))
+            except Exception:
+                # Surface CoT verification output in live_backing.log so Taskcluster/Treeherder show it
+                cot_log = get_chain_of_trust_log_filename(context)
+                if os.path.exists(cot_log):
+                    shutil.copyfile(cot_log, get_log_filename(context))
```
jcristau: Instead of copying the file we can create a link or redirect artifact?

src/scriptworker/worker.py
```diff
@@ -53,7 +56,14 @@ async def do_run_task(context, run_cancellable, to_cancellable_process):
     try:
         if context.config["verify_chain_of_trust"]:
             chain = ChainOfTrust(context, context.config["cot_job_type"])
-            await run_cancellable(verify_chain_of_trust(chain))
+            try:
+                await run_cancellable(verify_chain_of_trust(chain))
+            except Exception:
```
jcristau, line 60: just catch `CoTError`

hneiva: That would make it skip unforeseeable exceptions, which we don't want.

https://github.com/mozilla-releng/scriptworker/blob/0ae448c9ac7afbb11d3b0f536c06f82eaf31ad3f/src/scriptworker/cot/verify.py#L2053-L2074
This block in `verify_chain_of_trust()` only captures a set of expected exceptions, but doesn't capture json, OS, Value or aiohttp errors.

src/scriptworker/worker.py
hneiva: What does redirecting the artifact mean? Just moving instead of copying? Do we know for a fact that there's nothing externally pointing to `chain_of_trust.log`?

What is the benefit of either vs just copying the file?

jcristau: https://docs.taskcluster.net/docs/reference/platform/queue/api#createArtifact describes link and redirect artifacts.  The benefit is to make it obvious that they're the same artifact, just like on generic-worker live.log redirects to live_backing.log.

src/scriptworker/worker.py
```diff
@@ -53,7 +55,21 @@ async def do_run_task(context, run_cancellable, to_cancellable_process):
...
+                if os.path.exists(get_chain_of_trust_log_filename(context)):
+                    try:
+                        await create_link_artifact(
+                            context,
+                            target_path="public/logs/live_backing.log",
+                            link_to="public/logs/chain_of_trust.log",
+                            content_type="text/plain",
+                        )
+                    except Exception as link_exc:
+                        log.warning("Failed to create live_backing.log link artifact: {}".format(link_exc))
```
jcristau: `log.exception`?

jcristau (approved):

PR #797, opened 2026-05-15
feat(lint): Replace black, isort, and flake8 with ruff

hneiva:
The `tox -e check` env now runs `ruff check` + `ruff format --check`; `pyproject.toml` carries a single `[tool.ruff]` block in place of `[tool.black]`
and `[tool.isort]`; the `[flake8]` section in tox.ini is removed.
    
Also fixes 18 pydocstyle (D-rule) violations that flake8_docstrings had been silently ignoring (the plugin was effectively a no-op without an explicit docstring-convention setting). One-time `ruff format` pass touched 7 files with cosmetic-only changes (long-call wrapping, implicit-string collapsing).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,11 +43,8 @@ dev = [
     "tox-uv",
     "aiohttp>=3",
     "asyncio_extras",
-    "black",
     "coveralls",
-    "flake8",
-    "flake8_docstrings",
-    "isort>=5",
+    "ruff",
     "mock",
     "pytest",
     "pytest-asyncio",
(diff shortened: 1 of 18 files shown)


hneiva:
I did my best to review the function docs/explanations that claude generated. I'm happy to adjust if anyone thinks it's not good enough.

pyproject.toml
```diff
@@ -85,34 +82,22 @@ verify_cot = "scriptworker.cot.verify:verify_cot_cmdln"
...
-  | ui
-)/
-'''
+target-version = "py311"
+extend-exclude = [
+    "docs"
+]
+
+[tool.ruff.lint]
+# F = pyflakes, E/W = pycodestyle, I = isort, D = pydocstyle (replaces flake8_docstrings)
```
Eijebong:
```suggestion
# F = pyflakes, E/W = pycodestyle, I = isort, D = pydocstyle
```

The end of that comment has no value outside of the changeset. It should be in the commit message if it has any value.

pyproject.toml
```diff
@@ -85,34 +82,22 @@ verify_cot = "scriptworker.cot.verify:verify_cot_cmdln"
...
+"tests/**" = ["D", "E501"]  # mirror current flake8 `tests:*:D` per-file-ignore + ignore long lines
 
-[tool.isort]
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-line_length = 160
+[tool.ruff.lint.pydocstyle]
+convention = "pep257"  # matches flake8_docstrings default
```
Eijebong:
```suggestion
convention = "pep257"
```

This comment has no value outside of that changeset and should be in the commit message.

pyproject.toml
```diff
@@ -85,34 +82,22 @@ verify_cot = "scriptworker.cot.verify:verify_cot_cmdln"
...
+extend-exclude = [
+    "docs"
+]
+
+[tool.ruff.lint]
+# F = pyflakes, E/W = pycodestyle, I = isort, D = pydocstyle (replaces flake8_docstrings)
+select = ["F", "E", "W", "I", "D"]
+
+[tool.ruff.lint.per-file-ignores]
+"tests/**" = ["D", "E501"]  # mirror current flake8 `tests:*:D` per-file-ignore + ignore long lines
```
Eijebong:
```suggestion
"tests/**" = ["D", "E501"]
```

This comment has no value whatsoever outside of that changeset and should be in the commit message.

Eijebong: Also this is saying it mirrors the flake8 but it doesn't. It relaxes E501 (line length) on all tests instead of just `test_github.py`

Probably want something like this instead

```
"tests/**" = ["D"]
# Ignore long lines in test_github.py because XXX
"tests/test_github.py" = ["E501"]
```

Eijebong (dismissed):
Would look better without the claude-isms.

bhearsum (approved):
In the future please separate out the tool change from the reformatting into separate commits for ease of review.

This is the second (I think?) repository these changes have happened in. It would be good to codify this in https://github.com/mozilla-releng/reps at some point.

PR #798, opened 2026-05-19
Expose the temporary tc credentials to the script

jcristau:
Historically scriptworker scripts have used static credentials and scopes to perform their task.  For builddecisionscript however I'd like to use the scopes from the task itself to create the downstream decision task, which means the script needs access to the temporary credentials returned by claimWork or reclaimTask.

### DIFF
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -59,6 +59,7 @@ class Context(object):
 
     config: Optional[Dict[str, Any]] = None
     credentials_timestamp: Optional[int] = None
+    credentials_fd: int = -1
     proc: Optional[task_process.TaskProcess] = None
     queue: Optional[Queue] = None
     session: Optional[aiohttp.ClientSession] = None
@@ -98,6 +99,8 @@ class Context(object):
         if claim_task:
             self.task = claim_task["task"]
             self.verify_task()
+            # flags=0 to let the child inherit this fd
+            self.credentials_fd = os.memfd_create("scriptworker_temp_creds", flags=0)
             self.temp_credentials = claim_task["credentials"]
             path = os.path.join(self.config["work_dir"], "task.json")
             assert self.task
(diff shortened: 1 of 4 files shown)


src/scriptworker/context.py
```diff
@@ -193,6 +193,10 @@ def temp_credentials(self) -> Optional[Dict[str, Any]]:
     def temp_credentials(self, credentials: Optional[Dict[str, Any]]) -> None:
         self._temp_credentials = credentials
         self.temp_queue = self.create_queue(self.temp_credentials)
+        if credentials and self.config:
+            os.makedirs(os.path.join(self.config["work_dir"], "secrets"), mode=0o700, exist_ok=True)
+            path = os.path.join(self.config["work_dir"], "secrets", "credentials.json")
+            self.write_json(path, credentials, "Writing temp credentials to {path}...")
```
bhearsum: I know the risk of exfiltration is low on scriptworkers, but it still seems a bit scary to write these out to disk. Can we stick the credentials just in the env instead? (Obviously it's still possible to exfiltrate, but that would avoid the possibility of eg: accidentally uploading as an artifact.)

jcristau: The problem with passing the credentials directly is those credentials change on reclaim, so we wouldn't be able to refresh.

One thing we could do is instead of passing a path to the file, passing an open file descriptor that refers to a deleted temporary file or a memfd?

bhearsum: Ah, yes, that seems important to be able to do (maybe not in the build-decision case, but certainly in the general case).

> One thing we could do is instead of passing a path to the file, passing an open file descriptor that refers to a deleted temporary file or a memfd?

Sounds fine to me.

(It's a bit unfortunate that scriptworker uses subprocesses to run scripts...if it was all done in the same process this would be easy to solve in a clean way.)

bhearsum (approved):

jcristau:
Updated to add tests.

PR #800, opened 2026-06-12
Switch the temporary credentials file from memfd to a regular tempfile

jcristau:
In #798 I completely overlooked the fact that this code needs to run on mac not just linux, so this reverts to a non-linux-specific mechanism.


bhearsum (approved):
I think I pushed for the memory file, also overlooking this fact. Thanks for fixing this.

PR #810, opened 2026-07-08
config: make cot_product optional

jcristau:
- set the default cot_product to None
- when cot_product is None, apply_product_config resolves by-cot-product mappings to None as well
- check_config allows str values for cot_product despite the None default, and checks that verify_chain_of_trust is off when cot_product is not set, since chain of trust verification relies on cot_product-specific config


ahal (approved):
Thanks!

PR #811, opened 2026-07-30
fix: Gracefully handle event loops in fresh process

hneiva:
In Python 3.12/3.13 they deprecated get_event_loop() returning a new loop when none is current. In Python 3.14 it now raises a RuntimeError.
Reference: https://docs.python.org/3.14/library/asyncio-eventloop.html#asyncio.get_event_loop

### DIFF
--- a/src/scriptworker/context.py
+++ b/src/scriptworker/context.py
@@ -25,7 +25,7 @@ import arrow
 from taskcluster.aio import Queue
 
 from scriptworker import task_process
-from scriptworker.exceptions import CoTError
+from scriptworker.exceptions import CoTError, ScriptWorkerException
 from scriptworker.utils import load_json_or_yaml_from_url, makedirs, scriptworker_session
 
 log = logging.getLogger(__name__)
@@ -235,13 +235,21 @@ class Context(object):
 
     @property
     def event_loop(self) -> asyncio.AbstractEventLoop:
-        """asyncio.BaseEventLoop: the running event loop.
+        """asyncio.BaseEventLoop: the event loop set on the context, or the running one.
 
         This fixture mainly exists to allow for overrides during unit tests.
 
+        Raises:
+            ScriptWorkerException: if no loop has been set and none is running.
+
         """
         if not self._event_loop:
-            self._event_loop = asyncio.get_event_loop()
+            try:
+                self._event_loop = asyncio.get_running_loop()
+            except RuntimeError as exc:
+                raise ScriptWorkerException(
+                    "No event loop is running and none was set on the context. Set ``context.event_loop`` before use, or read it from a coroutine."
+                ) from exc
         return self._event_loop
 
     @event_loop.setter
--- a/src/scriptworker/worker.py
+++ b/src/scriptworker/worker.py
@@ -236,14 +236,16 @@ def main(event_loop=None):
 
     Args:
         event_loop (asyncio.BaseEventLoop, optional): the event loop to use.
-            If None, use ``asyncio.get_event_loop()``. Defaults to None.
+            If None, create one with ``asyncio.new_event_loop()``. Defaults to None.
 
     """
     context, credentials = get_context_from_cmdln(sys.argv[1:])
     log.info("Scriptworker starting up at {} UTC".format(arrow.utcnow().format()))
     log.info("Worker FQDN: {}".format(socket.getfqdn()))
     cleanup(context)
-    context.event_loop = event_loop or asyncio.get_event_loop()
+    if event_loop is None:
+        event_loop = asyncio.new_event_loop()
+    context.event_loop = event_loop
 
     done = False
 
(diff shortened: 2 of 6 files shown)


HISTORY.rst
```diff
@@ -4,6 +4,14 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`__.
 
+63.2.1 - 2026-07-30
```
jcristau: Date is wrong, and the changelog is incomplete.

src/scriptworker/context.py
```diff
@@ -237,11 +237,18 @@ def projects(self, projects: Optional[Dict[str, Any]]) -> None:
...
+
         This fixture mainly exists to allow for overrides during unit tests.
 
         """
         if not self._event_loop:
-            self._event_loop = asyncio.get_event_loop()
+            try:
+                self._event_loop = asyncio.get_running_loop()
+            except RuntimeError:
+                self._event_loop = asyncio.new_event_loop()
```
jcristau: This seems kind of awful.  Can we do something consistent here instead?  What do the tests actually require?

jcristau (dismissed):

HISTORY.rst
hneiva: it wasn't wrong yesterday 😅 

I'll remove the changelog and version bump and let that happen in a new commit

src/scriptworker/context.py
hneiva: I agree it's awful.
But it's better than `if asyncio.get_event_loop_policy()._local._loop is not None`

They don't have a way of checking if there is a current event loop and actually recommend this exact method 🫠

src/scriptworker/context.py
```diff
@@ -235,13 +235,21 @@ def projects(self, projects: Optional[Dict[str, Any]]) -> None:
...
+
         """
         if not self._event_loop:
-            self._event_loop = asyncio.get_event_loop()
+            try:
+                self._event_loop = asyncio.get_running_loop()
+            except RuntimeError as exc:
+                raise ScriptWorkerException(
+                    "No event loop is running and none was set on the context. Set ``context.event_loop`` before use, or read it from a coroutine."
+                ) from exc
```
hneiva, line 252: Only change here now is that we raise a `ScriptWorkerException` instead of letting asyncio raise a `RuntimeError`

src/scriptworker/worker.py
```diff
@@ -236,14 +236,21 @@ def main(event_loop=None):
...
     context, credentials = get_context_from_cmdln(sys.argv[1:])
     log.info("Scriptworker starting up at {} UTC".format(arrow.utcnow().format()))
     log.info("Worker FQDN: {}".format(socket.getfqdn()))
     cleanup(context)
-    context.event_loop = event_loop or asyncio.get_event_loop()
+    # a loop we create is ours to close; a caller-supplied one is left open
+    if event_loop is None:
+        owned_event_loop = asyncio.new_event_loop()
+        context.event_loop = owned_event_loop
+    else:
```
bhearsum: It's a tad unfortunate that the only use for this path is in tests...but that's not easy to change considering some of the tests appear to need to re-use the same event loop for other things. Definitely out of scope here.

src/scriptworker/worker.py
```diff
@@ -262,15 +269,20 @@ async def _handle_sigusr1():
...
+            except Exception:
+                log.critical("Fatal exception", exc_info=1)
+                raise
+        else:
+            log.info("Scriptworker stopped at {} UTC".format(arrow.utcnow().format()))
+            log.info("Worker FQDN: {}".format(socket.getfqdn()))
+    finally:
+        if owned_event_loop is not None:
+            # closing the loop also deregisters the signal handlers added above
+            owned_event_loop.close()
```
bhearsum: To be honest, I don't really think there's a need to explicitly close this, which would avoid the need to keep track of it like this. There's no python code that runs after this point...so there's no harm in not explicitly closing it IMO.

Not going to block on this though.

tests/test_worker.py
```diff
@@ -62,6 +73,116 @@ async def foo(arg, credentials):
...
+        with pytest.raises(ScriptWorkerException):
+            worker.main(event_loop=event_loop)
+        assert not event_loop.is_closed()
+    finally:
+        event_loop.close()
+        os.remove(tmp)
+
+
+def test_main_repeated_calls_do_not_leak_loops(mocker, context):
+    """Restarting ``main()`` in one process doesn't accumulate open loops."""
```
bhearsum: note: the only way for `main` to be called outside of tests is by running `scriptworker`. There is no way for it to be called more than once per process. IMO you should scrap this test, as it doesn't cover any remotely real scenario.

bhearsum (approved):
I'm r+'ing, but please seriously consider scrapping the noted test, and maybe even dropping the event loop closing to avoid unnecessary complexity (and tests).

bhearsum (approved):
Looks fantastic, thanks for simplifying it!

PR #815, opened 2026-08-05
feat: set Content-MD5 header to avoid silent data corruption

bhearsum:
In [bug 2060889](https://bugzilla.mozilla.org/show_bug.cgi?id=2060889) we're looking at some chain of trust verification failures caused by hash mismatches between `chain-of-trust.json` and the files in GCS. After a great deal of debugging, I ended up finding that [internal aiohttp retries on PUT requests can send incomplete data](https://github.com/aio-libs/aiohttp/issues/13329). The short version of that bug is that the file handle is advanced every time a request is prepared, and if a failure+retry happens after that, it advances to the next chunk of data rather than resending the failed chunk.

GCS uploads [support the Content-MD5 header](https://docs.cloud.google.com/storage/docs/data-validation#xml-single-request-upload), which will cause a 400 error to be returned if the uploaded data does not match the md5 provided in the header. Setting this will let us retry at our level if an upload is corrupt in this specific scenario, but more importantly, for any other future scenario that may come up. This means that we don't benefit from aiohttp's retries, which might save a bit of bandwidth and time, but it's a more general solution.

A different way to work around this would be avoid passing plain file handles and either pass the data (impractical due to the size of the files), or one of aiohttp's Payload objects. The latter seems to work, although they're not well documented, and I don't understand it enough to feel confident in doing so.

### DIFF
--- a/src/scriptworker/artifacts.py
+++ b/src/scriptworker/artifacts.py
@@ -6,6 +6,7 @@ in S3.
 """
 
 import asyncio
+import base64
 import fnmatch
 import gzip
 import logging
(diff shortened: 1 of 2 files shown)


ahal (approved):
Sounds reasonable!

Seems like a pretty bad bug in aiohttp though, which is surprising. Maybe we're not using an up-to-date version.

bhearsum:
> Sounds reasonable!
> 
> Seems like a pretty bad bug in aiohttp though, which is surprising. Maybe we're not using an up-to-date version.

Indeed! I filed https://github.com/aio-libs/aiohttp/issues/13329 for it, which is being worked on in https://github.com/aio-libs/aiohttp/pull/13330

PR #817, opened 2026-08-07
WIP - Check gecko and comm for Enterprise usecase

lissyx:


src/scriptworker/cot/verify.py
```diff
@@ -1952,12 +1952,26 @@ def get_source_url(obj):
...
+                "{name} {task_id}: {source_env_prefix} {repo} doesn't match source {source}!".format(
+                    name=obj.name, task_id=obj.task_id, source_env_prefix=source_env_prefix, repo=repo, source=source
+                )
             )
-        )
+    else:
+        comm_repo = get_repo(obj.task, source_env_prefix="COMM", repo_type="head")
+        match_comm_repo = comm_repo and verify_repo_matches_url(comm_repo, source)
+
+        if not match_repo and comm_repo and not match_comm_repo:
```
ahal: The `and comm_repo` in this condition is a security bug.

Let's say there's a normal Firefox Enterprise push to `mozilla/enterprise-firefox`, then `COMM_HEAD_REPOSITORY` will not be set and this condition will always evaluate to `False`. This would allow a malicious task to spoof `metadata["source"]` without raising a CoT error.

Instead the condition should just be:
```
if not match_repo and not match_comm_repo:
```

